PR #11, opened 2021-03-19
Fix golangci-lint errors

drpaneas:
In case you want to enable `golangci-lint` for this project, feel free to use my config at https://github.com/romie-gr/romie/blob/master/.golangci.yml


flavio:
Thanks @drpaneas, nice seeing you again :)

PR #12, opened 2021-03-19
Add linting via golangci-lint

flavio:
* Perform linting via golangci-lint
* Enable GitHub Action
* Fix linting errors and warnings


ereslibre (approved):
+1 for linting

PR #16, opened 2021-03-24
Add support for mutating policies

flavio:
Allow the definition of mutating policies


flavio:
Depends on: https://github.com/chimera-kube/policy-server/pull/34

ereslibre (approved):

PR #17, opened 2021-04-02
Use `GITHUB_TOKEN` instead of a personal access token

ereslibre:
This repo still requires other changes, but this is an isolated one that should work fine.


flavio:
We have to wait for the package to be created for the first time before we can actually merge it.

flavio (approved):
LGTM, but let's wait for the 1st `kubewarden/kubewarden-controller` image to be pushed with the regular PAT.

PR #18, opened 2021-04-02
Rename chimera to kubewarden

ereslibre:
-   Also, rename AdmissionPolicy to ClusterAdmissionPolicy, since we will
    introduce a namespaced AdmissionPolicy soon; we will keep
    ClusterAdmissionPolicy for policies that need to be applied to the
    whole cluster, or to resources that are not namespaced.

-   Rename CRD asciidocs.
    Also, add a way to easily generate them

-   Use example that works in the README
    
    With all the refactoring and renaming changes, while we tag a new
    version of the controller, use an example that works on the README.
    
    When a new tag is created, update the README, so we pin the sample to
    a well known version.


flavio:
@ereslibre I think it looks good. I made another commit to ensure the domain is changed from `kubewarden.suse.com` to `kubewarden.io`

Can you please review my changes?

README.md
```diff
@@ -46,7 +46,7 @@ The following snippet defines a Kubewarden Policy based on the
 policy:
 
 ```yaml
-apiVersion: kubewarden.suse.com/v1alpha1
+apiVersion: kubewarden.io/v1alpha1
```
ereslibre, line 49: It's common for the group to be of the form `subdomain.domain` for external API types, not `domain` only. I don't know if `domain` works or if there's somewhere on the machinery where a subdomain is needed. Thing is I have only seen subdomains, never domains directly.

ereslibre (commented):
Strange, I cannot either approve or request changes. I was going to request changes, but I just comment instead :)

README.md
ereslibre: My comment applies here and to all instances of this change, we could use `policy.kubewarden.io/v1alpha1`, for example.

flavio: good point, let me change that to `policies.kubewarden.io`. I'll update the PR

flavio:
I updated the domain name. @ereslibre please take a look, especially check what I changed compared to your commit

config/crd/bases/secops.kubewarden.io_clusteradmissionpolicies.yaml
```diff
@@ -0,0 +1,102 @@
+
```
ereslibre, line 1: The name of this file is in purpose?

ereslibre (commented):
Will review in depth tomorrow, a first question though.

ereslibre (commented):
LGTM, besides the base filename renamed to secops :)

config/crd/bases/secops.kubewarden.io_clusteradmissionpolicies.yaml
flavio: Sorry, that's a leftover (I initially though about calling everything like `secops.kubewarden.io` instead of `policies.kubewarden.io`).

I've removed it from the PR

ereslibre: Thanks!

ereslibre:
Let's get it merged and continue enhacing later :), thanks @flavio!

Issue #22, opened 2021-04-08
GitHub Action: automatically create releases


flavio:
Right now our GH action only builds the container image of kubewarden-controller and pushes it to our container registry.

The action should be expanded to create a new release entry on GitHub every time we push a tag that starts with `v<something>`. This is currently not done and, as a result of that, kubewarden is not listing the v0.1.0 release, but only the v0.0.1 one.

PR #23, opened 2021-04-09
Add Github release action

ereslibre:
This action will create a Github release whenever a tag `v<version>`
is pushed to the repository.

Image building and publishing into the image repository is still
handled by the `container-image` workflow.

Fixes https://github.com/kubewarden/kubewarden-controller/issues/22

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+on:
+  push:
+    tags:
+    - 'v*'
+
+name: publish release
+
+jobs:
+  build:
+    name: Create new Github release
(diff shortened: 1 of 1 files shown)


flavio (approved):
LGTM. I wonder if we shouldn't start to keep a CHANGELOG and add that into the releases

ereslibre:
> LGTM. I wonder if we shouldn't start to keep a CHANGELOG and add that into the releases

I'm creating an issue later today to keep track and let's improve it later.

ereslibre:
Registered https://github.com/kubewarden/kubewarden-controller/issues/26

PR #25, opened 2021-04-09
Allow to configure policy-server service account

ereslibre:
The policy-server can use the provided service account, so it can be
granted permissions to list certain resources that will be exposed to
context-aware policies.

Related: https://github.com/kubewarden/helm-charts/issues/4

A helm chart PR will follow that depends on this PR.


flavio (approved):
LGTM, thanks!

Issue #26, opened 2021-04-10
Create an initial CHANGELOG


ereslibre:
Create an initial CHANGELOG and see what automation, if any could be integrated to generate it to reduce manual tasks.

**Acceptance criteria**

* An initial CHANGELOG is introduced
* There is a clear way to add entries to it and the process is easily repeatable when new releases are about to be created

ereslibre:
Do we expect a CHANGELOG with all commits listed from the last tag to the current HEAD, a RELEASE_NOTES text file where commits are scanned for release notes, or both?

cc/ @flavio

ereslibre:
I think we can start by using https://github.com/git-chglog/git-chglog

PR #27, opened 2021-04-13
feat: add changelog generation tooling

ereslibre:
Fixes https://github.com/kubewarden/kubewarden-controller/issues/26

After this is merged, I will run this mechanism and create the initial CHANGELOG.md with a tagged release.

### DIFF
--- a/.chglog/CHANGELOG.tpl.md
+++ b/.chglog/CHANGELOG.tpl.md
@@ -1,0 +1,56 @@
+{{ if .Versions -}}
+<a name="unreleased"></a>
+## [Unreleased]
+
+{{ if .Unreleased.CommitGroups -}}
+{{ range .Unreleased.CommitGroups -}}
+### {{ .Title }}
+{{ range .Commits -}}
+- {{ if .Scope }}**{{ .Scope }}:** {{ end }}{{ .Subject }}
+{{ end }}
+{{ end -}}
+{{ end -}}
+{{ end -}}
+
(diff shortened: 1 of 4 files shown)


flavio:
Is there a way to change the automatically generated changelog?

Sometimes is not a good idea to leave all the commit messages inside of it

ereslibre:
> Sometimes is not a good idea to leave all the commit messages inside of it

Since we are providing in the configuration a list of filters (https://github.com/kubewarden/kubewarden-controller/pull/27/commits/1a21af8313773768f0067961f9b1d85a3d088d64#diff-f27299617f81a75a6e22804bf0f13d035bf301b7a2b5ee7906bdda95be97055dR10-R13), you need to prefix your commit headline with any of those: `feat`, `fix`, `perf` or `refactor`.

Any other commit won't be included in the changelog. By default, the tool includes pull request merge commits. If we consider it's noisy and don't have value we can remove that part, but I wanted to give it a try and decide later on with the experience. If you have a different opinion we can update the behavior.

flavio:
I'm fine with making a test, I just fear this will always let something slip into the changelog. When that happens I would like the ability to manually fix it.

Would that be doable?

ereslibre:
> Would that be doable?

Not at this time. The only source of truth is the repo, and the contents are regenerated from scratch on every run.

ereslibre:
Merging as discussed, we plan to separate the changelog and the release notes.

PR #32, opened 2021-05-19
Upgrade from kubebuilder v2 -> v3

flavio:
This involves lots of changes, unfortunately there's no other way.

This upgrade is required because only kubebuilder v3 can handle multiple
versions of the same custom resource to be defined at the same time.
We need that to change the current ClusterAdmissionPolicy from v1alpha1
to v1alpha2.


PROJECT
```diff
@@ -1,7 +1,17 @@
 domain: kubewarden.io
+layout:
+- go.kubebuilder.io/v3
+multigroup: true
+projectName: kubewarden-controller-migrated
```
ereslibre, line 5: Leftover project name?

config/default/kustomization.yaml
```diff
@@ -1,12 +1,12 @@
 # Adds namespace to all resources.
-namespace: kubewarden-controller-system
+namespace: kubewarden-controller-migrated-system
```
ereslibre, line 2: `migrated` leftover

config/default/kustomization.yaml
```diff
@@ -1,12 +1,12 @@
...
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: kubewarden-controller-
+namePrefix: kubewarden-controller-migrated-
```
ereslibre, line 9: `migrated` leftover

PROJECT
```diff
@@ -1,7 +1,17 @@
...
 repo: github.com/kubewarden/kubewarden-controller
 resources:
-- group: kubewarden
+- api:
+    crdVersion: v1
+    namespaced: true
```
ereslibre, line 10: Based on https://book.kubebuilder.io/reference/project-config.html, I think this should be false, given that `ClusterAdmissionPolicy` is cluster based.

apis/policies/v1alpha1/clusteradmissionpolicy_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-
+Copyright 2021.
```
ereslibre, line 2: See my comment on boilerplate.go.txt, we have to update that, and regenerate with `make generate`.

hack/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-
+Copyright 2021.
```
ereslibre, line 2: Update this file, so the generated headers have the correct contents.

ereslibre (changes requested):
Thank you @flavio. It's pretty big, some comments on a first pass.

PROJECT
flavio: Oh no... I'm going to fix that.

PROJECT
flavio: I didn't know kubebuilder introduced that into the `PROJECT` file. I will change it, but in the meantime it looks like the `//+kubebuilder:resource:scope=Cluster` directive inside of `apis/policies/v1alpha1/clusteradmissionpolicy_types.go` works just fine

flavio:
Thanks for having spotted the issues. I've changed the "high level" files and then I ran `make generate && make manifests`.
To my surprise, not a lot changed. You can see that by looking at the second commit I just pushed.

Obviously, I'll squash everything into a single commit before merging the code

ereslibre:
> To my surprise, not a lot changed.

Yes, just tried it, only the deepcopy generated file. Seems we will have to update manually then? Thought there was an automated way to rewrite that with kubebuilder/controller-gen, besides the sed obvious one...

ereslibre (changes requested):
LGTM, modulo the removal of the `tag` Makefile target :)

Makefile
```diff
@@ -11,52 +11,42 @@ else
...
 endif
 
-all: manager
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# This is a requirement for 'setup-envtest.sh' in the test target.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
 
-tag:
```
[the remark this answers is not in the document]
flavio: Good point, I fixed that with https://github.com/kubewarden/kubewarden-controller/pull/32/commits/38f571f486868aed7a6869010ea4f7b29c608a03

flavio:
@ereslibre : please let me know if the latest fix addresses your concerns. If that's the case, I'll squash and merge this PR

ereslibre (approved):

Issue #43, opened 2021-07-13
Allow configuration of policy-server logging strategy


flavio:
Policy server allows to configure the following aspects of his log behaviour:

  * Log level
  * Log format

Both aspects can be configured either via cli flag or via dedicated environment variables.

Currently kubewarden-controller reads the configuration params of policy server from a dedicated ConfigMap called `policy-server`. This ConfigMap is created by our helm chart.

The kubewarden-controller should be extended to read the logging-specific params from the `policy-server` ConfigMap, and then use them to create the right Deployment object for policy server.

The next section cover the logging params that can be set.

## Log level

The log level to be used can be specified either via the `--log-level` flag or via the `KUBEWARDEN_LOG_LEVEL` environment variable.

## Log format

The log format to be used can be specified either via the `--log-fmt` flag or via the `KUBEWARDEN_LOG_FMT` environment variable.

Depending on the log format, other options might have to be specified.

### Jaeger

When the `jaeger` format is used, logs are sent to a Jaeger collector.

Setting `KUBEWARDEN_LOG_FMT=jaeger` is not enough, our controller has to support also the following scenarios.

#### Tuning via environment variables

By default, policy-server will attempt to send the trace a events to a Jaeger collector that is listening on localhost. This will not work with our deployment.

The user can configure some Jaeger options via these dedicated environment variables (note well: there are no cli flags):

| Name                            | Description                                                      | Default                                                                                          |
|---------------------------------|------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| OTEL_EXPORTER_JAEGER_AGENT_HOST | Hostname for the Jaeger agent                                    | "localhost"                                                                                      |
| OTEL_EXPORTER_JAEGER_AGENT_PORT | Port for the Jaeger agent                                        | 6832                                                                                             |
| OTEL_EXPORTER_JAEGER_ENDPOINT   | HTTP endpoint for Jaeger traces                                  |  "http://localhost:14250" |
| OTEL_EXPORTER_JAEGER_TIMEOUT    | Maximum time the Jaeger exporter will wait for each batch export | 10s                                                                                              |
| OTEL_EXPORTER_JAEGER_USER       | Username to be used for HTTP basic authentication                | -                                                                                                |
| OTEL_EXPORTER_JAEGER_PASSWORD   | Password to be used for HTTP basic authentication                | -                                                                                                |

Our controller should always forward these environment variables into the policy-server deployment.

#### Jaeger operator

[...]

PR #44, opened 2021-07-20
Policy server logging

flavio:
This PR makes the controller aware of the logging options of policy-server. This fixes https://github.com/kubewarden/kubewarden-controller/issues/43

The PR also changes the way the policy-server Deployment is updated. Prior to this PR, the Deployment definition was **never** changed by the controller.

Current limitations: the Deployment definition is updated only when a policy is added/removed. This is a know issue which is going to be tackled with another PR


flavio:
@viccuad, @ereslibre I think we can go ahead, review and merge this code. By doing that @viccuad will have a solid foundation for his current work.

I'll file other issues to keep track of the other tasks to accomplish: extend the helm chart, make the controller watch also the `policy-server` ConfigMap and react on changes done to it

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -167,9 +168,20 @@ func policyServerDeploymentSettings(cfg *corev1.ConfigMap) PolicyServerDeploymen
 		settings.Image = buf
 	}
 
-	buf, found = cfg.Data[constants.PolicyServerLogLevel]
-	if found {
-		settings.EnvVars["KUBEWARDEN_LOG_LEVEL"] = buf
+	relevantEnvVarPrefixes := []string{
+		"KUBEWARDEN_", "OTEL_",
```
viccuad, line 160: :100:  nice!

viccuad (approved):
LGTM! nice things, even if we change on reading the configmap :).

ereslibre (approved):
LGTM

flavio:
Closing the PR, the new architecture makes it obsolete

PR #46, opened 2021-07-23
Add RFC about new architecture

flavio:
This RFC describes how we can change the current architecture to make the Kubewarden stack easier to configure and, at the same time, improve scalability and resiliency.

[Rendered link](https://github.com/flavio/kubewarden-controller/blob/policy-server-crd-proposal/rfc/0001-policy-server-custom-resource.md).

### DIFF
--- a/rfc/0001-policy-server-custom-resource.md
+++ b/rfc/0001-policy-server-custom-resource.md
@@ -164,7 +164,7 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: psp-capabilities
 spec:
-  policy_server: reserved-instance-for-tenant-a
+  policyServer: reserved-instance-for-tenant-a
   module: registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.3
   rules:
     - apiGroups: [""]
@@ -412,4 +412,3 @@ On the other hand, we are about to start working on a Rancher integration. We sh
 the UI on a solid foundation. Which translates to: if we build the UI to manage the
 Custom Resources defined inside of this proposal we will have a something future proof. Instead,
 if we design on the current architecture we might have to change **also** the external UI code!
-
--- a/rfc/0001-policy-server-custom-resource.md
+++ b/rfc/0001-policy-server-custom-resource.md
@@ -241,12 +241,18 @@ Deletion of a `PolicyServer`:
 
   * Thanks to the finalizer, the `PolicyServer` resource is not yet deleted
     from etcd.
-  * Go through all the policies that are owned by the `PolicyServer` instance,
-    for each one of them:
-    * Delete the `ValidatingWebhookConfiguration` or the `MutatingWebhookConfiguration`
-      resource associated with the policy
-    * Delete the `ClusterAdmissionPolicy` from etcd.
-  * Delete the `PolicyServer` object from etcd
+  * Ensure these operations take place. Note well: the order of execution
+    matters
+    * For each policy owned by the `PolicyServer` instance:
+      - Delete the `ValidatingWebhookConfiguration` or the `MutatingWebhookConfiguration`
+        resource associated with the policy
+      - Delete the `ClusterAdmissionPolicy` from etcd
+    * Delete the `PolicyServer` object from etcd
+
+We have to research whether this can be done by the Kubernetes Garbage Collector
+by setting the [owner references](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#owner-references)
+or whether this has to be handled by us to ensure the right ordering takes place.
+
 
 ## Validation and mutation of our own Custom Resources
 
--- a/rfc/0001-policy-server-custom-resource.md
+++ b/rfc/0001-policy-server-custom-resource.md
@@ -3,7 +3,7 @@
 | Feature Name | policy-server custom resource    |
 | Start Date   | Jul 22nd 2021                    |
 | Category     | Architecture                     |
-| RFC PR       | [fill this in after opening PR]  |
+| RFC PR       | [PR46](https://github.com/kubewarden/kubewarden-controller/pull/46)  |
 
 # Summary
 [summary]: #summary
@@ -72,9 +72,9 @@ This RFC aims to solve both problems.
 The proposed solution is to:
 
   * Introduce a new Custom Resource Definition describing a PolicyServer deployment
-  * Create a parent-child relationship between a PolicyServer CR and the ClusterAdmissionPolicies
-    CR: one PolicyServer can have many ClusterAdmissionPolicies; a ClusterAdmissionPolicy
-    belongs to one PolicyServer.
+  * Bind the ClusterAdmissionPolicies CR to the PolicyServer CR: one PolicyServer
+    can own many ClusterAdmissionPolicies; a ClusterAdmissionPolicy
+    is bound to one PolicyServer.
 
 By creating a PolicyServer object we can move all the configuration details of the
 Policy Server Deployment in a nicely structured place. Providing to the end
@@ -368,6 +368,12 @@ Maybe we can extend our helm chart to have these configuration knobs:
 However, if we introduce the cert-manager dependency, we could also leverage that to
 create the certificates used by the Policy Server instances.
 
+Finally, we could also avoid the usage of cert-manager by having the Kubewarden controller
+generate its own CA authority and secrets. This is similar to what we are
+currently doing.
+
+The right approach needs more discussion, but this can be done after this RFC
+is approved.
 
 ## Kubewarden controller developer experience
 
--- a/rfc/0001-policy-server-custom-resource.md
+++ b/rfc/0001-policy-server-custom-resource.md
@@ -1,0 +1,415 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | policy-server custom resource    |
+| Start Date   | Jul 22nd 2021                    |
+| Category     | Architecture                     |
+| RFC PR       | [fill this in after opening PR]  |
+
+# Summary
+[summary]: #summary
+
+> Brief (one-paragraph) explanation of the feature.
+
+Introduce a `PolicyServer` Custom Resource definition that allow users to
+describe a Policy Server deployment.
+
+Allow `ClusterAdmissionPolicies` to target a specific PolicyServer instance.
+
+# Motivation
+[motivation]: #motivation
+
+> - Why are we doing this?
+> - What use cases does it support?
+> - What is the expected outcome?
+
+> Describe the problem you are trying to solve, and its constraints, without coupling them too closely to the solution you have in mind. If this RFC is not accepted, the motivation can be used to develop alternative solutions.
+
+Currently the Kubewarden controller deploys a single Deployment of Policy Server.
+All the `ClusterAdmissionPolicies` are then evaluated by the Policy Server Pods
+that belong to this Deployment.
+The configuration of this Policy Server Deployment is kept inside of a ConfigMap
+called `policy-server`.
+
+Right now, the kubewarden-controller is monitoring the ClusterAdmissionPolicies
+resources. Changes against these resources trigger an update of the `policy-server`
+ConfigMap, and then a restart of the Policy Server Deployment.
+
+The `policy-server` ConfigMap contains a series of configuration options related
+with the Deployment (e.g.: the image to use, the size of the replica,...) and
+a snapshot of the `policies.yml` file. The latter one is the file that contains
+the list of policies to load, where to pull them, their configuration options,...
+
+Currently, user-made changes to the `policy-server` ConfigMap (e.g. manual
+edit via `kubectl`, or `helm upgrade`) **do not** lead to the Policy Server Deployment
+to be updated.
+Moreover, when changing the `policy-server` ConfigMap, the user
+must be really careful to not alter the contents of the `policies.yml` key. That
+can result in catastrophic outcome (all policies being dropped by the Policy Server,
+leading to massive failures to all the admission reviews, leading to potential
+denial of service against the Kubernetes cluster).
+
+> **Problem #1:** provide a better UX to update the tuning
+> options of the Policy Server Deployment. This must be easy and safe.
+
+Thinking long term, there's a scalability and reliability issue with our current
+architecture. All the policies are loaded by each Policy Server process. Inherently,
+all the admission reviews are sent to the single deployment.
+
+While the current setup is already HA and can be scaled by increasing the number
+of replicas, there are still some hard limits and not-so-uncommon scenarios
+that could lead to a service disruption:
+
+  * Defining too many ClusterAdmissionPolicies can lead to Policy Server processes
+    consuming a lot of memory and even hit resource limits
+  * A "noisy" tenant/namespace or a frequently used policy can slow down policy
+    evaluation and potentially bring down a Policy Server instance
+
+> **Problem #2:** provide a way to partition policies. Allow the operator to create
+> dedicated Policy Server Deployments, handling only a chosen set of policies.
+
+This RFC aims to solve both problems.
+
+The proposed solution is to:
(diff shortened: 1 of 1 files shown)


rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,415 @@
...
+  * Delete the `ValidatingWebhookConfiguration` or the `MutatingWebhookConfiguration`
+    resource associated with the policy.
+  * Delete the `ClusterAdmissionPolicy` from etcd
+  * Perform the reconciliation loop of the `PolicyServer` that owned the policy
+
+Deletion of a `PolicyServer`:
+
+  * Thanks to the finalizer, the `PolicyServer` resource is not yet deleted
+    from etcd.
+  * Go through all the policies that are owned by the `PolicyServer` instance,
```
viccuad, line 244: Just learned about [owner references](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#owner-references). 
Would it make sense for the controller to add owner references to policies with the policy-server as parent? (and maybe do the same for WebhookConfigurations and policies?).

That way things would get scheduled for deletion automatically, and can be a signifier that the controller could find the policy-server that is in control of those policies, as listed in the policies.

Owner references seem to be a list; one can have several, which would make it work for several policy-servers consuming the same policies.

kkaempf (approved):
I'd like to see an effort estimation before we start with the implementation.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,415 @@
...
+
+The `ClusterAdmissionPolicy` CR will be extended to have a new attribute: `policy_server`:
+
+```yaml
+apiVersion: policies.kubewarden.io/v1alpha2
+kind: ClusterAdmissionPolicy
+metadata:
+  name: psp-capabilities
+spec:
+  policy_server: reserved-instance-for-tenant-a
```
ereslibre, line 167:
```suggestion
  policyServer: reserved-instance-for-tenant-a
```

Nit.

rfc/0001-policy-server-custom-resource.md
ereslibre: We can use the cascade deletion, yes. I would leave it open in the wording of this document and decide in the implementation phase.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,415 @@
...
+However, if we introduce the cert-manager dependency, we could also leverage that to
+create the certificates used by the Policy Server instances.
+
+
+## Kubewarden controller developer experience
+
+Right now, to develop the controller, we just run `make run` and that's it.
+
+If we bring the webhook endpoint back, the local development might become more cumbersome.
+This needs more research, but it's not a blocking point IMHO.
```
ereslibre, line 371:
```suggestion
This needs more research, but it's not a blocking point.
```

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,415 @@
...
+
+This will potentially solve the 1st problem we have: provide a nice UX to configure
+Policy Server, plus it would reduce the maintenance efforts on our side.
+
+This wouldn't however solve the scalability and resiliency problems of Policy Server.
+
+Last but not least, from past experience (Cloud Foundry work at SUSE, other community
+projects), moving things on helm is tempting (you can go faster), but over a certain
+limit this will not scale. It will become hard to maintain the helm chart and impossible
+to add all the business logic you might need.
```
ereslibre, line 397: Also, there's the point of updating the policy list today by changing the configuration is not possible and can be problematic if done wrong. With the new proposal, this UX is in a safe spot now.

ereslibre (commented):
LGTM, some comments.

ereslibre (changes requested):
Please, note that with the `policy_server` -> `policyServer` suggestion, there are some other occurences to replace.

rfc/0001-policy-server-custom-resource.md
flavio: This is a good point. I was pondering if, by setting ownership relations, we could even avoid to use the finalizers.
I think this cannot be done, because we have the following resources:

|Resource | Scope |
|------------------|-----|
| PolicyServer CR | Cluster |
| Policy Server Service | Namespace |
| Policy Server Deployment | Namespace |
| Policy Server ConfigMap | Namespace |
| Policy Server Secret | Namespace |
| ValidatingWebhookConfiguration/MutatingWebhookConfiguration | Cluster |

I don't know if the Kubernetes GC allows a Cluster-scoped resource to own namespaced ones and, most important of all, if it handles the cleanup properly...

I know there were limitations in the past, but I couldn't find anything right now

rfc/0001-policy-server-custom-resource.md
flavio: Sorry, I don't get it. What do you mean with "policy list", are you referring to the contents of the `policy.yml` file that we feed to policy-server?

If that the case, this list is not meant to be managed by the user at the "raw" level (aka, changing the contents of the ConfigMap). The user manages that by using the ClusterAdmissionPolicy resource.

Am I missing something? :thinking:

ereslibre: > Sorry, I don't get it. What do you mean with "policy list", are you referring to the contents of the policy.yml file that we feed to policy-server?

Yes.

We are enumerating the advantages of the new implementation in comparison to the old one. Take this from this RFC:

> Currently, user-made changes to the policy-server ConfigMap (e.g. manual edit via kubectl, or helm upgrade) do not lead to the Policy Server Deployment to be updated. Moreover, when changing the policy-server ConfigMap, the user must be really careful to not alter the contents of the policies.yml key. That can result in catastrophic outcome (all policies being dropped by the Policy Server, leading to massive failures to all the admission reviews, leading to potential denial of service against the Kubernetes cluster).

This is also something we can mention briefly here again I think.

flavio: Addressed

flavio:
> I'd like to see an effort estimation before we start with the implementation.

@kkaempf if we decide to go with this approach, we can then split down this RFC into smaller cards and then estimate them individually.

This is definitely something to discuss on Mon, during our planning

rfc/0001-policy-server-custom-resource.md
flavio: I've updated the doc to mention the usage of owner references.

flavio:
@kubewarden/kubewaden-developers I think this ready to be discussed during the next planning meeting. I've incorporated all the feedback you gave to me.

Please take a last look before we do planning on Monday :pray:

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+
+> **Problem #2:** provide a way to partition policies. Allow the operator to create
+> dedicated Policy Server Deployments, handling only a chosen set of policies.
+
+This RFC aims to solve both problems.
+
+The proposed solution is to:
+
+  * Introduce a new Custom Resource Definition describing a PolicyServer deployment
+  * Create a parent-child relationship between a PolicyServer CR and the ClusterAdmissionPolicies
```
ereslibre, line 75:
```suggestion
  * Create a parent-child relationship between a PolicyServer CR and the ClusterAdmissionPolicies
```

After a second thought, I think is better if we talk about "binding" rather than "parent-child" relationship.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+    CR: one PolicyServer can have many ClusterAdmissionPolicies; a ClusterAdmissionPolicy
+    belongs to one PolicyServer.
```
ereslibre, line 77:
```suggestion
    belongs to one PolicyServer.
```

"is bound to"

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+    - When the user specifies a value for the `policyServer` attribute, ensure
+      there's actually a `PolicyServer` object with that name.
+  * `PolicyServer`:
+    - Set our finalizer
+    - Reject the delete operations against the `PolicyServer` named `default`
+    - Reject update operations that attempt to change the name of the `PolicyServer`.
+      The name of a `PolicyServer` must be immutable, a lot of relationships
+      between Kubernetes resources and the `PolicyServer` are built on that (for example,
+      the ConfigMap, the Service, the ClusterAdmissionPolicy)
+
```
ereslibre, line 275: We also want to prohibit `PolicyServer` deletion if any `ClusterAdmissionPolicy` is bound to it.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+cert-manager is already mandatory to install many other Kubernetes add-ons.
+
+Maybe we can extend our helm chart to have these configuration knobs:
+
+  * Install cert-manager
+  * Do not install cert-manager, because it already exists. I think this
+    should be the recommended option for production deployments
+  * Generate the secrets via helm. This [seems doable](https://medium.com/nuvo-group-tech/move-your-certs-to-helm-4f5f61338aca).
+    We would not be able to rotate them, this is definitely something for quick evaluation
+    of Kubewarden. That's not meant to be used in production
```
ereslibre, line 366: Another option exists: kubewarden-controller generates this secrets upon start and registers the webhook. It just needs a powerful enough kubeconfig with rights to do some key operations that lead to it registering the webhooks to itself on the cluster where it's running.

The previous option can either be used for non-production and production environments, depending on the trust level of the execution environment etc of kubewarden-controller itself.

I agree that maybe? in the majority of production environments users will prefer to use cert-manager.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+However, if we introduce the cert-manager dependency, we could also leverage that to
+create the certificates used by the Policy Server instances.
+
+
+## Kubewarden controller developer experience
+
+Right now, to develop the controller, we just run `make run` and that's it.
+
+If we bring the webhook endpoint back, the local development might become more cumbersome.
+This needs more research, but it's not a blocking point.
```
ereslibre, line 377: In the spirit of the previous comment, if we put the knobs needed, the controller can do all this on its own (like our old controller did if you remember :P)

ereslibre (approved):
I did a last proof-read. It's awesome. Some comments. Huge LGTM.

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
...
+  - name: KUBEWARDEN_LOG_LEVEL
+    value: debug
+  - name: KUBEWARDEN_LOG_FMT
+    value: jaeger
+  annotations:
+    sidecar.jaegertracing.io/inject: default
+```
+
+The `PolicyServer` CR will be a cluster-wide resource. Take makes each
+`PolicyServer` resource identifiable by its unique `name`.
```
raulcabello, line 157: Can we make the PolicyServer CR namespace scoped? I think we will need that if we want to use owner references. Owner references link resources within the same namespace, so I'm not sure if that will work for cluster scoped resources. In the end the controller, Policy Server deployment, services and configmaps will be in a namespace, so having the CR in the same namespace might be ok.

kkaempf:
> > I'd like to see an effort estimation before we start with the implementation.
> 
> @kkaempf if we decide to go with this approach
Do we have a real alternative ?

rfc/0001-policy-server-custom-resource.md
flavio: I've resolved that using the `binding` term

rfc/0001-policy-server-custom-resource.md
flavio: Do we? I think this is useful for quick cleanup operations

ereslibre: Allowing to delete a `PoilcyServer` when `ClusterAdmissionPolicy` are bound to it means that we will remove the webhook (at least temporarily until recreated, if recreated at all), and although the `ClusterAdmissionPolicy` is persisted during all that time, requests could have been accepted that the policy would have rejected if been active.

I think kubewarden should not try to be smart here, and prioritize safety.

rfc/0001-policy-server-custom-resource.md
flavio: The `ClusterAdmissionPolicy` resource, which will belong to a `PolicyServer`, is cluster-wide. The same applies to the `ValidatingWebhookConfiguration` and the `MutatingWebhookConfiguration` ones. These are the high level resources he should bind together.

This binding should definitely be possible, because they are all cluster-wide.

As for the namespaced resources, we have to figure out if the binding can be established. If this proves to not be possible, we can still manage them. We could even leave them behind when the cluster-wide resources are removed and everything would just keep working (ofc, this is not what we want to do).

rfc/0001-policy-server-custom-resource.md
```diff
@@ -0,0 +1,426 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | policy-server custom resource    |
+| Start Date   | Jul 22nd 2021                    |
+| Category     | Architecture                     |
+| RFC PR       | [fill this in after opening PR]  |
```
viccuad, line 6: (:smile:)

viccuad (approved):

rfc/0001-policy-server-custom-resource.md
raulcabello: I was thinking of making both ClusterAdmissionPolicy and PolicyServer namespace scoped, I think binding should be possible if both are in the same namespace. We could then use owner references, which I am not sure if we could use them if the resources are cluster wide

ereslibre: We have plans for introducing a namespaced `AdmissionPolicy` in the future, so an administrator can grant privileges at the namespace level that users can operate on, creating/modifying/deleting policies of their own I think we should not change the layout of namespaced/cluster-scoped just to get advantage of owner references.

rfc/0001-policy-server-custom-resource.md
flavio: Good point, solved

rfc/0001-policy-server-custom-resource.md
raulcabello: ah ok, thanks for the clarification

rfc/0001-policy-server-custom-resource.md
flavio: Fixed

rfc/0001-policy-server-custom-resource.md
flavio: I've added a note about that.

rfc/0001-policy-server-custom-resource.md
flavio: Resolved with another note

rfc/0001-policy-server-custom-resource.md
ereslibre: On a second thought: you are right. We can do it safely because we won't remove any webhook. Webhooks will fail to reach the service, and depending on their failure policy they will accept or reject.

Sounds good, no need to block the `PolicyServer` deletion as you have said.

flavio: I think we can discuss more in details what to do once we decide go go ahead :)

flavio:
@kkaempf I don't think we have a real alternative. We have the one mentioned into the "alternatives" section, but it has long term issues.

ereslibre:
Should we go ahead and merge this RFC? I think all of us agree on the design and outcomes of it.

viccuad:
Indeed :).

ereslibre:
Merged. Let's move this forward. We can always amend if something is not completely right.

Issue #49, opened 2021-08-11
Extend the `ClusterAdmissionPolicy` resource to allow binding to a `PolicyServer`


ereslibre:
The `ClusterAdmissionPolicy` resource can be bound to a given `PolicyServer` resource. This binding is done through an optional `policyServer` attribute inside the `ClusterAdmissionPolicy` resource. If provided, it has to match with an existing `PolicyServer` resource. If not provided, it will have a value of `default`.

**Acceptance criteria**
- Adapt the custom resource definition to have a new optional `policyServer` attribute.
- If the user does not provide anything on the `policyServer` attribute, the logic of the controller will default it to the `default` value.
- The new API version is `v1alpha3`.

PR #54, opened 2021-08-12
Extend the ClusterAdmissionPolicy resource to allow binding to a Policy Server

raulcabello:
Fixes kubewarden/kubewarden-controller#49


PROJECT
```diff
@@ -8,18 +8,18 @@ resources:
...
     namespaced: false
+  controller: true
   domain: kubewarden.io
   group: policies
-  kind: ClusterAdmissionPolicy
-  path: github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2
-  version: v1alpha2
+  kind: PolicyServer
+  path: github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha3
+  version: v1alpha3
```
flavio, line 16: I know we discussed that yesterday, but do we have to update the api version? Given the new field is optional, can't we just add it to the existing v1alpha2 policy?

Moreover, if we are forced to add a new version of the object, the old one must stay around for some time for compatibility reasons.

apis/policies/v1alpha3/clusteradmissionpolicy_types.go
```diff
@@ -23,6 +23,11 @@ import (
 
 // ClusterAdmissionPolicySpec defines the desired state of ClusterAdmissionPolicy
 type ClusterAdmissionPolicySpec struct {
+	// PolicyServer identifies an existing PolicyServer resource.
+	// +kubebuilder:default:=default
```
flavio, line 27: Nice, we don't have to do anything fancy to have `default` being written into this field :partying_face:

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
```diff
@@ -194,6 +194,10 @@ spec:
                       are ANDed.
                     type: object
                 type: object
+              policyServer:
+                default: default
+                description: PolicyServer identifies an existing PolicyServer resource.
+                type: string
```
ereslibre, line 200: So this is what `+kubebuilder:default:=default` is generating, nice. I wasn't sure if `+kubebuilder:default:=default` would require a webhook, but is really nice it doesn't and uses this API machinery feature :)

ereslibre (approved):
LGTM

flavio (changes requested):
I just skimmed through the PR, there are some comments at the beginning that influence the whole change

ereslibre (changes requested):
In this case, since this is a change that we didn't merge at all, I'd squash all commits into one, to keep history meaningful.

raulcabello:
@ereslibre yes, you are right. I was going to squash the 3 commits before merging in GH

flavio (approved):
LGTM, in this case I would squash the commits together

raulcabello:
done, thanks!

Issue #55, opened 2021-08-12
PolicyServer control loop: create root CA infrastructure


flavio:
The old controller creates a certificate authority and stores all its data (private key, pub cert) into a Kubernetes Secret.

This operation should now be performed by the reconciliation loop of `PolicyServer`, not of `ClusterAdmissionPolicy`.

## Admission criteria

* The `PolicyServer` control loop creates a Secret named `kubewarden-root-ca`
* The Secret contains the key + certificate of the root CA
* The Secret is not re-created if it already exists:
  * When running the kubewarden-controller for the 1st time: the secret name `kubewarden-root-ca` is created
  * When restarted, the kubewarden-controller doesn't overwrite the `kubewarden-root-ca` Secret
* The Secret resource is created into the same namespace where the controller runs. The name of the namespace is provided by the user through a parameter. The code doesn't have to infer that (this is already the current behaviour)

PR #60, opened 2021-08-12
PolicyServer control loop: create root CA infrastructure

raulcabello:
Fixes kubewarden/kubewarden-controller#55

### DIFF
--- a/internal/pkg/admission/policy-server-secret.go
+++ b/internal/pkg/admission/policy-server-secret.go
@@ -51,6 +51,9 @@ func (r *Reconciler) buildPolicyServerCARootSecret() (*corev1.Secret, error) {
 		return nil, fmt.Errorf("cannot generate policy-server secret CA: %w", err)
 	}
 	caPEMEncoded, err := pemEncodeCertificate(ca)
+	if err != nil {
+		return nil, fmt.Errorf("cannot encode policy-server secret CA: %w", err)
+	}
 	secretContents := map[string]string{
 		constants.PolicyServerCARootPemName:            string(caPEMEncoded),
 		constants.PolicyServerCARootPrivateKeyCertName: caPrivateKey.PrivateKey,
--- a/internal/pkg/admission/policy-server-secret.go
+++ b/internal/pkg/admission/policy-server-secret.go
@@ -22,7 +22,10 @@ func (r *Reconciler) reconcileSecret(ctx context.Context, secret *corev1.Secret)
 	return fmt.Errorf("error reconciling policy-server Secret: %w", err)
 }
 
-func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Context) (*corev1.Secret, error) {
+type generateCAFunc = func() ([]byte, *admissionregistration.KeyPair, error)
+type pemEncodeCertificateFunc = func(certificate []byte) ([]byte, error)
+
+func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Context, generateCA generateCAFunc, pemEncodeCertificate pemEncodeCertificateFunc) (*corev1.Secret, error) {
 	policyServerSecret := corev1.Secret{}
 	err := r.Client.Get(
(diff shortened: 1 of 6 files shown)


internal/pkg/admission/policy-server-secret.go
```diff
@@ -42,31 +42,22 @@ func (r *Reconciler) fetchOrInitializePolicyServerSecret(ctx context.Context) (*
 	return &policyServerSecret, nil
 }
 
-func (r *Reconciler) buildPolicyServerSecret() (*corev1.Secret, error) {
-	ca, caPrivateKey, err := admissionregistration.GenerateCA()
+var generateCA = admissionregistration.GenerateCA
+var pemEncodeCertificate = admissionregistration.PemEncodeCertificate
```
flavio, line 46: I didn't get it in the beginning, but then I saw the unit test :heart_eyes: :clap:

flavio (changes requested):
Looks great, there's just the comment I left about not handling an error

internal/pkg/admission/policy-server-secret.go
```diff
@@ -42,31 +42,22 @@ func (r *Reconciler) fetchOrInitializePolicyServerSecret(ctx context.Context) (*
...
-	}
-	servingCert, servingKey, err := admissionregistration.GenerateCert(
-		ca,
-		fmt.Sprintf("%s.%s.svc", constants.PolicyServerServiceName, r.DeploymentsNamespace),
-		[]string{fmt.Sprintf("%s.%s.svc", constants.PolicyServerServiceName, r.DeploymentsNamespace)},
-		caPrivateKey.Key())
-	if err != nil {
-		return nil, fmt.Errorf("cannot generate policy-server certificate: %w", err)
-	}
+	caPEMEncoded, err := pemEncodeCertificate(ca)
```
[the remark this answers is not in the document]
raulcabello: ups! you are totally right, I removed it by mistake. I'll add it in a new commit. Thanks!

internal/pkg/admission/policy-server-secret.go
ereslibre: It's nice. I would recommend passing a function as an argument to the functions that rely on this variable. This way there's no implicit hidden behavior that depends on a global package variable being set to different values.

internal/pkg/admission/policy-server-secret_test.go
```diff
@@ -0,0 +1,101 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.PolicyServerCARootSecretName,
+			Namespace: namespace,
+		},
+		StringData: mockSecretContents,
+		Type:       corev1.SecretTypeOpaque,
+	}
+
+	// Create a fake client to mock API calls. It will return the mock secret
+	cl := fake.NewClientBuilder().WithObjects(mockSecret).Build()
```
ereslibre, line 85: I wasn't aware of this feature! https://github.com/kubernetes-sigs/controller-runtime/commit/d915eb1a8c72db5587241cb612b4e901df20d45f :clap:

ereslibre (commented):
A small comment from my side regarding the usage of global variables to store function pointers. LGTM otherwise!

internal/pkg/admission/policy-server-secret.go
flavio: Yes, I agree on that. Another possibility is to make these variables part of the `Reconciler` struct.

One final note: please change their names to have the `Func` suffix, it makes it easier to understand what they are about

raulcabello: ok, I created package vars because in my opinion makes the code easier to read. But I get your point that someone else inside the package could potentially change the behavior. I changed it in the last commit, please have a look.
Thanks for the feedback!

ereslibre (approved):
LGTM, thanks @raulcabello!

flavio (approved):
I really like this new approach, you also defined types for these functions. This is more elegant :clap:

Issue #56, opened 2021-08-12
PolicyServer control loop: create certificates for Policy Server deployment


flavio:
Each Policy Server listens over https, the certificate used is provided via a Secret that is mounted by the Deployment definition.

This certificate is signed by the root CA as described [here](https://github.com/kubewarden/kubewarden-controller/issues/55).

## Admission criteria

- Each Policy Server instance has its own dedicated certificate. The certificate is signed by the CA created by kubewarden
- The key and the certificate are stored into a Secret. The Secret object is created by the `Policy Server` control loop
- The Secret is named using this convention: `policy-server-<name of the PolicyServer resource>`
- The Secret is created only once, when it doesn't exist.
- The Secret is created into the same namespace that contains the root CA Secret.

flavio:
Implemented

Issue #57, opened 2021-08-12
PolicyServer control loop: create Policy Server Deployment 


flavio:
Currently, the Policy Server Deployment is managed by the `ClusterAdmissionPolicy` control loop. This should now be handled by the `PolicyServer` control loop.

## Admission criteria

- The Deployment resource is managed by the `PolicyServer` control loop
- The attributes of the `PolicyServer` resource are used to create the final Deployment definition
- Changing one of the attributes of an existing PolicyServer resource, leads to a rollout of the Deployment associated with it. Eg: changing the `replicaSize` attribute inside of the `PolicyServer` object leads to an update of the already existing Deployment resource, which should then lead to a new Pods being created (or killed, based on the change done to the `replicaSize)
- The Deployment object is defined inside of the usual namespace
- The Deployment is still referencing a ConfigMap that holds the `policies.yml` file.
- The Deployment definition will keep having an annotation that specifies the version of the ConfigMap that was used when Deploying the resource

PR #61, opened 2021-08-13
PolicyServer control loop: create certificates for Policy Server deployment

raulcabello:
Fixes kubewarden/kubewarden-controller#56

I added an integration test that creates ca and policy server certs and validates them in a go https server. Right now it will also be executed with the unit tests when running go test ./... We can add a flag for the integration tests to run them separately as they take longer than the unit tests (this int test takes a bit more than 1 second in my machine). Maybe we don't need it now, but just keep it in mind for the future. What do you think?

### DIFF
--- a/internal/pkg/admission/mutating-webhook.go
+++ b/internal/pkg/admission/mutating-webhook.go
@@ -34,7 +34,7 @@ func (r *Reconciler) mutatingWebhookConfiguration(
 
 	service := admissionregistrationv1.ServiceReference{
 		Namespace: r.DeploymentsNamespace,
-		Name:      constants.PolicyServerServiceName,
+		Name:      constants.PolicyServerServiceNamePrefix,
 		Path:      &admissionPath,
 		Port:      &admissionPort,
 	}
--- a/internal/pkg/admission/policy-server-deployment.go
+++ b/internal/pkg/admission/policy-server-deployment.go
@@ -260,7 +260,7 @@ func (r *Reconciler) deployment(ctx context.Context, configMapVersion string) *a
 							Name: certsVolumeName,
 							VolumeSource: corev1.VolumeSource{
 								Secret: &corev1.SecretVolumeSource{
-									SecretName: constants.PolicyServerSecretName,
+									SecretName: constants.PolicyServerSecretNamePrefix,
 								},
 							},
 						},
--- a/internal/pkg/admission/policy-server-secret.go
+++ b/internal/pkg/admission/policy-server-secret.go
@@ -2,6 +2,8 @@ package admission
 
 import (
 	"context"
+	"crypto/rsa"
+	"crypto/x509"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
@@ -13,6 +15,10 @@ import (
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 )
 
+type generateCAFunc = func() (*admissionregistration.CA, error)
+type pemEncodeCertificateFunc = func(certificate []byte) ([]byte, error)
+type generateCertFunc = func(ca []byte, commonName string, extraSANs []string, CAPrivateKey *rsa.PrivateKey) ([]byte, []byte, error)
+
 func (r *Reconciler) reconcileSecret(ctx context.Context, secret *corev1.Secret) error {
 	err := r.Client.Create(ctx, secret)
 	if err == nil || apierrors.IsAlreadyExists(err) {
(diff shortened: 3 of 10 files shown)


internal/pkg/admission/mutating-webhook.go
```diff
@@ -34,7 +34,7 @@ func (r *Reconciler) mutatingWebhookConfiguration(
 
 	service := admissionregistrationv1.ServiceReference{
 		Namespace: r.DeploymentsNamespace,
-		Name:      constants.PolicyServerServiceName,
+		Name:      constants.PolicyServerServiceNamePrefix,
```
flavio, line 37: This is missing the name of this specific PolicyServer resource. It should be something like

```golang
Name: constants.PolicyServerServiceNamePrefix + r.Name
```

I'm not sure about r.Name, I'm just giving you some pseudo code :)

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -260,7 +260,7 @@ func (r *Reconciler) deployment(ctx context.Context, configMapVersion string) *a
 							Name: certsVolumeName,
 							VolumeSource: corev1.VolumeSource{
 								Secret: &corev1.SecretVolumeSource{
-									SecretName: constants.PolicyServerSecretName,
+									SecretName: constants.PolicyServerSecretNamePrefix,
```
flavio, line 263: Some here, we had to add the name of this PolicyService instance

internal/pkg/admission/policy-server-secret.go
```diff
@@ -22,8 +28,66 @@ func (r *Reconciler) reconcileSecret(ctx context.Context, secret *corev1.Secret)
...
+	err := r.Client.Get(
+		ctx,
+		client.ObjectKey{
+			Namespace: r.DeploymentsNamespace,
+			Name:      constants.PolicyServerSecretNamePrefix + policyServerName},
+		&policyServerSecret)
+	if err != nil && apierrors.IsNotFound(err) {
+		return r.buildPolicyServerSecret(policyServerName, caSecret, generateCert)
+	}
+	policyServerSecret.ResourceVersion = ""
```
flavio, line 42: this line should be moved after the next error check

internal/pkg/admission/policy-server-secret.go
```diff
@@ -22,8 +28,66 @@ func (r *Reconciler) reconcileSecret(ctx context.Context, secret *corev1.Secret)
...
+	}
+
+	return &policyServerSecret, nil
+}
+
+func (r *Reconciler) buildPolicyServerSecret(policyServerName string, caSecret *corev1.Secret, generateCert generateCertFunc) (*corev1.Secret, error) {
+	ca, err := extractCaFromSecret(caSecret)
+	servingCert, servingKey, err := generateCert(
+		ca.CaCert,
+		fmt.Sprintf("%s.%s.svc", constants.PolicyServerServiceNamePrefix+policyServerName, r.DeploymentsNamespace),
```
flavio, line 55: Maybe we can move the `constants.PolicyServerServiceNamePrefix+policyServerName` into a dedicated helper function and just call it from the different parts of the code that are already using this code.

This makes everything more consistent and easier to change in the future, just my 2 cents

internal/pkg/admission/reconciler.go
```diff
@@ -135,6 +135,25 @@ func (r *Reconciler) Reconcile(
 		policiesv1alpha2.PolicyServerSecretReconciled,
 	)
 
+	policyServerSecret, err := r.fetchOrInitializePolicyServerSecret(ctx, "psp-policies", policyServerCARootSecret, admissionregistration.GenerateCert)
```
flavio, line 138: The name of the secret here is hard-coded to something you used for testing :)

internal/pkg/admission/validating-webhook.go
```diff
@@ -34,7 +34,7 @@ func (r *Reconciler) validatingWebhookConfiguration(
 
 	service := admissionregistrationv1.ServiceReference{
 		Namespace: r.DeploymentsNamespace,
-		Name:      constants.PolicyServerServiceName,
+		Name:      constants.PolicyServerServiceNamePrefix,
```
flavio, line 37: The name of the service is wrong, should also include the name of this instance of PolicyServer

internal/pkg/admission/policy-server-service.go
```diff
@@ -22,7 +22,7 @@ func (r *Reconciler) reconcilePolicyServerService(ctx context.Context) error {
 func (r *Reconciler) service() *corev1.Service {
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      constants.PolicyServerServiceName,
+			Name:      constants.PolicyServerServiceNamePrefix,
```
flavio, line 25: This must include the name of this instance of PolicyServer

flavio (changes requested):
Overall good, there are some things that need to be fixed up

flavio:
@raulcabello this looks great, I love the integration test. I would leave that running right now, I don't mind the extra time it takes.

I think we should create a new issue to ensure we have a GH Action that runs the tests. So far we don't have it because setting up the whole kubebuilder env was a bit tricky... Can you file an issue for that maybe?

internal/pkg/admission/mutating-webhook.go
raulcabello: I was thinking of doing this in the next card https://github.com/kubewarden/kubewarden-controller/issues/57 
Since this is still done in the ClusterAdmissionPolicy loop and we won't have the policy name until we move the code to the PolicyServer loop. 
Maybe I shouldn't have introduced the prefix here. Should I leave as it was and then do it in the PolicyServer card?
Same applies to the other comments

internal/pkg/admission/policy-server-secret.go
raulcabello: yes, you are right. Thanks for finding this. For the other comments see my reply to the first one

internal/pkg/admission/mutating-webhook.go
flavio: Sorry, I didn't realize that. You can leave that as it is, and then add the suffix into another PR

raulcabello:
I just created a new issue for running the tests in github. I think it should be run when doing a PR and notify if any tests fail.
https://github.com/kubewarden/kubewarden-controller/issues/62 feel free to modify it

internal/pkg/admission/policy-server-secret.go
raulcabello: see next commit

flavio (approved):
LGTM, please address the naming concerns asap with another PR

Issue #58, opened 2021-08-12
PolicyServer control loop: create Service for the Policy Server Deployment 


flavio:
Each Policy Server Deployment is exposed via an in-cluster Service (`ClusterIP` type). This is how the Kubernetes API server reaches out to the webhook server.

This Service should now be handled by the control loop of `PolicyServer`.

## Admission criteria

- There's one Service object per PolicyServer resource
- The name of the Service is following this naming schema: `policy-server-<name of the PolicyServer resource`
- The Service resource is created into the usual namespace

PR #64, opened 2021-08-18
 PolicyServer control loop: create Policy Server Service

raulcabello:
Fixes kubewarden/kubewarden-controller#58

### DIFF
--- a/internal/pkg/admission/policy-server-deployment.go
+++ b/internal/pkg/admission/policy-server-deployment.go
@@ -205,19 +205,25 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      policyServer.NameWithPrefix(),
 			Namespace: r.DeploymentsNamespace,
-			Labels:    policyServer.AppLabel(),
+			Labels: map[string]string{
+				constants.AppLabelKey: policyServer.AppLabel(),
+			},
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: &policyServer.Spec.Replicas,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: policyServer.AppLabel(),
+				MatchLabels: map[string]string{
+					constants.AppLabelKey: policyServer.AppLabel(),
+				},
 			},
 			Strategy: appsv1.DeploymentStrategy{
 				Type: appsv1.RollingUpdateDeploymentStrategyType,
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels:      policyServer.AppLabel(),
+					Labels: map[string]string{
+						constants.AppLabelKey: policyServer.AppLabel(),
+					},
 					Annotations: templateAnnotations,
 				},
 				Spec: corev1.PodSpec{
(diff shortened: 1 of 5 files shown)


internal/pkg/admission/policy-server-deployment.go
```diff
@@ -205,19 +205,19 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      policyServer.NameWithPrefix(),
 			Namespace: r.DeploymentsNamespace,
-			Labels:    constants.PolicyServerLabels,
+			Labels:    policyServer.AppLabel(),
```
ereslibre, line 208: This is probably a personal preference, I'd do something like:

```suggestion
			Labels:    map[string]string{
			  constants.AppLabel: policyServer.AppLabel(),
			}
```

The `AppLabel` function would return the `string` only. I think is easier to add more labels if needed in the future and to keep them separate, but this is just bike shedding :)

internal/pkg/admission/reconciler.go
```diff
@@ -65,7 +65,19 @@ func (r *Reconciler) ReconcileDeletion(
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      policyServer.NameWithPrefix(),
+			Namespace: r.DeploymentsNamespace,
+		},
+	}
+	err = r.Client.Delete(ctx, service)
+	if err != nil {
+		r.Log.Error(err, "ReconcileDeletion: cannot delete PolicyServer Service %s", policyServer.Name)
+		errors = append(errors, err)
+	}
```
ereslibre, line 78: On deletion I think we should be cautious with erroring out on all cases. For example, I would not return if the error is a 404 (`IsNotFound`) of "children" resources. When reconciling deletions we want to delete as much as possible, and if an issue happens at the middle of the deletion reconciliation, but the error is that the resource being deleted is no longer persisted, is fine to keep going to the next step.

ereslibre: To sum up: I would not accumulate `IsNotFound` errors and just assume that is good. An extreme case: if the reconciliation finishes and all it did was error on every step because all resources are giving a 404, then we are done reconciling, and that is fine.

internal/pkg/admission/reconciler.go
```diff
@@ -175,7 +187,16 @@ func (r *Reconciler) Reconcile(
...
-	// TODO reconcile service and webhook
+
+	if err := r.reconcilePolicyServerService(ctx, policyServer); err != nil {
+		setFalseConditionType(
+			&policyServer.Status.Conditions,
+			policiesv1alpha2.PolicyServerServiceReconciled,
+			fmt.Sprintf("error reconciling service: %v", err),
+		)
+		return err
+	}
```
ereslibre, line 198: We are missing the `SetTrueConditionType` that sets the service reconciliation condition to `true`.

ereslibre (changes requested):
Thanks @raulcabello! Some comments :)

internal/pkg/admission/reconciler.go
raulcabello: yes, I totally missed that. Will be added in the next commit. Thanks!

internal/pkg/admission/policy-server-deployment.go
raulcabello: make sense, that way it would be easier to add labels in the future. I'll change it in the next commit

internal/pkg/admission/reconciler.go
raulcabello: yes, you are right. I'll check if the error is different that IsNotFound in the next commit. Someone could even manually remove a resource created by the operator, then delete the PolicyServer. In that case the reconciliation would fail because the resource was not there.

ereslibre (approved):
LGTM

flavio (approved):
LGTM

Issue #59, opened 2021-08-12
PolicyServer control loop: watch ClusterAdmissionPolicies 


flavio:
We can get rid of the control loop monitoring `ClusterAdmissionPolicies`. Events happening on these resources are going to be translated into `PolicyServer` events.

The `PolicyServer` controller will watch also `ClusterAdmissionPolicies`. Every event happening against a ClusterAdmissionPolicy will trigger a reconciliation loop of the `PolicyServer` that owns that resource.

The `PolicyServer` will then, inside of its regular reconciliation loop, query the Kubernetes API to obtain the list of all the `ClusterAdmissionPolicies` that it owns.

The list of current `ClusterAdmissionPolicies` owned by the `PolicyServer` is then compared with the contents of the `policies.yml` defined inside of the ConfigMap.

If a policy was **added**:
  * The ConfigMap object is updated
  * A new rollout of the Deployment is done
  * A new Validation/Mutation webhook object is created

If a policy was **updated**:
  * The ConfigMap object is updated
  * A new rollout of the Deployment is done
  * The Validation/Mutation webhook object must be reconciled

If a policy was **deleted**:
  * The controller computes the name of the Validation/Mutation webhook configuration that was referencing this policy
  * The controller deletes the Validation/Mutation webhook configuration object
  * The ConfigMap object is updated
  * A new rollout of the Deployment is done

About the Validation/Mutation webhook configuration: their names are still calculated starting from the name of the `ClusterAdmissionPolicy`. This behaviour is not going to change with the new architecture

## Acceptance  Criteria

- The control loop of `PolicyServer` is watching also the `ClusterAdmissionPolicy` objects
- Each `PolicyServer` has its own ConfigMap named `policy-server-<name of the PolicyServer resource>`
- The ConfigMap has a key named `policies.yml` which contains the list of policies to load
- The contents of `policies.yml` are calculated by PolicyServer, listing all the existing `ClusterAdmissionPolicies` objects its owns
- When a user creates a new ClusterAdmissionPolicy,
  - The ConfigMap is updated
  - The Deployment is rolled out
  - A new Validation/Mutation webhook configuration is created only after the deployment rollout is successfully done
- When a user updates and existing ClusterAdmissionPolicy,
  - The ConfigMap is updated
  - The Deployment is rolled out
  - The associated Validation/Mutation webhook configuration is updated, but only after the deployment rollout is successfully done
- When a user deletes and existing ClusterAdmissionPolicy,
[...]

PR #66, opened 2021-08-20
PolicyServer control loop: watch ClusterAdmissionPolicies

raulcabello:
Fixes kubewarden/kubewarden-controller#59

### DIFF
--- a/controllers/policies/policyserver_controller.go
+++ b/controllers/policies/policyserver_controller.go
@@ -20,17 +20,21 @@ import (
 	"context"
 	"fmt"
 	"github.com/go-logr/logr"
+	policiesv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2"
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/admission"
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	"github.com/pkg/errors"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
-
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	policiesv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"time"
 )
 
 // PolicyServerReconciler reconciles a PolicyServer object
@@ -69,6 +73,13 @@ func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// Reconcile
 	if err := admissionReconciler.Reconcile(ctx, &policyServer); err != nil {
+		if admission.IsPolicyServerNotReady(err) {
+			log.Error(err, "delaying policy registration since policy server is not yet ready")
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second * 5,
+			}, nil
+		}
 		return ctrl.Result{}, fmt.Errorf("reconciliation error: %w", err)
 	}
 	if err := r.updatePolicyServerStatus(ctx, &policyServer); err != nil {
@@ -91,8 +102,54 @@ func (r *PolicyServerReconciler) updatePolicyServerStatus(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	err := mgr.GetFieldIndexer().IndexField(context.Background(), &policiesv1alpha2.ClusterAdmissionPolicy{}, constants.PolicyServerIndexKey, func(object client.Object) []string {
+		clusterAdmissionPolicy := object.(*policiesv1alpha2.ClusterAdmissionPolicy)
+		return []string{clusterAdmissionPolicy.Spec.PolicyServer}
+	})
+	if err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&policiesv1alpha2.PolicyServer{}).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		Watches(&source.Kind{Type: &policiesv1alpha2.ClusterAdmissionPolicy{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			policy, ok := object.(*policiesv1alpha2.ClusterAdmissionPolicy)
+			if !ok {
+				return []ctrl.Request{}
+			}
+			// TODO check policy if server exist. use owner references or webhook for validating
+			var originalPolicy policiesv1alpha2.ClusterAdmissionPolicy
+			err := r.Client.Get(context.Background(), client.ObjectKey{
+				Name: policy.Name,
+			}, &originalPolicy)
+			if apierrors.IsNotFound(err) {
+				r.deleteWebhooks(policy)
+			}
+			return []ctrl.Request{
+				{
+					NamespacedName: client.ObjectKey{
+						Name:      policy.Spec.PolicyServer,
+						Namespace: policy.Namespace,
+					},
+				},
+			}
+
+		})).
 		Complete(r)
+}
+
+func (r *PolicyServerReconciler) deleteWebhooks(policy *policiesv1alpha2.ClusterAdmissionPolicy) {
+	validatingWebhook := &admissionregistrationv1.ValidatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: policy.Name,
+		},
+	}
+	_ = r.Client.Delete(context.Background(), validatingWebhook)
+	mutatingWebhook := &admissionregistrationv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: policy.Name,
+		},
+	}
+	_ = r.Client.Delete(context.Background(), mutatingWebhook)
 }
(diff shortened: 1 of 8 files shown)


controllers/policies/policyserver_controller.go
```diff
@@ -91,8 +102,54 @@ func (r *PolicyServerReconciler) updatePolicyServerStatus(
...
+			Name: policy.Name,
+		},
+	}
+	_ = r.Client.Delete(context.Background(), validatingWebhook)
+	mutatingWebhook := &admissionregistrationv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: policy.Name,
+		},
+	}
+	_ = r.Client.Delete(context.Background(), mutatingWebhook)
```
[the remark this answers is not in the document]
ereslibre: Also, for keeping correct track of that we will have to issue only one `Delete` request depending on whether the policy us mutating or not.

internal/pkg/admission/policy-server-configmap.go
```diff
@@ -34,38 +25,55 @@ type policyServerConfigEntry struct {
 func (r *Reconciler) reconcilePolicyServerConfigMap(
 	ctx context.Context,
 	policyServer *policiesv1alpha2.PolicyServer,
-	operation policyServerConfigMapOperation,
+	clusterAdminPolicies *policiesv1alpha2.ClusterAdmissionPolicyList,
```
ereslibre, line 28:
```suggestion
	clusterAdmissionPolicies *policiesv1alpha2.ClusterAdmissionPolicyList,
```

Typo? And instances tied to this one.

internal/pkg/admission/validating-webhook.go
```diff
@@ -19,13 +21,43 @@ func (r *Reconciler) reconcileValidatingWebhookConfiguration(
...
 	policyServerName string) error {
-	err := r.Client.Create(ctx, r.validatingWebhookConfiguration(clusterAdmissionPolicy, admissionSecret, policyServerName))
-	if err == nil || apierrors.IsAlreadyExists(err) {
+	webhook := r.validatingWebhookConfiguration(clusterAdmissionPolicy, admissionSecret, policyServerName)
+	err := r.Client.Create(ctx, webhook)
+	if err == nil {
 		return nil
 	}
+	if apierrors.IsAlreadyExists(err) {
+		return r.updateValidatingWebhook(ctx, clusterAdmissionPolicy, webhook)
```
ereslibre, line 30: I wish this could be possible with a `createOrUpdate` kind of request instead of having to issue two requests, but seems the API does not allow this...

ereslibre (commented):
Looking great, thanks @raulcabello!

Some comments, I would also replace `admin` with `admission` on the occurences :)

controllers/policies/policyserver_controller.go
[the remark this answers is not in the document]
raulcabello: that's a good point, but this code will be changed in the finalizers card. So I will take this in mind when implementing deletion with finalizers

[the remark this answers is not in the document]
ereslibre: Sounds fantastic!

ereslibre (approved):
Approving, I will check the `new-architecture` branch doing some manual smoke testing next week.

Really nice work @raulcabello!

internal/pkg/admission/validating-webhook.go
raulcabello: yes, that would be great..

internal/pkg/admission/policy-server-configmap.go
raulcabello: yes it is a typo. I just made a commit. Thanks!

internal/pkg/admission/reconciler.go
```diff
@@ -161,7 +173,12 @@ func (r *Reconciler) Reconcile(
 		return err
 	}
 
-	if err := r.reconcilePolicyServerConfigMap(ctx, policyServer, AddPolicy); err != nil {
+	var clusterAdmissionPolicies policiesv1alpha2.ClusterAdmissionPolicyList
+	if err := r.Client.List(ctx, &clusterAdmissionPolicies, client.MatchingFields{constants.PolicyServerIndexKey: policyServer.Name}); err != nil {
+		return fmt.Errorf("cannot retrieve cluster admin policies: %w", err)
```
ereslibre, line 178: This one too :)

raulcabello: fixed! :)

internal/pkg/admission/policy-server-configmap.go
```diff
@@ -34,38 +25,55 @@ type policyServerConfigEntry struct {
...
 		return fmt.Errorf("cannot lookup policies ConfigMap: %w", err)
 	}
 
+	newPoliciesYML, err := r.createPoliciesYML(clusterAdmissionPolicies)
+	if err != nil {
+		return fmt.Errorf("cannot create policies: %w", err)
+	}
+	currentPoliciesYML := cfg.Data[constants.PolicyServerConfigPoliciesEntry]
+
+	if currentPoliciesYML != newPoliciesYML {
```
flavio, line 49: I fear this could return false positives.

Let's say `currentPoliciesYML` is:
```yaml
policy1:
  url: registry.com/foo:1.2
  settings:
     name: flavio
     greeting: ciao
policy2:
  url: registry.com/bar:2
```


and `newPoliciesYML` is:
```yaml
policy2:
  url: registry.com/bar:2
policy1:
  url: registry.com/foo:1.2
  settings:
     name: flavio
     greeting: ciao
```
or something like:
```yaml
policy1:
  url: registry.com/foo:1.2
  settings:
     greeting: ciao
     name: flavio
policy2:
  url: registry.com/bar:2
```

This simple `!=` check will report the policy contents changed.

I think there's a high chance the ordering of these two structures is going to be completely messed up :(

flavio (changes requested):
LGTM, I just left some comments.

I think we will have to come up with a different comparison mechanism

internal/pkg/admission/policy-server-configmap.go
flavio: Maybe we could make your life easier by doing a change inside of policy-server...

Right now the `policy-server` ConfigMap has a single key called `policies.yml` which has a json string as value.
What if we change that so that:
  * The ConfigMap has multiple keys, one per `ClusterAdmissionPolicy`
  * The name of the key is going to be: `<ClusterAdmissionPolicy.name>.yml`
  * The value of the key is going to be a real Yaml document, that holds the information about the policy. Yaml can perserve the order of keys, which is what we care about

The Deployment will then mount the whole ConfigMap under something like `/policy-server/policies.d`.

The directory will look like that:
```
/policy-server/
     |
     ` policies.d/
           `- policy1.yml
           `- policy2.yml
```

The contents of `policy1,yml` would be:

```yaml
url: registry.com/foo:1.2
  settings:
     name: flavio
     greeting: ciao
```

The contents of `policy2.yml` would be:

```yaml
policy2:
  url: registry.com/bar:2
```

We would then change policy-server to have a new flag, something like `--policies-dir` which would load all the single files in there, merge them as a single Yaml and then do the usual stuff.

Just saying.... maybe this makes the Go comparison more reliable

controllers/policies/policyserver_controller.go
```diff
@@ -91,8 +102,54 @@ func (r *PolicyServerReconciler) updatePolicyServerStatus(
...
+			if !ok {
+				return []ctrl.Request{}
+			}
+			// TODO check policy if server exist. use owner references or webhook for validating
+			var originalPolicy policiesv1alpha2.ClusterAdmissionPolicy
+			err := r.Client.Get(context.Background(), client.ObjectKey{
+				Name: policy.Name,
+			}, &originalPolicy)
+			if apierrors.IsNotFound(err) {
+				r.deleteWebhooks(policy)
```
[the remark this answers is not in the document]
raulcabello: yes, this code will removed with the finalizers

internal/pkg/admission/policy-server-configmap.go
ereslibre: If that structure feels better I think we can go ahead, but I would not change the structure to fix something that is not very handy here.

I also think that we can check if comparing marshaled jsons work, go should alphabetically sort keys, so they should be the same regardless of the persisted order.

raulcabello: you are right, I'll fix this

raulcabello: we can parse the existing configmap into a map[string]policyServerConfigEntry, we are already creating a map[string]policyServerConfigEntry for the new policies. Then we can iterate and compare each key in the other map, and also compare the length. I'll write the code for that and I'll add unit tests to make sure it works in all scenarios.

raulcabello: mm actually this will not work for settings. 
```
type policyServerConfigEntry struct {
	URL      string               `json:"url"`
	Settings runtime.RawExtension `json:"settings,omitempty"`
}
```
It will fail if the order in the settings is different

ereslibre: Looking again I think I was right... With `encoding/json` we should be fine. I'm going to confirm but I think if we unmarshal, marshal and compare the marshaled objects we are good.

ereslibre: Yes, check: https://pkg.go.dev/encoding/json

```
The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:

- keys of any string type are used directly
- encoding.TextMarshalers are marshaled
- integer keys are converted to strings
```

raulcabello:
Now we are comparing the map[string]policyServerConfigEntry with reflect.DeepEquals(). They will be the same when all the values of the keys matches regardless of the order. 
If the Settings order is different, it will think they are different. This is because Settings is runtime.RawExtension which uses a Raw []byte. But I don't see how Settings would be in a different order unless the user changes the order, since it is converted from string to []byte I think it should always have the same order.

internal/pkg/admission/policy-server-configmap.go
```diff
@@ -40,35 +41,60 @@ func (r *Reconciler) reconcilePolicyServerConfigMap(
...
 }
 
+func shouldUpdatePolicyMap(currentPoliciesYML string, newPoliciesMap map[string]policyServerConfigEntry) (bool, error) {
+	var currentPoliciesMap map[string]policyServerConfigEntry
+
+	if err := json.Unmarshal([]byte(currentPoliciesYML), &currentPoliciesMap); err != nil {
+		return false, fmt.Errorf("cannot unmarshal policies: %w", err)
+	}
+
+	return !reflect.DeepEqual(currentPoliciesMap, newPoliciesMap), nil
```
ereslibre, line 78: I like this solution a lot, thanks @raulcabello. I cannot foresee any problems at this time.

ereslibre (commented):
It's a LGTM modulo https://github.com/kubewarden/kubewarden-controller/pull/66#discussion_r692965590.

Issue #67, opened 2021-08-20
Add finalizers to our resources


flavio:
The following resources will have finalizers set:

*  PolicyServer
*  ClusterAdmissionPolicy

What happens when a ClusterAdmissionPolicy is deleted by the user:
* The object is not yet deleted by Kubernetes GC, because of our finalizer
* The reconciliation loop of the policy server that owns the policy should be executed
* The reconciliation loop handles that case as an update:
  * add/update policies that do not have the deletion timestamp set
  * for policies that have the deletion timestamp set:
    * Remove the Validation/MutationWebhook
    * Udate the policies.yaml configuration accordingly.
    * Remove the finalizer from the resource

What happens when a PolicyServer is deleted by the user:
* The object is not yet deleted by Kubernetes GC, because of our finalizer
* The reconciliation loop of the policy server is executed
* The reconciliation loop goes checks if the policy server owns policies:
  * If the policy server owns some policies:
     * Delete all the policies owned by the policy server
     * Leave the finalizer inside of the policy server object
     * Schedule another reconciliation of the policy server in X seconds
  * If the policy server doesn't own any policy:
    * Remove all the resources related with the policy server (deployment, secret, configmap, service)
    * Remove our finalizer

TODO: check if setting an ownerref between the PolicyServer and its namespaced resources (like deployment, secret, configmap, service) can lead the Kubernetes GC to remove these objects. This would simplify our code (we would not have to execute one of the bullet points above)

PR #68, opened 2021-08-30
Add finalizers to our resources

raulcabello:
Fixes kubewarden/kubewarden-controller#67

### DIFF
--- a/internal/pkg/admission/reconciler.go
+++ b/internal/pkg/admission/reconciler.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/go-logr/logr"
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/admissionregistration"
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"strings"
-
-	"github.com/go-logr/logr"
 
 	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -92,6 +92,13 @@ func (r *Reconciler) ReconcileDeletion(
 		errors = append(errors, err)
 	}
 
+	patch := policyServer.DeepCopy()
+	patch.ObjectMeta.Finalizers = nil
+	err = r.Client.Patch(ctx, patch, client.MergeFrom(policyServer))
+	if err != nil && !apierrors.IsNotFound(err) {
+		r.Log.Error(err, "ReconcileDeletion: cannot remove finalizer "+policyServer.Name)
+		errors = append(errors, err)
+	}
 	if len(errors) == 0 {
 		return nil
 	}
@@ -130,6 +137,16 @@ func (r *Reconciler) Reconcile(
 	ctx context.Context,
 	policyServer *policiesv1alpha2.PolicyServer,
 ) error {
+	clusterAdmissionPolicies, err := r.getClusterAdmissionPolicies(ctx, policyServer)
+	if err != nil {
+		return fmt.Errorf("cannot retrieve cluster admission policies: %w", err)
+	}
+
+	err = r.deletePendingClusterAdmissionPolicies(ctx, clusterAdmissionPolicies)
+	if err != nil {
+		return fmt.Errorf("cannot delete pending cluster admission policies: %w", err)
+	}
+
 	policyServerCARootSecret, err := r.fetchOrInitializePolicyServerCARootSecret(ctx, admissionregistration.GenerateCA, admissionregistration.PemEncodeCertificate)
 	if err != nil {
 		setFalseConditionType(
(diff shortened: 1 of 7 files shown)


raulcabello:
I added a script for removing finalizers in the Makefile as make undeploy was getting stuck and resources was never deleted.

ereslibre (changes requested):
Super great job @raulcabello. Flawless except for one detail I comment.

I did a super light smoke test manually. This is the patch I have used on top of this PR:

```diff
diff --git a/Makefile b/Makefile
index d970734..725ceeb 100644
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:preserveUnknownFields=false"
+# When executing the `run` target, use this namespace to create resources such as Deployments
+RUN_DEPLOYMENT_NAMESPACE ?= "kubewarden"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -67,7 +69,8 @@ build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+	kubectl create namespace ${RUN_DEPLOYMENT_NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
+	go run ./main.go --deployments-namespace ${RUN_DEPLOYMENT_NAMESPACE}
 
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
diff --git a/config/samples/policies_v1alpha2_clusteradmissionpolicy.yaml b/config/samples/policies_v1alpha2_clusteradmissionpolicy.yaml
index 54d2133..020f08a 100644
--- a/config/samples/policies_v1alpha2_clusteradmissionpolicy.yaml
+++ b/config/samples/policies_v1alpha2_clusteradmissionpolicy.yaml
@@ -2,8 +2,10 @@ apiVersion: policies.kubewarden.io/v1alpha2
 kind: PolicyServer
 metadata:
   name: reserved-instance-for-tenant-a
+  finalizers:
+  - kubewarden
 spec:
-  image: ghcr.io/kubewarden/policy-server:v1.0.0
+  image: ghcr.io/kubewarden/policy-server:v0.1.10
   replicas: 1
   env:
     - name: KUBEWARDEN_LOG_LEVEL
@@ -17,6 +19,8 @@ apiVersion: policies.kubewarden.io/v1alpha2
 kind: ClusterAdmissionPolicy
 metadata:
   name: psp-capabilities
+  finalizers:
+  - kubewarden
 spec:
   policyServer: reserved-instance-for-tenant-a
   module: registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.3
```

- I create a cluster with `k3d`
- I run `make install`
- I run `make run`
- I run `k apply -f config/samples/policies_v1alpha2_clusteradmissionpolicy.yaml`. I have also splitted this file so I have two `ClusterAdmissionPolicies` with different names that I could create/update/delete in an easier way.

I can confirm that:

- The root CA secret is created
- The policy server secret with the server certificate/key is created with the right contents
- The configmap for the policy server is created with the right contents
- The policy server deployment is present
- The policy server service is present
- The mutating webhook configuration is present

When creating a new ClusterAdmissionPolicy, the deploment is rolled out again, with the ConfigMap updated. The webhook is installed. This is correct.

When deleting a ClusterAdmissionPolicy, the deployment is rolled out again, with the ConfigMap updated. The webhook configuration is removed. This is correct too.

When deleting a PolicyServer, the configmap, secrets, deployment, service, and ClusterAdmissionPolicies related to that policy server are delete-cascaded flawlessly. This is working as expected.

One thing I saw though that is not behaving as expected: after creating a ClusterAdmissionPolicy, when the policy is active (by active I mean: the webhook is installed and the policy should be receiving requests), their `.status.active` is never set to true, so if you do a `kubectl get clusteradmissionpolicies` you never see a `true` in the policy row, despite the policy is active (the webhook is installed and pointing to the service, with the deployment up and running). We should patch the `.status.active` to true when we have reconciled the webhook. Output:

```
~ » k get clusteradmissionpolicy
NAME                MUTATING   ACTIVE
psp-capabilities    true
psp-capabilities2   false
```

internal/pkg/admission/reconciler.go
```diff
@@ -92,6 +92,13 @@ func (r *Reconciler) ReconcileDeletion(
 		errors = append(errors, err)
 	}
 
+	patch := policyServer.DeepCopy()
+	patch.ObjectMeta.Finalizers = nil
+	err = r.Client.Patch(ctx, patch, client.MergeFrom(policyServer))
```
ereslibre, line 97: We should only remove the finalizer we know about. This allows for other controllers to react if they add finalizers too, for whatever reason.

I would add a constant with a `kubewarden` finalizer for example. On this logic, we get the list of finalizers and remove `kubewarden` when we are done. If we are the only ones, the list of finalizers will be empty, letting the Kubernetes GC remove the resource. Also, when we implement the mutating webhook, we can add the finalizer on the resource if not present.

internal/pkg/admission/reconciler.go
```diff
@@ -274,3 +322,45 @@ func (r *Reconciler) enablePolicyWebhook(
...
+						Name: policy.Name,
+					},
+				}
+				err := r.Client.Delete(context.Background(), validatingWebhook)
+				if err != nil && !apierrors.IsNotFound(err) {
+					return err
+				}
+			}
+			patch := policy.DeepCopy()
+			patch.ObjectMeta.Finalizers = nil
```
ereslibre, line 358: Same here, I think we should only remove our finalizer, letting other finalizers alone in case there are.

internal/pkg/admission/reconciler.go
raulcabello: yes, great idea! done

ereslibre (approved):
Looks great. Thank you @raulcabello. I would go ahead and merge this work :)

Issue #71, opened 2021-09-08
Fix controller panic when policy server does not have any annotation


raulcabello:
If the policy server does not have any annotation, the controller panic  "assignment to entry in nil map" in this line https://github.com/kubewarden/kubewarden-controller/blob/new-architecture/internal/pkg/admission/policy-server-deployment.go#L202

We should check if the map is nil before the assignment

PR #73, opened 2021-09-08
Fix controller panic when policy server does not have any annotation

raulcabello:
Fixes kubewarden/kubewarden-controller#71


flavio (approved):
LGTM, good catch

Issue #69, opened 2021-09-08
Implement validation webhooks for `ClusterAdmissionPolicy`


flavio:
We must provide a mutation webhook for the `ClusterAdmissionPolicy` resource.

The webhook will implement the following validations:
* `UPDATE` operation: do not allow to change the name of the policy server where the policy is scheduled. If the user wants to change it he will have to delete the policy and create a new one. At least for now

The webhook will implement the following mutations:
* `CREATE` operation: ensure the kubewarden finalizers are specified, if not, add them
* `CREATE` operation: set the policyStatus to `pending` when a `ClusterAdmissionPolicy` is created

Note well: it's totally fine to create a `ClusterAdmissionPolicy` object that is scheduled on a non-existing `PolicyServer`.

raulcabello:
I just updated the card to set the policy status to unscheduled more info https://github.com/kubewarden/kubewarden-controller/issues/72

viccuad:
Closed by https://github.com/kubewarden/kubewarden-controller/pull/76.

Issue #72, opened 2021-09-08
`ClusterAdmissionPolicy` reconciliation loop: handle policies that refence non-existing PolicyServer


flavio:
It's possible for the user to create a `ClusterAdmissionPolicy` object that refers a non-existing PolicyServer.

When that happens, the status of the `ClusterAdmissionPolicy` resource should report this issue.

ereslibre:
I think we can take a similar approach to what CustomResourceDefinitions do. Our status field looks like this currently:

https://github.com/kubewarden/kubewarden-controller/blob/f765da1e465ee8a4557e2810941fce16f3be61d5/apis/policies/v1alpha2/clusteradmissionpolicy_types.go#L167-L184

I think we can add a condition type `PolicyServerBound` with status `"True"` or `"False"`. Also, I think we can add another status field called `policyServerBound` that is a boolean. This is false by default (possibly defaulted by a webhook), and only set to true by a reconciliation loop that ensures the bound policy server exists.

Conditions are good for using `kubectl wait`, but any third party using client-go or a similar client can watch the resource and wait until the `policy.status.policyServerBound` is true.

flavio:
Is there a way to replicate the UX the user have when they create a Pod that uses a non-existing volume mount? Something that says "pending..."

ereslibre:
Yes, this would be achieved inside the `status` struct with a `policyStatus` field that contains a string, and internally is filled from a go "enum". This `policyStatus` could have different valid values: `unscheduled`, `unschedulable`, `pending` (meaning - the policy server exists, reconciling all resources) and `active`. If we go this way, we can remove the `policyActive` bool from status, it would get superseded by this string field.

A webhook would ensure that every policy created starts with an `unscheduled` state for example. If no policy server exists with the given name, the reconciliation loop would set it to `unschedulable`. Otherwise, it would set it to `pending`. When the reconciliation loop ends for that policy, if everything was reconciled correctly, it can set `policyStatus` to `active`.

If we show this field on the table (doable with kubebuilder that relies on apimachinery), then when you do `kubectl get clusteradmissionpolicy` you would get the policy and the status as another column in the table.

raulcabello:
@ereslibre I like this approach. Should we update https://github.com/kubewarden/kubewarden-controller/issues/69 to set the `policyStatus` to `unscheduled` when a `clusteradmissionpolicy` is created?

ereslibre:
[...]

PR #76, opened 2021-09-16
Add `Finalizers`, `PolicyStatus`, `PolicyActive` condition to ClusterAdmissionPolicies

viccuad:
In this PR:

- Add defaulter, validating, mutating webhooks for ClusterAdmissionPolicies, namely for adding finalizers, and checking that the policyServer is immutable.
Given the difficulty on testing, added `apis/policies/v1alpha2/webhook_suite_test.go`.

- Refactor `ClusterAdmissionPolicyStatus.PolicyActive` status into `PolicyStatus`: now it accepts `unscheduled`, `unschedulable`, `pending`, `active`.
- Refactor `PolicyServerBound` Condition to `ClusterAdmissionPolicyActive`.
  Test with:
  ```
  kubectl wait --for=condition=active clusteradmissionpolicy psp-capabilities
  ```
 
- Enable Kustomize patches for cert-manager.
- Add RBAC marker for secrets and confimaps, needed to run/test in-cluster.
- Drop unneeded funcs, refactor as needed to be able to test.
 
Closes: https://github.com/kubewarden/kubewarden-controller/issues/69, https://github.com/kubewarden/kubewarden-controller/issues/72.

Thanks @ereslibre  & @raulcabello  for the pairing!

(It may be a good idea to review per commits, instead of by changes).  
 
---

To test with an out-of-cluster controller, one would need to create a proxy to the webhook server (TBD) since the webhooks are registered against the controller manager webhook server. To allow for the manager to at least start without faiiures, a new `make dev-webhook-certs` target has been added. It creates and sets the certs in localhost and gets called on `make run`. 
Therefore, to test meanwhile skipping the registration of webhooks yet having a working controller:
```console
$ kubectl create namespace kubewarden-controller-system
$ make install
$ make run ENABLE_WEBHOOKS=false NAMESPACE=kubewarden-controller-system
```
To test with an in-cluster controller, cert-manager has been scaffolded for creating the needed cert to register the webhooks. 
Example with k3d:
```console
$ k3d cluster create
$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
$ make docker-build NAMESPACE=kubewarden-controller-system
$ k3d image import controller:latest
$ kubectl create namespace kubewarden-controller-system
$ make deploy IMG=controller:latest NAMESPACE=kubewarden-controller-system
## patch deployment with ImagePullPolicy: Never
```


apis/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -157,21 +157,39 @@ const (
...
 
 // ClusterAdmissionPolicyStatus defines the observed state of ClusterAdmissionPolicy
 type ClusterAdmissionPolicyStatus struct {
-	// PolicyActive represents whether this AdmissionPolicy is active,
-	// such that the Kubernetes API server should be forwarding
-	// admission review objects to the policy
-	PolicyActive bool `json:"policyActive"`
+	// PolicyStatus represents whether this ClusterAdmissionPolicy is unscheduled,
+	// unschedulable, pending, or active.
+	// +kubebuilder:default:=unscheduled
```
raulcabello, line 187: we can remove default from here since it's not working

apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+func (r *ClusterAdmissionPolicy) ValidateCreate() error {
+	clusteradmissionpolicylog.Info("validate create", "name", r.Name)
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *ClusterAdmissionPolicy) ValidateUpdate(old runtime.Object) error {
+	clusteradmissionpolicylog.Info("validate update", "name", r.Name)
+
+	if r.Spec.PolicyServer != old.(*ClusterAdmissionPolicy).Spec.PolicyServer {
```
raulcabello, line 65: maybe we can check if the type assertion old.(*ClusterAdmissionPolicy) was ok, if not return an error.
```suggestion
	clusteradmissionpolicy, ok = old.(*ClusterAdmissionPolicy)
```

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
viccuad: I would prefer not to, it ends up in the CRD, and then users are aware that policies start in that state (even if it is set by the controller).

controllers/policies/policyserver_controller.go
```diff
@@ -46,6 +47,7 @@ type PolicyServerReconciler struct {
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=policyservers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=policyservers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=policyservers/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=secrets;configmaps,verbs=get;list;watch;create;update;patch
```
raulcabello, line 51: Can we add patch and watch permissions for mutating and validating webhooks? https://github.com/kubewarden/helm-charts/pull/29/commits/e6e8a67d374f4ffaa0d412a61e7c4eda5110cd27
then I think we have the same rbac in the helm chart and in our local env

raulcabello (changes requested):
LGTM just some minor suggestions 
Thanks @viccuad !

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -157,21 +157,39 @@ const (
...
+	PolicyStatus ClusterAdmissionPolicyStatusEnum `json:"policyStatus"`
 	// Conditions represent the observed conditions of the
 	// ClusterAdmissionPolicy resource.  Known .status.conditions.types
 	// are: "PolicyServerSecretReconciled",
+	// "PolicyServerBound",
```
viccuad, line 192: This is a leftover, there shouldn't be a `PolicyServerBound` and there should be a `ClusterAdmissionPolicyActive`.
This comment seems to be outdated, or I don't understand why policies can have conditions that are for the PolicyServer.

viccuad: Fixed in https://github.com/kubewarden/kubewarden-controller/pull/76/commits/eb9fe88174d48da76d048841f63a8c820149550d.

controllers/policies/policyserver_controller.go
viccuad: That sounds like a great idea!
Thanks for checking with the Helm chart. Do'h, when I checked, I checked against main and not new-architecture. I was super puzzled..

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
ereslibre: I also think is misleading. A `kubebuilder:default` marker has a very specific behavior that one can take for granted. Given this is not having the expected behavior (the api machinery is not really persisting a defaulted status field) I think is better to not include it here. We are using our own means for "defaulting" it on the reconciliation phase, but that has no relationship with this marker.

viccuad: Fair enough then :).
Reverted commit in https://github.com/kubewarden/kubewarden-controller/pull/76/commits/84a55a2754bd4b53cca50d942b7245547b52d6b6.

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
```diff
@@ -286,7 +286,7 @@ spec:
               conditions:
                 description: 'Conditions represent the observed conditions of the
                   ClusterAdmissionPolicy resource.  Known .status.conditions.types
-                  are: "PolicyServerSecretReconciled", "PolicyServerConfigMapReconciled",
+                  are: "PolicyServerSecretReconciled", "PolicyServerBound", "PolicyServerConfigMapReconciled",
                   "PolicyServerDeploymentReconciled", "PolicyServerServiceReconciled"
                   and "PolicyServerWebhookRegistered"'
```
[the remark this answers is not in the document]
ereslibre: Feel free to ignore this huge comment, I'd be good to go ahead with the current PR and then I'd propose a change on top of this, I don't want to block this work.

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -157,30 +157,47 @@ const (
...
+	// PolicyStatus represents whether this ClusterAdmissionPolicy is unscheduled,
+	// unschedulable, pending, or active.
+	PolicyStatus ClusterAdmissionPolicyStatusEnum `json:"policyStatus"`
 	// Conditions represent the observed conditions of the
 	// ClusterAdmissionPolicy resource.  Known .status.conditions.types
 	// are: "PolicyServerSecretReconciled",
 	// "PolicyServerConfigMapReconciled",
 	// "PolicyServerDeploymentReconciled",
 	// "PolicyServerServiceReconciled" and
 	// "PolicyServerWebhookRegistered"
```
ereslibre, line 194: Leftover

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
```diff
@@ -288,7 +288,7 @@ spec:
                   ClusterAdmissionPolicy resource.  Known .status.conditions.types
                   are: "PolicyServerSecretReconciled", "PolicyServerConfigMapReconciled",
                   "PolicyServerDeploymentReconciled", "PolicyServerServiceReconciled"
-                  and "PolicyServerWebhookRegistered"'
+                  and "PolicyServerWebhookRegistered" "ClusterAdmissionPolicyActive"'
```
ereslibre, line 291: `PolicyServerWebhookRegistered` leftover

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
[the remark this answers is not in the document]
raulcabello: Currently we are watching changes in the ClusterAdmissionPolicy and that triggers a PolicyServer reconciliation loop if the ClusterAdmissionPolicy references an existing PolicyServer. Then in the PolicyServer reconciliation loop we take care of all the steps you mentioned. We also set the condition for the ClusterAdmissionPolicy, but I checked and we just set the PolicyServerWebhookConfigurationReconciled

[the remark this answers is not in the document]
viccuad: fixed in https://github.com/viccuad/kubewarden-controller/commit/eb9fe88174d48da76d048841f63a8c820149550d.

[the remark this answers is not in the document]
raulcabello: PolicyServerWebhookConfigurationReconciled is now called ClusterAdmissionPolicyActive

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
viccuad: Indeed, missed it before. Fixed it in fixed in https://github.com/kubewarden/kubewarden-controller/pull/76/commits/bded155b91ac3a063fd892ff69e0d0a04f499852.

controllers/policies/policyserver_controller.go
viccuad: Done in https://github.com/viccuad/kubewarden-controller/commit/ce1df17001656bb972c9c4a943958a86e0b08435.

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
raulcabello: I think we just set the ClusterAdmissionPolicyActive condition in the ClusterAdmissionPolicy, I am going to check it in my local env

controllers/policies/policyserver_controller.go
raulcabello: great! thanks!

apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
viccuad: true! Done in https://github.com/viccuad/kubewarden-controller/commit/b975030d4f7e0f1e234b1336711a35655539ea04.

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -157,21 +157,39 @@ const (
...
 // ClusterAdmissionPolicyStatus defines the observed state of ClusterAdmissionPolicy
 type ClusterAdmissionPolicyStatus struct {
-	// PolicyActive represents whether this AdmissionPolicy is active,
-	// such that the Kubernetes API server should be forwarding
-	// admission review objects to the policy
-	PolicyActive bool `json:"policyActive"`
+	// PolicyStatus represents whether this ClusterAdmissionPolicy is unscheduled,
+	// unschedulable, pending, or active.
+	// +kubebuilder:default:=unscheduled
+	PolicyStatus ClusterAdmissionPolicyStatusEnum `json:"policyStatus"`
```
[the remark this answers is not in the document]
viccuad: This means that it will not be required, to change the status. Is that what we want :thinking:?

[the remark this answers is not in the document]
ereslibre: This does not say whether this argument is required or not. This is information for the Go marshaler: from the docs (https://pkg.go.dev/encoding/json):

```
The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string. 
```

I think is a good practice to marshal only what's relevant, and if that doesn't cause issues (on some circumstances you want to say the difference between a non-set field and a defaulted field, to take different actions).

raulcabello:
@viccuad I can't create policies without settings, I get The ClusterAdmissionPolicy "privileged-pods" is invalid: spec.settings: Invalid value: "null": spec.settings in body must be of type object: "null"
But if I remove the mutating webhook then I can create them

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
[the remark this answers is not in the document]
viccuad: Adding the omitempty will remove it as required for updating the status:

https://github.com/kubewarden/kubewarden-controller/blob/b975030d4f7e0f1e234b1336711a35655539ea04/config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml#L372-L373

Are we ok with that?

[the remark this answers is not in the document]
ereslibre: Right, didn't know controller-gen would do this automatically. Let's keep it as it is.

ereslibre (approved):
Thanks @viccuad!, really great work, we learnt quite a lot of things in the process 👏

apis/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -190,7 +206,7 @@ type ClusterAdmissionPolicyStatus struct {
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:storageversion
 //+kubebuilder:printcolumn:name="Mutating",type=boolean,JSONPath=`.spec.mutating`,description="Whether the policy is mutating"
-//+kubebuilder:printcolumn:name="Active",type=boolean,JSONPath=`.status.policyActive`,description="Whether the policy is active and receiving admission reviews"
+//+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.policyStatus`,description="Status of the policy"
 type ClusterAdmissionPolicy struct {
```
viccuad, line 210: (Opening comment thread here as I can't open on the specific code, as it is unchanged)

From @raulcabello:
> @viccuad I can't create policies without settings, I get The ClusterAdmissionPolicy "privileged-pods" is invalid: spec.settings: Invalid value: "null": spec.settings in body must be of type object: "null"
But if I remove the mutating webhook then I can create them

I can reproduce it, and I don't know from where it comes from. The policy spec is unchaged, and settings is marked as `optional` and `omitempty`.

viccuad: I don't see why the webhook would reject it, here I see it as approved:
```
2021-09-17T15:37:26.369Z    DEBUG    controller-runtime.webhook.webhooks    received request    {"webhook": "/mutate-policies-kubewarden-io-v1alpha2-clusteradmissionpolicy", "UID": "9d210836-8494-4835-999b-c3d89fa48604", "kind": "policies.kubewarden.io/v1alpha2, Kind=ClusterAdmission
Policy", "resource": {"group":"policies.kubewarden.io","version":"v1alpha2","resource":"clusteradmissionpolicies"}}                                                                                                                                                                         
2021-09-17T15:37:26.369Z    INFO    clusteradmissionpolicy-resource    default    {"name": "psp-capabilities-no-settings"}                                                                                                                                                                  
2021-09-17T15:37:26.369Z    DEBUG    controller-runtime.webhook.webhooks    wrote response    {"webhook": "/mutate-policies-kubewarden-io-v1alpha2-clusteradmissionpolicy", "code": 200, "reason": "", "UID": "9d210836-8494-4835-999b-c3d89fa48604", "allowed": true}
```

viccuad: marking .spec.settings as `//+nullable` makes it work. but I don't see why is it needed, given that is marked as `//+optional` already.

raulcabello: yes with `//+nullable` works! It make sense to add nullable here. You can have an optional field with a default value, so it would never be null

viccuad: it seems that in 1.20 (where I am), objects with empty values get pruned automatically if they don't have `nullable: true` (see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable).

Committed it in https://github.com/kubewarden/kubewarden-controller/pull/76/commits/5c01067db40feadbb6f1dca16b4430ad4457ba2d.

raulcabello: Thanks! @viccuad great work!

raulcabello (approved):

Issue #62, opened 2021-08-16
Github action to run tests


raulcabello:
Add a github action that run unit and integration tests on pull requests. Notify if tests fail.

## Aceptance criteria

- Unit and integration tests should run when doing a pull request
- Team should be notified if tests fail.

PR #78, opened 2021-09-23
Run tests for all branches in github actions 

raulcabello:
fixes  kubewarden/kubewarden-controller#62

This is how it will look like after adding the status check: https://github.com/raulcabello/kubewarden-controller/pull/1 

I'll squash the commits before merging


flavio (approved):
squash them and then merge it :rocket:

PR #83, opened 2021-09-27
Rename condition from `active` to `PolicyActive`

ereslibre:
This name is more in line to how other conditions are named in other core API resources.


viccuad (approved):
Indeed, way better :).

raulcabello (approved):
it looks better!

PR #85, opened 2021-09-27
Add uninstall script

flavio:
This script can be used as a last resort when `helm uninstall` doesn't work

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,15 @@
+name: ShellCheck
+on:
+  push:
+    paths:
+      - "scripts/**"
+  pull_request:
+    paths:
+      - "scripts/**"
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - run: shellcheck ./scripts/*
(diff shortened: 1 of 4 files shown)


viccuad:
Maybe is a good idea to give it a pass with `shellcheck`. I would be glad to have a `shellcheck ./*` on the make targets, or a GHA workflow too ([example](https://github.com/rancher-sandbox/hypper/blob/main/.github/workflows/shellcheck.yml)). But maybe it's too much, given that we don't have so many.

flavio:
@viccuad I've moved the script into a different location, fixed the shellcheck warnings and added a GH action too

Thanks for the tip

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    paths:
+      - "scripts/**"
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - run: shellcheck ./scripts/*
```
raulcabello, line 15: github check fails because of the webhook-cert directory https://github.com/kubewarden/kubewarden-controller/pull/85/checks?check_run_id=3721380603

raulcabello (commented):
LGTM, the uninstall script is really good. Great job!
github check fails because of the webhook-cert subdirectory

ereslibre (approved):

.github/workflows/shellcheck.yml
flavio: This should be fixed now

PR #86, opened 2021-09-28
POC: kube webhook wrapper

ereslibre:
This POC makes it very easy to develop a controller that has webhooks. Despite it's possible to test everything in isolation in the default Kubebuilder setup, this change takes advantage of https://github.com/ereslibre/kube-webhook-wrapper to develop your controller locally, while a Kubernetes cluster is running in a containerized environment, such as Kind or K3d.

The wrapper will automatically register the webhooks in the server (as Kubebuilder does), and it will also automatically create the required certificate for the web server, and register the webhooks with the `kubeconfig` file that is in the environment. All this automation only happens if a special envvar `KUBEWARDEN_DEVELOPMENT_MODE` is set to something other than `0` or `false`. If this envvar is not present, or if it's set to `0` or `false`, the behavior will be exactly the same as if the code generated by Kubebuilder was used.

Examples how to test it:

```
$ KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_HOST=172.28.0.1 make run
```

```
$ KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_HOST=172.28.0.1 go run . -deployments-namespace kubewarden
```

This is a PR to ask for feedback, and if we are fine with the idea, create a final PR.


raulcabello (approved):
I love this! Running in local makes development much easier! 
I just tried on minikube and it works fine! I just had to change a few things:
- add host.minikube.internal in https://github.com/ereslibre/kube-webhook-wrapper/blob/main/webhookwrapper/mod.go#L126
- add host.minikube.internal in https://github.com/ereslibre/kube-webhook-wrapper/blob/main/webhookwrapper/certificates.go#L74

It should be fine to leave host.minikube.internal in https://github.com/ereslibre/kube-webhook-wrapper/blob/main/webhookwrapper/certificates.go#L74 (we can have both localhost and minikube), but maybe we can add an optional parameter for https://github.com/ereslibre/kube-webhook-wrapper/blob/main/webhookwrapper/mod.go#L126 to support minikube? If this new parameter is not there, it will use the host as it is doing now.

minikube info https://minikube.sigs.k8s.io/docs/handbook/host-access/

Thanks! :)

ereslibre:
@raulcabello thanks for the feedback! How are you registering the webhooks? I'd like to know a little more about that setup.

raulcabello:
@ereslibre This what I did:
- make deploy. This installs everything on my local cluster: CRDs, webhooks, RBAC, ..
- manually remove webhooks created by make deploy. We can improve this :) 
- add host.minikube.internal in the go code as I mentioned in my previous comment
- execute KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_HOST=0.0.0.0 go run . -deployments-namespace kubewarden. This creates certificates and webhooks
Then I created a PolicyServer and the controller webhook running outside kubernetes was successfully called

Let me know if you want to know anything else, or I can show it to you in a call if you want

ereslibre:
@raulcabello With the latest change I think this use case is covered.

One thing is where the server is listening in your host, and the other is how this is "advertised" to the API server. The certificate needs to have a SAN entry for the "advertised" host, whether it's an IP address or a host -- both are supported.

If only one of those is provided, the other will be defaulted to the provided one. So, for example, with `k3d` I can execute:

- `KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_LISTEN=172.28.0.1 go run . -deployments-namespace kubewarden`: this means that the server is both listening and advertised as `172.28.0.1`.
- `KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_ADVERTISE=172.28.0.1 go run . -deployments-namespace kubewarden`: this means that the server is both listening and advertised as `172.28.0.1`.

In your environment, now you could run it like:

- `KUBEWARDEN_DEVELOPMENT_MODE=1 WEBHOOK_LISTEN=0.0.0.0 WEBHOOK_ADVERTISE=host.minikube.internal go run . -deployments-namespace kubewarden`. The certificate will only contain `host.minikube.internal` as a DNS SAN entry, that is what the API server will validate.

Let me know if it works for you :)

raulcabello:
@ereslibre yeah! it works perfectly! Thanks!

ereslibre:
I opened https://github.com/kubewarden/kubewarden-controller/pull/92

Issue #89, opened 2021-09-28
Fix flaky test


raulcabello:
TestCAAndCertificateCreationInAHttpsServer fails sometimes - > https://github.com/kubewarden/kubewarden-controller/runs/3732530549#step:4:59 

There is a race condition in the TestCAAndCertificateCreationInAHttpsServer test. It works fine most of the times, but sometimes https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/policy-server_integration_test.go#L59 will get called before the server is ready https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/policy-server_integration_test.go#L51

PR #90, opened 2021-09-28
Fix flaky test TestCAAndCertificateCreationInAHttpsServer

raulcabello:
fixes kubewarden/kubewarden-controller#89

### DIFF
--- a/internal/pkg/admission/policy-server_integration_test.go
+++ b/internal/pkg/admission/policy-server_integration_test.go
@@ -9,10 +9,14 @@ import (
 	"net/http"
 	"sync"
 	"testing"
+	"time"
 )
+
+const port = "8181"
 
 func TestCAAndCertificateCreationInAHttpsServer(t *testing.T) {
 	const domain = "localhost"
+	const maxRetries = 10
 	r := createReconcilerWithEmptyClient()
 	// create CA
 	caSecret, err := r.buildPolicyServerCARootSecret(admissionregistration.GenerateCA, admissionregistration.PemEncodeCertificate)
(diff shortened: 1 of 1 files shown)


viccuad (approved):
Nice! and I see you have fixed a couple of golangci-lint errors in the way :sweat_smile: .

flavio (approved):
LGTM

ereslibre:
I requested a review here because I wanted to comment two things:

I don't understand the purpose of the test, or what we are testing. I think it's safe to assume that something as core as this is already tested extensively in the go stdlib itself. Please, let me know if I'm missing something.

The other thing that caught my eye was the use of sleeping and retries. I think it would be good to switch to a `Listener` + `ServeTLS` combo if we want to do it, but I would question the need for this test at all.

raulcabello:
@ereslibre This is an integration test. We are testing that we create a secret that contains a valid certificate and a valid CA. In order to verify the certificate and CA is valid, we create an https server with the certificate we created and then do a tls dial with the CA. So we are testing the creation of the secret, ca and cert which is something we do in our code. Do you know a better way of testing this without creating a https server?
I think we would still have the same problem with `Listener` + `ServeTLS` as the problem is that sometimes (very rarely) we tried to do the tls dial before the server is ready

ereslibre:
> Do you know a better way of testing this without creating a https server?

I don't have a really good answer without really reproducing what we do on the implementation but on a different way. All in all, the test also depends on the implementation of the golang HTTP server being able to interpret the certificate we create. That does not say anything about how the Rust crates that will use will be able to use them or not.

In any case, is fine, I guess is a matter of having different opinions about the test itself, let's keep it as is. Sorry for the noise.

PR #94, opened 2021-09-30
cluster admission policy: show bound policy server in table

ereslibre:

PR #95, opened 2021-09-30
cluster admission policy: default to `default` policyServer explicitly

ereslibre:
When no `policyServer` is provided in a cluster admission policy,
default it to the `default` one that will have been created when
installing Kubewarden.

This, along with
https://github.com/kubewarden/kubewarden-controller/pull/94, makes it
explicit to see to what policy server every policy is bound just by
listing them.

### DIFF
--- a/apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
+++ b/apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
@@ -51,6 +51,9 @@ var _ webhook.Defaulter = &ClusterAdmissionPolicy{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *ClusterAdmissionPolicy) Default() {
 	clusteradmissionpolicylog.Info("default", "name", r.Name)
+	if r.Spec.PolicyServer == "" {
+		r.Spec.PolicyServer = "default" //nolint:goconst
+	}
 	controllerutil.AddFinalizer(r, constants.KubewardenFinalizer)
 }
 


apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
```diff
@@ -51,6 +51,9 @@ var _ webhook.Defaulter = &ClusterAdmissionPolicy{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *ClusterAdmissionPolicy) Default() {
 	clusteradmissionpolicylog.Info("default", "name", r.Name)
+	if r.Spec.PolicyServer == "" {
+		r.Spec.PolicyServer = "default"
```
ereslibre, line 55: I thought about making this constant... but there is only this usage still. Given that it's created in the helm chart etc, so I went ahead with hardcoding here...

flavio: You either turn that into a constant or you disable this lint in this area of the code. I would just bite the bullet and make it a constant

flavio (approved):
LGTM, just make sure you fix the linting error before merging

apis/policies/v1alpha2/clusteradmissionpolicy_webhook.go
ereslibre: In this instance I prefer to ignore the lint. It's the same string, but they have different meanings in the different occurences...

flavio (approved):
LGTM

viccuad (approved):
nice! could we add a simple test to webhook_suite_test.go?

ereslibre:
I think with https://github.com/kubewarden/kubewarden-controller/blob/9d0d1a02817eb12f82fffd5ea4a792b15cd35baf/apis/policies/v1alpha2/clusteradmissionpolicy_types.go#L27 is enough. That we don't need a webhook for this (we are reusing an existing one, so it wouldn't be that bad anyhow).

I'm confirming because certainly when retrieving the resource I see the `policyServer: default` correctly set, but I don't know if this is persisted with the default, or if it's defaulted on every read (I would prefer that the defaulted resource is persisted).

Closing in the meanwhile, I think we don't need this.

ereslibre:
More information on : https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

I have confirmed that after creating a clusteradmissionpolicy without a policyServer, the https://github.com/kubewarden/kubewarden-controller/blob/9d0d1a02817eb12f82fffd5ea4a792b15cd35baf/config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml#L198 CRD defaulting feature is storing the defaulted version.

I checked by looking at the sqlite3 database of a k3s with k3d cluster.

PR #112, opened 2021-10-25
feat: add OpenTelemetry integration

ereslibre:
Related: https://github.com/kubewarden/policy-server/issues/106

- Add sample `otel-collector.yaml.example` file containing a basic
OpenTelemetry collector resource that can be deployed using the
OpenTelemetry operator.

- If metrics are enabled, expose metrics port on the policy-server
service, so a prometheus instance running on the cluster can be
instructed to scrape this endpoint.

### DIFF
--- a/internal/pkg/admission/policy-server-service.go
+++ b/internal/pkg/admission/policy-server-service.go
@@ -3,15 +3,37 @@ package admission
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 
 	policiesv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 )
+
+var (
+	// This is the port where the Policy Server service will be exposing metrics. Can be overridden
+	// by an environment variable KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT
+	metricsPort = constants.PolicyServerMetricsPort
+)
+
+func init() {
+	envMetricsPort := os.Getenv(constants.PolicyServerMetricsPortEnvVar)
+	if envMetricsPort != "" {
+		var err error
+		metricsPortInt32, err := strconv.ParseInt(envMetricsPort, 10, 32)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "port %s provided in %s envvar cannot be parsed as integer: %v. Aborting.\n", envMetricsPort, constants.PolicyServerMetricsPortEnvVar, err)
+			os.Exit(1)
+		}
+		metricsPort = int(metricsPortInt32)
+	}
+}
 
 func (r *Reconciler) reconcilePolicyServerService(ctx context.Context, policyServer *policiesv1alpha2.PolicyServer) error {
 	err := r.Client.Create(ctx, r.service(policyServer))
@@ -22,7 +44,7 @@ func (r *Reconciler) reconcilePolicyServerService(ctx context.Context, policySer
 }
 
 func (r *Reconciler) service(policyServer *policiesv1alpha2.PolicyServer) *corev1.Service {
-	return &corev1.Service{
+	svc := corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      policyServer.NameWithPrefix(),
 			Namespace: r.DeploymentsNamespace,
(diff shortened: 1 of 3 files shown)


internal/pkg/admission/policy-server-service.go
```diff
@@ -33,13 +34,38 @@ func (r *Reconciler) service(policyServer *policiesv1alpha2.PolicyServer) *corev
...
+	if metricsEnabled(policyServer) {
+		svc.Spec.Ports = append(
+			svc.Spec.Ports,
+			corev1.ServicePort{
+				Name:       "metrics",
+				Port:       constants.PolicyServerMetricsPort,
+				TargetPort: intstr.FromInt(constants.PolicyServerMetricsPort),
+				Protocol:   corev1.ProtocolTCP,
+			},
+		)
```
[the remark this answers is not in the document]
ereslibre: Endpoints from the service will be discovered, and will be added to targets by the prometheus operator, so we are good in this sense.

viccuad (approved):
Looks good, besides the golangci-lint errors.

internal/pkg/admission/policy-server-service.go
```diff
@@ -3,16 +3,37 @@ package admission
...
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 )
 
+var (
+	// This is the port where the Policy Server service will be exposing metrics. Can be overriden
```
flavio, line 20: Typo:

```suggestion
	// This is the port where the Policy Server service will be exposing metrics. Can be overridden
```

internal/pkg/admission/policy-server-service.go
```diff
@@ -3,16 +3,37 @@ package admission
...
 	"fmt"
+	"os"
+	"strconv"
 
 	policiesv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
```
flavio, line 14: I guess we're using that to address the potential buffer overflow when parsing a string back into an Int.

I haven't looked at the code, but I would prefer to use something provided by the core Go library. The reason is: I fear we might ran into security issues if we do not update this dependency.

According to [that](https://github.com/kubernetes/kubernetes/issues/81121), we could simply use [`strconv.ParseInt`](https://pkg.go.dev/strconv#ParseInt)

internal/pkg/admission/policy-server-service.go
```diff
@@ -3,16 +3,37 @@ package admission
...
+	// This is the port where the Policy Server service will be exposing metrics. Can be overriden
+	// by an environment variable KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT
+	metricsPort = constants.PolicyServerMetricsPort
+)
+
+func init() {
+	envMetricsPort := os.Getenv(constants.PolicyServerMetricsPortEnvVar)
+	if envMetricsPort != "" {
+		var err error
+		metricsPort, err = strconv.Atoi(envMetricsPort)
```
flavio, line 29:
```suggestion
		metricsPort, err = strconv.ParseInt(envMetricsPort)
```

internal/pkg/admission/policy-server-service.go
```diff
@@ -3,16 +3,37 @@ package admission
...
+		if err != nil {
+			fmt.Printf("port %s provided in %s envvar cannot be parsed as integer: %v. Aborting.\n", envMetricsPort, constants.PolicyServerMetricsPortEnvVar, err)
```
flavio, line 31: Maybe replace with:
```suggestion
			fmt.Fprintf(os.Stderr, "port %s provided in %s envvar cannot be parsed as integer: %v. Aborting.\n", envMetricsPort, constants.PolicyServerMetricsPortEnvVar, err)
```

internal/pkg/admission/policy-server-service.go
```diff
@@ -33,13 +34,38 @@ func (r *Reconciler) service(policyServer *policiesv1alpha2.PolicyServer) *corev
 		Spec: corev1.ServiceSpec{
 			Ports: []corev1.ServicePort{
 				{
-					Port:     constants.PolicyServerPort,
-					Protocol: corev1.ProtocolTCP,
+					Name:       "policy-server",
+					Port:       constants.PolicyServerPort,
+					TargetPort: intstr.FromInt(constants.PolicyServerPort),
```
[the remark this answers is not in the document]
flavio: Also, see the initial comment about replacing `intstr` with `strconv.ParseInt`

flavio (changes requested):
LGTM, I left some comments

internal/pkg/admission/policy-server-service.go
ereslibre: Why so? `ParseInt` expects three arguments. `Atoi` is more handy, performs some safety checks and relies on `ParseInt` when necessary. Am I missing something?

ereslibre: Ugh, I totally see the security issue here now. Thanks. Addressing it.

internal/pkg/admission/policy-server-service.go
flavio: I see :+1: 

Kubernetes really lives inside of its own dimension :)

flavio (approved):
LGTM. I see there are other linter issues, but they should be fixed with another PR, before tagging the next release

jvanz (approved):
Besides the comments from the linter, LGTM!

PR #118, opened 2021-11-03
Update changelog for release

viccuad:
Updated manually for now, as git-chglog removes manually added changelog content for v0.4.0.
An option would be to craft the template to contain those old entries: https://github.com/git-chglog/git-chglog/issues/55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 ## [Unreleased]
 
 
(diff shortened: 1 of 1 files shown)


viccuad:
Once this hits master, I will push the v0.4.1 tag to https://github.com/kubewarden/kubewarden-controller/pull/118/commits/ac9ded8f412c2a7224e3185e4861664a42c572ef.

PR #121, opened 2021-11-04
Policy counter

### LINKED ISSUE #101: Metric: policy counter
The goal of this metric is to allow an operator to understand how many policies are installed in the cluster.

The policy provides also some labels, that can be used to filter the metric and provide really specific data.

## Metric name

`kubewarden_policy_total`

## Metric type

Counter

## Use cases

* How many policies have been active inside of the cluster during a certain time frame

## Labels

* `name` - string: name of the policy as registered in the Admission Policy custom resource
* `policy_server` - string: policy server bound to this policy
* `module` - string: URI pointing to the location where the policy is pulled from
* `failurePolicy` - string: `ignore` | `fail`. The failure policy that will be used to register the Kubernetes webhook
* `mutating` - boolean: whether this policy is mutating or not
* `cluster` - boolean: whether this policy is cluster wide (ClusterAdmissionPoilcy) or namespaced -- future (AdmissionPolicy)
* `namespace` - string: applies only to `AdmissionPolicy` resources (future). The namespace where the AdmissionPolicy resource is persisted
* `policyStatus` - string: `unscheduled` | `unschedulable` | `pending` | `active`

jvanz:
Adds the integration necessary to send metric to the OpenTelemetry collector and define a metric to count the number of active policies.

Closes #101


jvanz:
The code is able to connect to OpenTelemetry Collector and send metrics. But I would like to test it end-to-end (with Prometheus) in a k8s cluster. I'm doing that now.

jvanz:
> The code is able to connect to OpenTelemetry Collector and send metrics. But I would like to test it end-to-end (with Prometheus) in a k8s cluster. I'm doing that now. That's why I'm keeping the PR as draft. Even then, review are welcome. ;)

After a call with the team, as we can see the metric in the OpenTelemetry collector, we decided to mark this ready for reviews

internal/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	POLICY_COUNTER_METRIC_DESCRIPTION = "How many policies are installed in the cluster"
+)
+
+func New(openTelemetryEndpoint string) error {
+	ctx := context.Background()
+
+	driver := otlpgrpc.NewDriver(
+		otlpgrpc.WithInsecure(),
+		otlpgrpc.WithEndpoint(openTelemetryEndpoint),
+	)
```
jvanz, line 30: I got this default configuration ( `otlpgrpc.WithInsecure()` )from the OpenTelemetry examples. Which means it does not use secure connections. I was thinking to add more configuration for this. What do you think? Does it make sense add a secure connection to a sidecar collector? (considering that the user is using a sidecar)

Even if we decided to use sidecar as recommended way, the user can change how to deploy the OpenTelemetry. Which can cause issues... So, I believe it is still valuable add more confiuration here.

flavio (commented):
I left some comments. I've also another final question: I see we're updating the policy count only when a new policy is added, not when a deletion happens.

I guess that's because a Prometheus Counter can only be increased. Am I right? I wonder if we shouldn't use a Gauge for that

controllers/policies/policyserver_controller.go
```diff
@@ -162,10 +163,12 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				policy.Status.PolicyStatus = policiesv1alpha2.ClusterAdmissionPolicyStatusUnscheduled
 				err := r.Reconciler.UpdateAdmissionPolicyStatus(context.Background(), policy)
 				if err != nil {
+					metrics.RecordPolicyCount(policy)
 					r.Log.Error(err, "cannot update status of policy "+policy.Name)
 					return []ctrl.Request{}
 				}
 			}
+			metrics.RecordPolicyCount(policy)
```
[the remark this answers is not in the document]
jvanz: Yeah, I think you are right. As the metrics have the policy status I was trying to record any status change. So,  I removed the line 171 and add another one to record when the status is set to unschedulable.

jvanz:
> I left some comments. I've also another final question: I see we're updating the policy count only when a new policy is added, not when a deletion happens.

Yes
> I guess that's because a Prometheus Counter can only be increased. Am I right?

Yes. 
> I wonder if we shouldn't use a Gauge for that

Yeah. It makes sense. But I think we should now add the policy status for this metric, right? Otherwise, we will have new time series for each status where will never go down to zero. If we do that, we will lose the feature to check how many policies are active or another status. I need to test more this to ensure if this makes sense or not.

controllers/policies/policyserver_controller.go
```diff
@@ -162,6 +163,7 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				policy.Status.PolicyStatus = policiesv1alpha2.ClusterAdmissionPolicyStatusUnscheduled
 				err := r.Reconciler.UpdateAdmissionPolicyStatus(context.Background(), policy)
 				if err != nil {
+					metrics.RecordPolicyCount(policy)
```
viccuad, line 166: I'm not sure about this line; I feel it should be out of the `if err!= nil` block. I also feel that a fresh new unscheduled policy is going to be counted twice: here in this line and then later on.
If the intent is to update the policy with the RecordPolicyCount(), so the fields are also updated, could it be possible to call `metrics.RecordPolicyCount()` only once, just inside of `UpdateAdmissionPolicyStatus()`?

jvanz: Oh... you are right. Moving the function call to `UpdateAdmissionPolicyStatys` is much better.

viccuad (approved):
LGTM!

PR #122, opened 2021-11-05
refactor: Fix golangci-lint 1.43 issues

viccuad:
The GH Action for golangci-lint now uses golangci-lint v1.43.
This PR makes golangci-lint happy.


flavio (approved):
Thanks for having fixed these warnings! :heart:

flavio:
> I also would like to enable gating of PRs so they can't be merged unless all checks are green.

Done, I've made also the linter checks mandatory

PR #127, opened 2021-11-24
Adds end-to-end tests.

jvanz:
Adds basic end-to-end tests in the release process. The Github workflow has been updated to run unit tests and end-to-end tests before the release creation. Thus, the release is created only with all the tests passed.

There are some topics that I would discuss with the team.

* Should we keep the files used to run the e2e in this repo?
* Should we keep the actions created to be used in the e2e-tests in this repo or move to another repo?

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,0 +1,16 @@
+name: Build container image every change.
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+
+
+jobs:
+  build:
+    name: Build
+    uses: jvanz/kubewarden-controller/.github/workflows/container-image.yml@main
+    with:
+      push-image: true
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,16 +1,32 @@
-on:
-  push:
-    branches:
-    - main
-    tags:
-    - 'v*'
+name: Build container image
 
-name: build container image
+on:
+  workflow_call:
+    inputs:
+      push-image:
+        type: boolean
+        required: true
+    outputs:
+      repository:
+        description: "Repository used to build the container image"
+        value: ${{ jobs.build.outputs.repository }}
+      tag:
+        description: "Tag used to build the container image"
+        value: ${{ jobs.build.outputs.tag }}
+      artifact:
+        description: "Uploaded artifact with the container tarball"
+        value: ${{ jobs.build.outputs.artifact }}
+
+
 
 jobs:
   build:
     name: Build container image
     runs-on: ubuntu-latest
+    outputs:
+      repository: ${{ steps.setoutput.outputs.repository }}
+      tag: ${{ steps.setoutput.outputs.tag }}
+      artifact: ${{ steps.setoutput.outputs.artifact }}
     steps:
       -
         name: Checkout code
@@ -29,24 +45,18 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
       -
-        name: Build and push development container image
+        name: Retrieve tag name
         if: ${{ startsWith(github.ref, 'refs/heads/') }}
-        uses: docker/build-push-action@v2
-        with:
-          context: .
-          file: ./Dockerfile
-          platforms: linux/amd64, linux/arm64
-          push: true
-          tags: |
-            ghcr.io/kubewarden/kubewarden-controller:latest
+        run: |
+          echo TAG_NAME=latest >> $GITHUB_ENV
       -
         name: Retrieve tag name
         if: ${{ startsWith(github.ref, 'refs/tags/') }}
         run: |
-          echo TAG_NAME=$(echo $GITHUB_REF | sed -e "s|refs/tags/||") >> $GITHUB_ENV
+            echo TAG_NAME=$(echo $GITHUB_REF | sed -e "s|refs/tags/||") >> $GITHUB_ENV
       -
-        name: Build and push tagged container image
-        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        name: Build and push container image
+        if: ${{ inputs.push-image }}
         uses: docker/build-push-action@v2
         with:
           context: .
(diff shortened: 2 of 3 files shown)


jvanz:
This is related to the issue #127

Makefile
```diff
@@ -61,6 +61,10 @@ test: manifests generate fmt vet ## Run tests.
 	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.7.2/hack/setup-envtest.sh
 	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
 
+e2e-tests:
+	k3d cluster delete e2e-test-cluster
+	k3d cluster create e2e-test-cluster
```
flavio, line 66: I guess we will run the tests locally too at one point. This is what we're aiming for, isn't it?

flavio (commented):
This is great work!

My main concern is... I get the **feeling** (I might be wrong) that we're moving into something that is heavily event-driven. I have mixed feeling about that. It is definitely harder to understand and maybe to debug too.
Just to be clear, I'm referring to all the new GH actions

.github/workflows/release.yml
```diff
@@ -3,18 +3,36 @@ on:
...
+    name: "Run Golang linter"
+    uses: kubewarden/kubewarden-controller/.github/workflows/golangci-lint.yml@main
+  unit-tests:
+    name: "Run unit tests"
+    uses: kubewarden/kubewarden-controller/.github/workflows/unit-test.yml@main
+  build-container-image:
+    name: "Build container image"
+    uses: kubewarden/kubewarden-controller/.github/workflows/container-image.yml@main
+    with:
+      push-image: false
```
ereslibre, line 19: Can we make `@main` a parameter of the workflow? Adding to previous concerns, on the release workflow we might be releasing a different commit than what `main` points to, and in that timeframe, unit tests, linter... might have changed. Ideally we should be referring to the current commit if possible.

viccuad (commented):
This is great work and has a lot of coverage!

However, I share the same concern as Flavio with the tight coupling to GHA. I'm afraid that the moment the CI grows, it's gonna be painful for me to make big changes to it. Maybe it's my ptsd from previous kubernetes CIs, where the testsuites were not under our control and we either had everything tightly coupled with the scheduler (Concourse), or we decoupled it but ended being a shellscript variable mess as we had too many parametrization in a design that was not simple to start with.

Maybe we have differing opinions on what we want the CI for.
Do we want to be able to run it locally (to aid kubewarden development, and the CI itself)? Do we want to be able to run subsets of tests locally?
Do we intend to expand it in the future to run performance tests? if so, does it need to run outside of GH?
Do we want to support several Kubernetes providers or be provider agnostic? (just using the current kubeconfig and be done with it, to change from k3s to minikube, AKS, etc)
Do we foresee the need to test policies/kubewarden specific to a stack? and if so, if the stack is not simple, would it be useful to just run the CI against a preexisting stack from a colleage/customer?
Is there anything specific that would need to be tested for the sigstore story?
Do we want to somehow test telemetry functionality?
Do we want to include tests for UI in the future? do they share the same setup as the rest of the CI?

This could get supercomplicated indeed, and I would prefer your current approach than overblowing it for sure. Yet I had in mind writing the bulk of setup and test commands on bash scripts, that get run with `set -ex` to print and stop on failure. Something like this:

```
vic@viccuad2 ~$ tree ci
ci
├── 10_create_cluster
│   └── 10_k3s.sh
├── 20_install_deps
│   ├── 10_cert-manager.sh
│   ├── 20_prometheus.sh
│   ├── 30_opentelemtry.sh
│   └── 40_grafana.sh
├── 30_install
│   └── 10_kubewarden.sh
├── 40_test.sh
│   ├── 10_install_policy.sh
│   ├── 20_validation_policy.sh
│   ├── 30_mutating_policy.sh
│   ├── 33_foo_test.sh
│   │   ├── 00_setup.sh
│   │   ├── 10_test_battery1.sh
│   │   ├── 20_test_battery2.sh
│   │   └── 99_cleanup.sh
│   ├── 40_perf_test.sh
│   └── 50_telemetry_test.sh
└── 99_cleanup

vic@viccuad2 ~$ cat ci/30_install/10_kubewarden.sh
#!/usr/bin/env sh

# IMAGE_REPO gets set in the runner
# IMAGE_TAG gets set in the runner

# Install Kubewarden Helm repository:
helm repo add --force-update kubewarden https://charts.kubewarden.io
# Install Kubewarden CRDS:
helm upgrade --install --wait --namespace kubewarden --create-namespace kubewarden-crds kubewarden/kubewarden-crds
# Install Kubewarden controller
helm upgrade --install --wait --set image.repository="$IMAGE_REPO" --set image.tag="$IMAGE_TAG" --namespace kubewarden kubewarden-controller kubewarden/kubewarden-controller
kubectl get pods --namespace kubewarden -o yaml
```

The local runner `run.sh` would consist in:
- a list of default values for variables (IMAGE_TAG, IMAGE_REPO, etc) which can be overriden manually, for example in the GHA run
-  a couple of shellscript lines that just run any executable in the `ci` folder in alphanumerical order. This happens recursively, so all scripts end up being run in the defined order.
The runner can be tweaked to just run the desired folder(s) (only setting up the cluster for dev, or only running a specific part of the testsuite).

The GHA runner can just call the same local runner, setting correct variables, into the jobs (so no need for a step for each specific command, just a GHA step for each "folder" call).

Maybe this is too brittle the moment we have a lot of variables, and would be better done in a test framework, dunno.

edit 2021-12-3: I would like the folder layout also for automated linting (shellscript). And being able to run the CI locally means we can use it as debugger when there's failed tests or problems in the CI itself. Also, good CIs may be left-over product, with value in itself.

.github/actions/create-kubewarden-cluster/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+        name: "Install Helm"
+        with:
+          version: 'latest' # default is latest stable
+        id: install
+      - name: "Install cert-manager"
+        run: kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
+        shell: bash
+      - name: "Wait for cert-manager be ready"
+        run: kubectl wait --for=condition=Available deployment --timeout=3m -n cert-manager --all
+        shell: bash
```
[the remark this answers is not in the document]
jvanz: This is a line copied from our docs just for the wait `cert-manager` to be ready before continuing the installation. Avoiding some possible weird problems due to missing components. To avoid one additional step, I'll move the command to the previous command. Sounds good?

.github/workflows/ci.yml
```diff
@@ -4,18 +4,21 @@ on:
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-go@v2
-        with:
-          go-version: '1.16'
-      - run: make test
+  linter:
+    name: "Run Golang linter"
+    uses: kubewarden/kubewarden-controller/.github/workflows/golangci-lint.yml@main
```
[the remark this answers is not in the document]
jvanz: I see. I'm using the main because is the branch that I have the files. But you are right. This can be an issue. I'll check if it is possible to use some Github env vars or context to get the current branch and use it on this.

.github/actions/create-kubewarden-cluster/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - uses: AbsaOSS/k3d-action@v1.5.0
+        name: "Create Kubernetes cluster"
+        with:
+          cluster-name: "e2e-test-cluster"
+      - name: "Download local image"
+        if: ${{ inputs.load-local-image }}
+        uses: actions/download-artifact@v2
+        with:
+          name: kubewarden-controller-image-${{inputs.controller-image-tag}}
+          path: /tmp
```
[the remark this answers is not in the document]
jvanz: From the assets. I did that to avoid publishing a container image in the registry which is not ready for use. 

[See an example](https://github.com/jvanz/kubewarden-controller/actions/runs/1500495322)

Makefile
jvanz: Yes. I decided to use the k3d to allow us to run the tests locally. But I'm not working helper script to allow that yet.

.github/workflows/release.yml
jvanz: Good idea. I think is doable. I'll work on that while checking all the concerns from you and Flavio.

jvanz:
@viccuad thanks for your comment. I'll give you a better feedback soon describing better the goals in the issue trying to address all your questions.

jvanz:
I'm closing this PR because I forgot that the `main` branch has a PR open for it in my fork, and I do not like this bunch of branch updates with no reason from the review process.

jvanz:
> @viccuad thanks for your comment. I'll give you a better feedback soon describing better the goals in the issue trying to address all your questions.

@viccuad , I've updated the issue which we are using to track the story of the end-to-end tests. I've added what it is expected of these changes. I still need to think more about the scripts structure you proposed though.

viccuad:
Thanks @jvanz!  :heart:

Issue #126, opened 2021-11-19
Create RFC about how to implement the audit mode for policies


ereslibre:
We have been calling this "Audit mode" but the RFC refers to this feature as "report mode". Auditing is left to the concept of auditing a cluster against all existing policies (policies might change over time enforcing different things).

PR #128, opened 2021-11-25
Add Report Mode RFC

ereslibre:
Fixes: https://github.com/kubewarden/kubewarden-controller/issues/126

What we have been calling "audit" mode in our issues is what this RFC describes. I think the name used in the RFC is closer to what we mean: "report" mode. A policy can be in "report" mode (what is described in this RFC) or in "enforce" mode (the only mode we have right now).

Audit was part of this RFC as well, but I wanted to split it so I can open another one with only Audit. In my opinion, Audit refers to the fact of auditing the existing state of the cluster, this is: checking that all resources that are persisted at this time are validated by all policies that target them.

### DIFF
--- a/rfc/0002-report-mode.md
+++ b/rfc/0002-report-mode.md
@@ -1,0 +1,124 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Report Mode                      |
+| Start Date   | Nov 25th 2021                    |
+| Category     | Policy Enforcement               |
+| RFC PR       | [PR128](https://github.com/kubewarden/kubewarden-controller/pull/128)  |
+
+# Summary
+[summary]: #summary
+
+> Brief (one-paragraph) explanation of the feature.
+
+The report mode will improve the visibility of a cluster administrator with regards to policies and
+how they act. It will enable the cluster administrator to understand the consequences of fully
+activating a policy before it starts to reject requests.
+
+# Motivation
+[motivation]: #motivation
+
+> - Why are we doing this?
+> - What use cases does it support?
+> - What is the expected outcome?
+
+> Describe the problem you are trying to solve, and its constraints, without coupling them too
+> closely to the solution you have > in mind. If this RFC is not accepted, the motivation can be
+> used to develop alternative solutions.
+
+Currently, Policies are deployed and become active immediately -- accepting, mutating or rejecting
+requests.
+
+With this context, as a cluster administrator it is not easy to understand the consequences of
+deploying new policies in a running cluster. This RFC works on the strategy to improve this
+situation and let the administrator gain visibility before policies are truly enforced in the
+cluster.
+
+Introduce a report mode in Kubewarden that acts in the following way:
+
+1. Deploy policies without rejecting or mutating requests, so actions that would have been taken had
+   the policy been active are logged. We will call this method "reporting"  mode, because it lets
+   the cluster operator understand if a policy in this mode would be rejecting requests that impact
+   the regular functionality of the cluster.
+
+## Examples / User Stories
+[examples]: #examples
+
+> Examples of how the feature will be used. Interactions should show the action and the response.
+> When appropriate, provide user stories in the form of "As a [role], I want [feature], so [that]."
+
+As a Kubernetes administrator I want to deploy new policies in a state that they don't perform any
+request rejections, but only inform about what would have been rejected had the policy been active.
+
+As a Kubernetes administrator I want to deploy new policies in "reporting" mode, and then flip the
+switch of the policy to activate it after it has been in "reporting" mode for a short while that I
+considered enough to learn if it would have an undesired impact in the cluster.
+
+As a Kubernetes administrator I want to plot in Grafana acceptions/rejections/mutations coming from
+policies in "report" mode in the same way that I do with policies in "enforce" mode -- the only mode
+Kubewarden had until now.
+
+# Detailed design
+[design]: #detailed-design
+
+> This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with the
+> product to understand, and for somebody familiar with the internals to implement.
+
+> This section should cover architecture aspects and the rationale behind disruptive technical
+> decisions (when applicable), as well as corner-cases and warnings.
+
+## "report" mode
+
+### CRD
+
+The `ClusterAdmissionPolicy` CRD will be extended so it includes a `mode` attribute:
+
+```go
+type ClusterAdmissionPolicySpec struct {
+   // ...
+   Mode string `json:"mode,omitempty"`
+   // ...
+}
+```
+
+Two values are allowed: `enforce` and `report`. If `mode` is omitted,
+it will be defaulted to `enforce`.
+
+It's possible to update the `mode` from `report` to
+`enforce`.
+
+Allowing to change `mode` from `enforce` to `report` will be rejected
+-- so it's not possible to deactivate policies just by having `UPDATE`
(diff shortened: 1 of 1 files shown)


rfc/0002-report-mode.md
```diff
@@ -0,0 +1,122 @@
...
+   // ...
+   Mode string `json:"mode,omitempty"`
+   // ...
+}
+```
+
+Two values are allowed: `enforce` and `report`. If `mode` is omitted, it will be defaulted to
+`enforce`. It's possible to update the `mode` from `report` to `enforce`. Allowing to change `mode`
+from `enforce` to `report` requires further consideration -- so it's not possible to deactivate
+policies just by having `UPDATE` permissions.
```
flavio, line 86: I don't understand what would prevent us from changing a policy mode from "enforce" to "report". You hint at a technical limitation, what is that about?

rfc/0002-report-mode.md
```diff
@@ -0,0 +1,122 @@
...
+### Policy Server
+
+The Policy Server has to be updated so the `mode` is included in the [`Policy` struct used for
+settings](https://github.com/kubewarden/policy-server/blob/c8d64da87448b7f9250a1d6b5e56817f25b56359/src/settings.rs#L11-L19).
+
+The Policy Server will always accept the request if this struct `mode` is set to `report`.
+
+The Policy Server will log with INFO level every evaluation specifying the `mode` of the policy. If
+the policy is mutating the resource, the resulting object will be printed in the structured logging
+output.
```
flavio, line 97: Nitpicking, instead of logging I would refer to tracing. This is more powerful

rfc/0002-report-mode.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Metrics
+
+A new `PolicyReportEvaluation` metric has to be created akin to the current `PolicyEvaluation`
+metric. The reason for not adding `mode` to the existing `PolicyEvaluation` is to keep alerting
+queries for true evaluations easy (without the need to filter `mode = "enforce"`). The metric will
+have the same baggage as the `PolicyEvaluation` metric.
```
flavio, line 104: What about the other queries? I wouldn't want to come up with duplicating each data series we create.

Can't we just filter out by baggage?

flavio (commented):
Overall looks good to me.

On top of the comments I left inside of the RFC, I have another question: how is an operator going to understand which resources would be impacted by a policy?

Does the operator have to do some spelunking inside of prometheus/grafana/jaeger?
Would it be possible to add a special label/annotation to all the resources that would have been blocked? The idea is: with a simple `kubectl` invocation I can immediately see what would have been blocked by a policy once its operational mode is changed from "report" to "enforce".

rfc/0002-report-mode.md
ereslibre: RBAC won't help us here. I think the only way we can prevent this is by using our own webhook to disable `enforce` -> `report` transition for security. If someone wants to disable a policy temporarily they need to have DELETE permissions, and then they can recreate with `report` if desired. Point is, if a user only has `UPDATE` and `CREATE` permissions, they should not be allowed to transition from `enforce` to `report`.

ereslibre: Since transitioning from `report` to `enforce` is stronger, I don't see a reason to forbid.

rfc/0002-report-mode.md
ereslibre: I would keep things simple at this time. Evaluations is what let the user understand what this policy would be rejecting had it been in `enforce` mode. Our policy count in the controller won't see any change, we can add the `mode` on the baggage, but not a different metric.

I consider evaluations need a different metric because it would impact graphs for things that are not really rejected.

ereslibre:
> On top of the comments I left inside of the RFC, I have another question: how is an operator going to understand which resources would be impacted by a policy?

I think this is part of the `Audit` feature that I'm preparing an RFC for too. If by "understand what resources would be impacted by a policy" you mean already persisted policies, I think this is "Audit" as in "Auditing" an existing state of the cluster to new or different constraints established afterwards.

> Does the operator have to do some spelunking inside of prometheus/grafana/jaeger?

In this mode, they can either look at the logs from the policy server as it accepts/rejects/mutates or at add the new metric to prometheus, so they can learn about what requests would have been rejected in real time.

> Would it be possible to add a special label/annotation to all the resources that would have been blocked? The idea is: with a simple kubectl invocation I can immediately see what would have been blocked by a policy once its operational mode is changed from "report" to "enforce".

I like the idea as we have discussed several times, but after further consideration I see some disadvantages:

1. Automated deployment tooling that might try to identify differences on the spec and redeploy the resource. They would identify the resource as modified every time.
2. Adding an annotation comes with the responsibility to clean up the annotation. When? If we delete the policy? If we promote the policy from `report` to `enforce`?
3. Adding an annotation is very tied to CREATE, UPDATE and PATCH requests. The rest of the supported verbs is not so clearly visible and you still need to look at the logs/metrics to understand if a DELETE request is blocked.
4. Adding an annotation is not clear enough when targeting UPDATE requests. If we get an UPDATE request, do we annotate the existing object, that might be valid? How does the operator know that what is wrong is the new UPDATE request, but that the object persisted is fine?

I think it doesn't bring very good properties to this problem and we should try to avoid modifying the spec for this.

rfc/0002-report-mode.md
viccuad: >  Since transitioning from report to enforce is stronger, I don't see a reason to forbid.

It could produce DOS against something not properly configured for example, I would forbid that change too.

ereslibre: Note that what I am proposing is going to stronger is fine. Not the other way around. If that causes a DOS is not our fault IMO, we have to provide the tools for that to not happen (like the report mode)

ereslibre: For what is worth they could have DOS'ed from the very beginning by creating the policy in enforce mode directly.

rfc/0002-report-mode.md
jvanz: My 50 cents on the discussion... If added the `mode` in the baggage we can create a variable that allows users to switch among modes just by changing one field. Something similar to what we did for the graphs showing data for specific policies. So, AFAICS, the user will not need to rewrite the query.

rfc/0002-report-mode.md
flavio: I see, thanks for the explanations.

Overall I agree with:
* `report` -> `enforce` being allowed :green_circle: 
* `enforce` -> `report` **not** being allowed :red_circle:

rfc/0002-report-mode.md
flavio: Ok, I see we're talking just about one metric. Still, I think we should use the same metric and eventually update the grafana dashboard to filter out the rejections that are "coming" from a policy in "report" mode

ereslibre: Honestly, I don't have a blocker here. My only point here is that we don't dirt the evaluations metric in such a way that for getting the metric for final evaluations the user now needs to filter out a boolean, so they can ignore policies on report mode.

ereslibre: +1, I'm updating the RFC to remove the new metric. I started my reply before you wrote @flavio.

ereslibre: Updated.

viccuad (approved):

jvanz (approved):

flavio (approved):

flavio:
I think we can merge it

Issue #130, opened 2021-11-29
Unable to remove unschedulable clusteradmissionpolicy


chrisns:
👋 
```
$ kubectl apply -f - <<EOF
apiVersion: policies.kubewarden.io/v1alpha2
kind: ClusterAdmissionPolicy
metadata:
  name: privileged-pods
spec:
  policyServer: typo
  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
  rules:
  - apiGroups: [""]
    apiVersions: ["v1"]
    resources: ["pods"]
    operations:
    - CREATE
    - UPDATE
  mutating: false
EOF

$ kubectl get ClusterAdmissionPolicy
NAME               POLICY SERVER                    MUTATING   STATUS
privileged-pods    typo                             false      unschedulable

$kubectl delete clusteradmissionpolicies.policies.kubewarden.io privileged-pods
clusteradmissionpolicy.policies.kubewarden.io "privileged-pods" deleted
***hangs***
^C
$ kubectl get ClusterAdmissionPolicy
NAME               POLICY SERVER                    MUTATING   STATUS
privileged-pods    typo                             false      unschedulable

```
only way to remove is apparently to edit the `ClusterAdmissionPolicy` and remove the `kubewarden` `finalizer` then it disappears

Because the `policyServer` field is immutable,  the only way to fix a typo like this is to remove the finalizer and try again 😢

viccuad:
Thanks for reporting this! Yes, indeed, the UX here can be fixed.

On the implementation side, this is happening because policies get reconciled as part of the PolicyServer loop.
We can defer adding the Finalizer to when the policy is scheduled in a PolicyServer (so in this case a badly configured ClusterAdmissionPolicy will not have a finalizer yet). Another option, suggested by @raulcabello, is to remove the finalizer [at this point](https://github.com/kubewarden/kubewarden-controller/blob/main/controllers/policies/policyserver_controller.go#L177-L183) in the watch, for now.

PR #135, opened 2021-12-10
fix: Delete Policies not associated with PolicyServers, don't update Policy.Status if policy r-o

viccuad:
- fix: Delete Policies not associated with PolicyServers by removing their Finalizers. For example, those that have a typo on
  their spec.policyServer. Doing this in the watch of ClusterAdmissionPolicies & PolicyServers, but without triggering the reconcile loop of PolicyServers.
- fix: Don't Update Policy.Status if Policy is read-only.
- Minor rewording.
- Bump cyclomatic complexity to 16 from 13 for the watch block in policyserver_controller.go.

Fixes: https://github.com/kubewarden/kubewarden-controller/issues/130


.golangci.yml
```diff
@@ -52,6 +52,6 @@ linters:
 
 linters-settings:
   cyclop:
-    max-complexity: 13
+    max-complexity: 16
```
viccuad, line 55: I'm not happy with this at all; the watch is getting complicated. While it looks ok as it checks conditions on the ClusterAdmissionPolicy, and updates .Status or removes Finalizer, it's doing too much, Particularly, it's doing Update operations, and we already have errors surfacing from targetting old versions of the Object when performing Updates on the .Status.

ereslibre: What do you think about extracting the finalizer GC to a specific loop? The loop would only do that, so we keep separate loops for different purposes.

viccuad: That seems like a good idea in principle, but I think only solves half the problem. It seems that the current watch gets triggered twice on deletion, regardless of Finalizer (opened https://github.com/kubewarden/kubewarden-controller/issues/136, see example 2).

I may be missing knowledge on the internal Kubernetes GC loop. Is marking for deletion (setting DeletionTimestamp) idempotent? Or does it perform 2 Updates, hence why the watch gets triggered twice? Is the watch FIFO or FILO (looks FILO to me?)

About separating into 2 loops, wouldn't the update to remove the Finalizer trigger the other loop?
Maybe we could have 2 loops:
- one triggered on Delete operations: checks if policy has Finalizer && an associated PolicyServer. If not, remove Finalizer. If it has associated PolicyServer, call PolicyServer reconciler which would delete the Policy itself.
- One triggered on Create/Update: sets the Policy.Status.

ereslibre: > Is marking for deletion (setting DeletionTimestamp) idempotent?

Haven't checked if

> I may be missing knowledge on the internal Kubernetes GC loop. Is marking for deletion (setting DeletionTimestamp) idempotent? Or does it perform 2 Updates, hence why the watch gets triggered twice? Is the watch FIFO or FILO (looks FILO to me?)

I haven't checked that deep if it's FIFO or LIFO, but I don't think that matters much, we cannot rely on that implementation detail. I **think** is neither, the API Server lists resources in a given order, and loops over those GC'ing those with deletionTimestamp != null and no finalizers, but I don't think it sorts by deletion timestamp in any way. Haven't checked the source code though.

> Or does it perform 2 Updates, hence why the watch gets triggered twice? Is the watch FIFO or FILO (looks FILO to me?)

I have checked by adding a finalizer of my own to a namespace I created. If I delete several times, the `deletionTimestamp` is kept if it was set already. Further deletions do not refresh this timestamp. In that context, it is idempotent.

> About separating into 2 loops, wouldn't the update to remove the Finalizer trigger the other loop and itself again?

Leaving aside this technical details I would go ahead with what you propose afterwards. If our preconditions are different on each loop, there is no way in which one can impact the other. I don't have a strong recommendation at the moment on how to split the loops, but I can think of:

1) Loop 1 (the one we already have). Create/Update + Happy path delete. In this case, deletion is only reconciled if the policy server is found. We reconcile the policy server normally, and when we are done, we remove the finalizer.

2) Loop 2: deletion is only reconciled if the policy server is not found. We don't reconcile anything, we just remove the finalizer and let the Kubernetes GC get rid of the ClusterAdmissionPolicy on its next run.

raulcabello: I agree that the watch block contains too much logic and it's getting complicated. I like the idea of having 2 watch loops. If loop 1 is setting the status and loop 2 is removing the finalizer for policies without a PolicyServer I think we might have some race conditions problems. What about if we split them like this:
- Loop 1 reconciles policies bounded to an existing PolicyServer
- Loop 2 reconciles policies not bounded to any PolicyServer (create, set status and delete)

Loop 1 can trigger PolicyServer reconcilation loop, whereas loop 2 will never do. I think we would not have conflicts with this. In addition to deletion of policies without PolicyServer I would move this to Loop 2 https://github.com/kubewarden/kubewarden-controller/blob/main/controllers/policies/policyserver_controller.go#L177-L183

ereslibre: The plan LGTM. Thanks both!

viccuad: Those 2 loops sounds good, but I'm wondering how the implementation looks. If I understand correctly, Loop 1 stays on the watch of PolicyServer, but how to not let status updates or deletes trigger that watch?
And Loop 2 would then be a new reconcile Loop for ClusterAdmissionPolicies?

On the topic, do we want to have this refactoring superseding this PR, for addressing the bug?

raulcabello: I was thinking of just another watch. So here https://github.com/kubewarden/kubewarden-controller/blob/main/controllers/policies/policyserver_controller.go#L154 we would have two 
`Watches(&source.Kind{Type: &policiesv1alpha2.ClusterAdmissionPolicy{}} ...`

viccuad: Thanks, I understand now! So 2 watches, and we let them trigger both. We ask for the list of PolicyServers, and see if there's any of them that match policy.spec.policyServer. And then, continue with our loop.

Still, IIUC, https://github.com/kubewarden/kubewarden-controller/issues/136 will still be happening in Loop 2 (and now it triggers 4 watches, twice per loop).

raulcabello: yes https://github.com/kubewarden/kubewarden-controller/issues/136 is a separate issue.

raulcabello: @viccuad It should be ok, but keep in mind that updates trigger 2 watch calls, one with the old object and another with the new one. More info https://github.com/kubewarden/kubewarden-controller/issues/136#issuecomment-992604483

viccuad: Looking at https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler@v0.10.0, I see no useful options for the handler function, so I'm sticking with `handler.EnqueueRequestsFromMapFunc()`.

So, 2 objects per call, 1 old, 1 new. I see no way to skip the old one, which gets evaluated first. Any update on it invalidates the watch call for the new object, and triggers 2 more watch calls with the updated object.

After thinking, I see no difference between having 2 watches. I prefer not having 2 watches to reduce the number of pointers to old and new objects, so I just separated the case of orphan policies into a function call instead to reduce cyclomatic complexity.

I provided an extended explanation of my process in https://github.com/kubewarden/kubewarden-controller/issues/136#issuecomment-994738058.

raulcabello: It's ok with me to move it to a separate method and keep it in the same watch. https://github.com/kubewarden/kubewarden-controller/issues/136#issuecomment-994984217 I added a comment for the other issue, but this one looks good to me

raulcabello (approved):
LGTM Thanks!

PR #141, opened 2021-12-31
 Mutating field in the policy configmap

jvanz:
Adds a new field in the configmap where the policies info are store. The new "Mutating" field tells Policy Server that the policy can mutate requests or not.

### DIFF
--- a/internal/pkg/admission/common_test.go
+++ b/internal/pkg/admission/common_test.go
@@ -1,0 +1,43 @@
+package admission
+
+import (
+	"time"
+
+	policiesv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/policies/v1alpha2"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func createReconciler() (Reconciler, policiesv1alpha2.ClusterAdmissionPolicy, policiesv1alpha2.ClusterAdmissionPolicy) {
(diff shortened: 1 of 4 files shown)


flavio (changes requested):
I would rename `mutating` to something different, like "disable_mutation" (see https://github.com/kubewarden/policy-server/pull/139)

viccuad (commented):
Thanks for tackling this!

I wonder about the test, is it needed? The reconciler adds policies to the list regardless of their type, so we aren't testing here that they are added as validating or mutating by taking into account the new setting. To me it seems that the test should be in the end-to-end tests, and you have already added it with your work there.

flavio:
This is related with https://github.com/kubewarden/allow-privilege-escalation-psp-policy/issues/9

jvanz:
> Thanks for tackling this!
> 
> I wonder about the test, is it needed? The reconciler adds policies to the list regardless of their type, so we aren't testing here that they are added as validating or mutating by taking into account the new setting. To me it seems that the test should be in the end-to-end tests, and you have already added it with your work there.

I've added [tests](https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/5) to validate the controller and policy server together. But I think we should use the Kubebuilder  [testing infrastructure](https://kubebuilder.io/cronjob-tutorial/writing-tests.html)  to validate the controller behavior alone. I`ll take a look on it to add something here as well.

flavio (approved):
LGTM

viccuad (approved):
Approving :) (my question was about `TestCreatePoliciesMap`, as I mistakenly didn't see the need for it, but please disregard it).

jvanz:
> Approving :) (my question was about `TestCreatePoliciesMap`, as I mistakenly didn't see the need for it, but please disregard it).

Sorry, I've read your comment to fast the first times. Now I understand better what you said.

viccuad:
No problem! I misunderstood the code in the test!

viccuad:
Fixes https://github.com/kubewarden/policy-server/issues/137.

PR #144, opened 2022-01-20
Mount /tmp as an empty dir to store policies

ereslibre:
Since https://github.com/kubewarden/policy-server/pull/138, the layout of the policy-server container image has been simplified.

Its current layout looks like:

- /etc/group
- /etc/passwd
- /policy-server

Given we don't have a LSB-based layout, make /tmp be mounted by the Kubewarden controller as an empty dir.


flavio (approved):
LGTM

PR #145, opened 2022-01-20
fix(policy-server): Run the policy server with readonly root

flavio:
Ensure the Policy Server container runs with a readonly root filesystem.
This improves the security of the project.

This partially addresses https://github.com/kubewarden/helm-charts/issues/51


flavio:
While reviewing @ereslibre latest fix to the controller (the one that adds `/tmp` as an `emptyDir`), I realized it's now possible to enforce the Policy Server container to run with a readonly root filesystem

ereslibre (approved):

PR #147, opened 2022-01-21
Add initial RFC about policy signing

flavio:
This is an RFC that describes how we can implement policy verification inside of Policy Server and kwctl.

[Rendered version](https://github.com/flavio/kubewarden-controller/blob/sigstore-integration-rfc/rfc/0003-policy-signing.md)

### DIFF
--- a/rfc/0003-policy-signing.md
+++ b/rfc/0003-policy-signing.md
@@ -1,0 +1,609 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Policy Signing                   |
+| Start Date   | 21 Jan 2022                      |
+| Category     | Security                         |
+| RFC PR       | [#147](https://github.com/kubewarden/kubewarden-controller/pull/147)  |
+
+# Summary
+[summary]: #summary
+
+We want to provide a way to verify the integrity and authenticity of our policies, so that end-users can trust them.
+
+To accomplish that, we will provide ways to sign and verify policies.
+
+The main driver for this change is to be able to comply with [`SLSA`](https://slsa.dev/)
+directives about Secure Software Supply Chain.
+
+# Motivation
+[motivation]: #motivation
+
+Our policies are binary blobs, that makes hard to ensure the policy we're about to run
+matches with the source code tagged inside of a git repository.
+
+This is what we want to achieve with this proposal:
+
+* policy authors: have a way to sign their policies, to attest their provenance.
+* end users: have a way to verify the integrity of a policy that they want to load inside of their cluster. Allow them to execute only policies they trust.
+
+## Examples / User Stories
+[examples]: #examples
+
+### Execute only policies coming from trusted parties
+
+> As a Kubernates administrator,
+> I want to enforce only Kubewarden policies that have been produced by authors I trust to be loaded.
+
+### Execute only policies that have been approved by selected people
+
+> As a group of Kubernetes administrators,
+> we want to deploy in production only policies that we have approved.
+
+In that case, the Kubernetes administrators would:
+
+  * Download a pre-built Kubewarden policy
+  * Verify it's being signed by its author
+  * Push it to a (possibly internal) registry
+  * All the Kubernetes administrators would sign the policy
+  * The policy is loaded only if all the Kubernetes administrators signed the policy
+
+
+# Detailed design
+[design]: #detailed-design
+
+We're going to use [Sigstore](https://sigstore.dev) to sign and verify our policies.
+
+Sigstore allows policies to be signed either with a pub/private keypair or with
+[KEYLESS](https://github.com/sigstore/cosign/blob/main/KEYLESS.md)
+signing.
+
+## How Sigstore signatures work
+
+This is a quick overview of the different kind of signatures that could be produced by Sigstore.
+
+Even though we will use just one method to sign our policies, we will have to support
+all of them in order to allow the use cases described above. We don't know how users and 3rd party
+developers will sign their policies.
+
+It's important to know that a policy can be signed in multiple ways. Multiple people can sign the
+policy. The same user can sign it multiple times, using different methods.
+
+### Public/private key pair
+
+This is the simplest way of signing. The policy author uses his private key to sign
+the policy.
+
+The end user needs the public key of the author to verify the policy.
+
+### Keyless signing outside of GitHub Action
+
+The policy author signs the policy in keyless mode. This is the workflow:
+
+  * The author executes `cosign sign ...`
+  * Cosign prints a URL on the standard output
+  * Author clicks on the link, he's redirected to a dedicated landing page of Sigstore
+  * Author picks an identity provider from the list shown on the page (GitHub, Outlook, Google,...)
+  * Author is redirected to the identity provider, enters his credentials and allows the identity provider to share some details with Sigstore.
+    This is the same process that happens when, for example, you sign into a website using your Gmail identity
+  * The identity provider returns a OIDC token
+  * The information inside of the OIDC token is used to generate some details of the final signature
+
+As an end result, the signature will look like that:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://github.com/login/oauth",
+    "Subject": "user@provider.com"
+  }
+}
+```
+
+The end user needs to know the email address used by the policy author.
+
+### Keyless signing inside of GitHub Action
+
+Each GitHub Action runs inside of an environment where a OIDC token is provided
+by GitHub.
+
+Sigstore's cosign can leverage this already existing OIDC token to sign artifacts. This
+signature process happens in non-interactive mode, which is great for CD pipelines.
+
+This is how the signature looks like:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://token.actions.githubusercontent.com",
+    "Subject": "https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main"
+  }
+}
+```
+
+The relevant information are the `Issuer` and `Subject` fields.
+
+All the objects signed via GitHub Actions in non-interactive mode have the `Issuer` field set to
+`https://token.actions.githubusercontent.com`.
+
+The `Subject`, on the other hand, is built using:
+
+  * The repository that contains the source code of the policy
+  * The full path to the Yaml file where the `cosign sign` invocation is written
+  * The branch: `main` or a `tag`
+
+This somehow unique `Subject` is nice, because it's more granular.
+However, that makes the end user experience uglier.
+
+In the previous cases, the identity of the policy author never changed: it was
+either his public key or his email address.
+
+In this case, the `Subject` keeps changing:
+
+  * It's different on a per policy basis: each policy lives inside of a different
+    repository
+  * It's different even for the very same policy:
+    * It changes based on the release: that's because the git tag is part of the URL
+    * Even keeping the same tag (like `main` - which would be horrible from a security
+      POV), the URL could change if the Yaml file that invokes `cosign sign` is renamed.
+
+## Signing Kubewarden policies
+
+We want to sign our policies inside of the GitHub Action that takes care of pushing
+them to our container registry.
+We want to sign them in keyless mode, using the GitHub Action OIDC token.
+
+Given that our builds take place inside of GitHub Actions, the usage
+of keyless signing offers us the most secure setup we can achieve inside of this
+context.
+
+This leads us to find a solution that solves the problem of the "mutating `Subject`".
+We want end users have a good experience but, at the same time, we don't want to weaken
+the strengths of our signatures.
+
+Given a `Subject`:
+
+```json
+{
+  "Subject": "https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main"
+}
+```
+
+We can easily understand who is the owner of the repository that built the policy.
+That could be an individual user or an organization.
+
+We plan to allow end users to define an "allow list" of GitHub users/organizations
+that can be trusted.
+
+## Verifying policies
+
+Policy verification settings are going to be defined using a dedicated configuration
+file. This file format is going to be shared between Policy Server and kwctl.
+
+The proposed format is the following one:
+
+```yaml
+apiVersion: v1
+
+allOf:
+  - kind: githubAction
+    owner: kubewarden   # mandatory
+    annotations:
+      env: prod
+
+anyOf: # at least `anyOf.minimumMatches` are required to match
+  minimumMatches: 2 # default is 1
+  signatures:
+  - kind: pubKey
+    owner: flavio # optional
+    key: .... # mandatory
+    annotations:  # optional
+      env: prod
+      foo: bar
+  - kind: pubKey
+    owner: victor # optional
+    key: .... # mandatory
+  - kind: genericIssuer
+    isser: https://github.com/login/oauth
+    subjectEqual: alice@example.com
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectEqual: https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectUrlPrefix: https://github.com/flavio/
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectUrlPrefix: https://github.com/kubewarden # <- it will be post-fixed with `/` for security reasons
+  - kind: githubAction
+    owner: flavio   # mandatory
+    repo: policy1 # optional
+  - kind: pubKey
+    owner: alice # optional
+    key: .... # mandatory
+```
+
+The file is composed by two sections:
+
+  * `allOf`: all the `SignatureRequirement`s specified inside of this section are mandatory.
+    Each policy must be signed by all of the mentioned Signers
+  * `anyOf`: this is a list of `SignatureRequirement`s that are searched inside of the policy
+    signatures. The `minimumMatches` value (which has a default value of `1`)
+    defines the minimum number of verified signatures that each policy must have. This behavior can be used as a feature when migrating between signatures.
+
+
+### Types of signatures
+
+There are different types of `SignatureRequirement` objects, we're goint to describe
+them in the next sections..
+
+Before doing that, it's important to point out that each `SignatureRequirement`
+object has an optional attribute called `annotations`. This is a
+key/value map with a list of cosign annotations that must be found inside of
+the cosign Signature.
+
+For example, given the following `SignatureRequirement`:
+
+```yaml
+- kind: pubKey
+  owner: flavio # optional
+  key: .... # mandatory
+  annotations:
+    env: prod
+    foo: bar
+```
+
+This requirement is satisfied only when all these conditions are true:
+
+  * The signature is verified using the public key provided inside of the `key`
+    attribute
+  * The signature must have all the annotations specified: `env` with value `prod` and
+    `foo` with key `bar`. It doesn't matter if the signature has even more annotations.
+
+#### SignatureRequirement `pubKey`
+
+This `SignatureRequirement` targets signatures produced via a private key:
+
+```yaml
+  - kind: pubKey
+    owner: flavio # optional
+    key: |
+      -----BEGIN PUBLIC KEY-----
+      MBFKHFDGHKIJH0CAQYIKoZIzj0DAQcDQgAEX0HFTtCfTtPmkx5p1RbDE6HJSGAVD
+      BVDF6SKFSF87AASUspkQsN3FO4iyWodCy5j3o0CdIJD/KJHDJFHDFIu6sA==
+      -----END PUBLIC KEY-----
+```
+
+The `owner` attribute is optional, while the `key` one is mandatory.
+
+#### SignatureRequirement `genericIssuer`
+
+This is used to target signatures produced in keyless mode.
+
+The `issuer` attribute is mandatory, and is used to match the `Issuer`
+attribute inside of the certificate issued by Fulcio's PKI.
+
+The `SignatureRequirement` then performs a comparison operation
+abainst the `Subject` attribute of the certificate issued by Fulcio's PKI.
+These are the constrains we plan to offer since the first release:
+
+  * `subjectEqual`: the `Subject` inside of the signature must be
+    equal to the one specified inside of the `SignatureRequirement`
+  * `subjectUrlPrefix`: the value provided inside of the `SignatureRequirement`
+    is terminated with a `/` char for security reasons, unless it already ends with that. Then this
+    value must match with the prefix of the signature's `Subject`.
+
+Let's make some examples.
+
+Performing a strict check of the `Subject`:
+
+```yaml
+- kind: genericIssuer
+  isser: https://github.com/login/oauth
+  subjectEqual: alice@example.com
+
+```
+
+The following signature is going to satisfy the `SignatureRequirement`
+shown above:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://github.com/login/oauth",
+    "Subject": "alice@example.com"
+  }
+}
+```
+
+Performing a URL check:
+
+```yaml
+- kind: genericIssuer
+  issuer: https://token.actions.githubusercontent.com
+  subjectUrlPrefix: https://github.com/flavio
+```
+
+The string inside of `subjectUrlPrefix` is automatically suffixed with the `/`
+char, because of that it becomes: `https://github.com/flavio/`.
+
+The following signature is going to satisfy the `SignatureRequirement`
+shown above:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://token.actions.githubusercontent.com"
+    "Subject": "https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main"
+  }
+}
+```
+
+On the other hand, this signature is **not** going to be trusted:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://token.actions.githubusercontent.com"
+    "Subject": "https://github.com/flavio-hacker/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main"
+  }
+}
+```
+
+#### SignatureRequirement `githubAction`
+
+A lot of users are relying on GitHub Actions to implement their CD pipelines.
+We expect many policies to be signed using the seamless integration that GitHub
+Actions offer.
+
+While these signatures can be verified with the `genericIssuer` and the `subjectUrlPrefix`
+constrain, we want to offer a better user experience.
+
+Because of that, we provide the `SignatureRequirement` with type `githubAction`:
+
+```yaml
+- kind: githubAction
+  owner: flavio                  # mandatory
+  repo: policy-secure-pod-images # optional
+```
+
+This defines the following attributes:
+
+  * `owner` (required): the ID of the GitHub user or Organization to be trusted
+  * `repo` (optional): the name of the repository to be trusted
+
+The following signature will be verified:
+
+```hcl
+{
+  "critical": {
+    // not relevant
+  },
+  "optional": {
+    "Bundle": {
+    // not relevant
+    },
+    "Issuer": "https://token.actions.githubusercontent.com"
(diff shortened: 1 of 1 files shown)


rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+  - https://github.com/kubewarden/
+  "https://github.com/login/oauth":
+  - john@acme.org
+annotations:
+  env: prod
+  foo: bar
+```
+
+The file is composed by three sections, which we will analyze in the next
+sections.
```
flavio, line 232: I don't like to perform this kind of verification, I think it can lead to unexpected results.

For example, given the following configuration:

```yaml
issuers:
  "https://token.actions.githubusercontent.com":
  - https://github.com/flavio
```

Would lead to **trust** these kind of issuers:

* `https://github.com/flavio/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.0.2`
* `https://github.com/flavio-the-hacker/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.0.2`

A missing `/` inside of the setting could lead to this unexpected behaviour.

Moreover, I might be paranoid, but I think someone might come up asking for a way to provide RegExp to validate the `Subject`. I think this would be super scary from a security POV, many things could go wrong that way.

I think it would be better to provide a more strict and explicit configuration:

```yaml
issuers:
  GitHubActions:
  - owner: flavio
    repositories:
      - repo1
      - repo2
   - owner: kubewarden
```

In that case, we would trust the `Subject`s generated by:
* user `flavio`: only `repo1` and `repo2`. All the other flavio's repos would not be trusted
* org `kubewarden`: all the repos are trusted

Pros:
  * more explicit and secure
  * less chances to shoot yourself in the foot

Cons:
  * Less generic
  * Could force us to provide more ad-hoc code for other environments (GitLab, Tekton, ...)

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+    john:|
+      -----BEGIN PUBLIC KEY-----
+      MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAELKhD7F5OKy77Z582Y6h0u1J3GNA+
+      kvUsh4eKpd1lwkDAzfFDs7yXEExsEkPPuiQJBelDT68n7PDIWB/QEY7mrA==
+      -----END PUBLIC KEY-----
+```
+
+* `keys`: list of keys that are trusted. The owner's name is not relevant during
+  the verification process.
+* `threshold`: an integer number of keys whose signatures are required in order
```
flavio, line 262: I'm not 100% sure about this value, I fear it could make everything more complicated

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+
+Subject | Verified | Reason |
+------------------|----------|--------|
+`https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main` | ✅ | The policy is coming from `https://github.com/flavio/`
+`https://github.com/flavio/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.0.2` | ✅ | The policy is coming from `https://github.com/flavio/`
+`https://github.com/kubewarden/privileged-pods/.github/workflows/release.yml@refs/tags/v0.2.0` | ✅ | The policy is coming from `https://github.com/kubewarden/`
+`https://github.com/kubewarden/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.3.0` | ✅ | The policy is coming from `https://github.com/kubewarden/`
+`https://github.com/alice/fancy-policy/.github/workflows/cd.yml@refs/heads/main` | ✅ | The policy is coming from `https://github.com/alice/fancy-policy/`
+`https://github.com/alice/another-policy/.github/workflows/cd.yml@refs/heads/main` | ❌ | The policy is coming from a repository owned by Alice, but not trusted by the configuration
+
```
flavio, line 298: I don't like to perform this kind of verification, I think it can lead to unexpected results.

For example, given the following configuration:

```yaml
issuers:
  "https://token.actions.githubusercontent.com":
  - https://github.com/flavio
```

Would lead to **trust** these kind of issuers:

* `https://github.com/flavio/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.0.2`
* `https://github.com/flavio-the-hacker/psp-capabilities/.github/workflows/release.yml@refs/tags/v0.0.2`

A missing `/` inside of the setting could lead to this unexpected behaviour.

Moreover, I might be paranoid, but I think someone might come up asking for a way to provide RegExp to validate the `Subject`. I think this would be super scary from a security POV, many things could go wrong that way.

I think it would be better to provide a more strict and explicit configuration:

```yaml
issuers:
  GitHubActions:
  - owner: flavio
    repositories:
      - repo1
      - repo2
   - owner: kubewarden
```

In that case, we would trust the `Subject`s generated by:
* user `flavio`: only `repo1` and `repo2`. All the other flavio's repos would not be trusted
* org `kubewarden`: all the repos are trusted

Pros:
  * more explicit and secure
  * less chances to shoot yourself in the foot

Cons:
  * Less generic
  * Could force us to provide more ad-hoc code for other environments (GitLab, Tekton, ...)

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+  "https://token.actions.githubusercontent.com":
+  - https://github.com/flavio/
+  - https://github.com/kubewarden/
+  - https://github.com/alice/fancy-policy/
+  "https://github.com/login/oauth":
+  - john@acme.org
+```
+
+The string entered by the user is then used to match the beginning of the `Subject`
+found inside of a signature.
```
ereslibre, line 286: I think it should be interpreted as an URL, and matched as an URL when checked. So, we always automatically append the trailing slash. URL's are always checked with trailing slashed appended. Comparison is not a string-based one.

This helps with your concern in https://github.com/kubewarden/kubewarden-controller/pull/147/files#r790110976

rfc/0003-policy-signing.md
ereslibre: I think that if we parse the list of subjects as URL's when they are not e-mail addresses this should not be a problem.

This being said: we can support both if we want. In a model in which we allow different keys as `issuers` children that might behave different:

```yaml
issuers:
  GitHubActions:
  - owner: flavio
    repositories:
      - repo1
      - repo2
  - owner: kubewarden
```

And also:

```yaml
issuers:
  GenericIssuers:
    "https://token.actions.githubusercontent.com":
      - https://github.com/flavio/repo1
      - https://github.com/flavio/repo2
      - https://github.com/kubewarden
```

Would be equivalent.

With this you have the best of both worlds I think. However, I doubt we really need to start with this from the early beginning. If not, then the question would be: should we start with `GenericIssuers` or with `GitHubActions` only? In my opinion, I would prefer to start with `GenericIssuers` only or with both.

I don't like that we have a single way of specifying issuers and that is tied with GitHub without an alternative.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+**Note well:** `issuers` defines a per-issuer allow list. That means the policies
+**do not have** to be signed by all the subjects mentioned inside of the list.
+They must all be signed by at least one of these issuers.
+
+### Annotations
+
+This is an entirely optional section that applies both to key and keyless
+signatures.
+
+Cosign allows the signer to specify custom key-value attributes to embedded
```
ereslibre, line 313:
```suggestion
Cosign allows the signer to specify custom key-value attributes to be embedded
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+
+There's no reason to not implement this feature.
+
+# Alternatives
+[alternatives]: #alternatives
+
+> - What other designs/options have been considered?
+> - What is the impact of not doing this?
+
+## Specify per-policy signature settings
```
ereslibre, line 366: I think having `annotations` in a way that they can be set in `ClusterAdmissionPolicy` is interesting.

A PolicyServer `ps1` might specify a "baseline" of `annotations`: `["annotation1", "annotation2"]`.

If a `ClusterAdmissionPolicy` specifies specific policy annotations `["extra1", "extra2"]`, then the policy must meet `["annotation1", "annotation2", "extra1", "extra2"] for it to be schedulable to `ps1`. Otherwise, it can't.

Is not an "alternative" to the main RFC, just a way to make it more expressive and configurable per-policy.

ereslibre (commented):
Looking good. Some comments.

rfc/0003-policy-signing.md
viccuad: I would prefer adding a root key `required_signatures`, where values can be `allOf` (all info listed here needs to be verified for accepting), or `anyOf` (at least 1 of all info listed here needs to be verified for accepting). Reminiscent of Json schema keywords that are the same.

An other option is to add 2 root keys, `allOf` and `anyOf`. Both or 1 of them can be specified, and both accept `pub_keys`, `keyless`, or `annotations` as children:

```yaml
allOf: # all info listed here needs to be verified for accepting
  pub_keys:
    key-name-irrelevant: |
              -----BEGIN PUBLIC KEY-----
              MBFKHFDGHKIJH0CAQYIKoZIzj0DAQcDQgAEX0HFTtCfTtPmkx5p1RbDE6HJSGAVD
              BVDF6SKFSF87AASUspkQsN3FO4iyWodCy5j3o0CdIJD/KJHDJFHDFIu6sA==
              -----END PUBLIC KEY-----

anyOf: # at least 1 of all info listed here needs to be verified for accepting
  keyless:
    - issuer: https://token.actions.githubusercontent.com # default tuple, we add it to our minimum config
      email: kubewarden@suse.de
  annotations:
    env: prod
    foo: bar
```

But with this second option we need to think if we could create any logic inconsistency. Which I think we do not, unless we add a `oneOf` key, that would signify "only one of the info provided".

rfc/0003-policy-signing.md
viccuad: I also favor the approach of just having a generic issuer for now, and look for either url or email. I'm not sure on which yaml would be better UX so that we look for either url or email. Either having a `genericURLIssuer`, `genericEmailIssuer` or the like, or have tuples `issuer, email`, `issuer, url`.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+`https://github.com/alice/another-policy/.github/workflows/cd.yml@refs/heads/main` | ❌ | The policy is coming from a repository owned by Alice, but not trusted by the configuration
+
+Given the previous configuration, we would also trust all the policies that have been
+signed with the `john@acme.org` email, as long as the identity provider is the
+GitHub website (this is the identity provider used by "human-driven" keyless signatures
+relying on the GitHub identity provider).
+
+**Note well:** `issuers` defines a per-issuer allow list. That means the policies
+**do not have** to be signed by all the subjects mentioned inside of the list.
+They must all be signed by at least one of these issuers.
```
viccuad, line 306: I think this might be confusing. I would prefer to just have tuples of `(issuer, email)`, `issuer, url)`, and use `anyOf`/ `allOf`:

```yaml
anyOf: # at least 1 of all info listed here needs to be verified for accepting
  keyless:
    - issuer: https://token.actions.githubusercontent.com
      url: https://github.com/kubewarden
    - issuer: https://oauth2.suse.com/auth
      email: kubewarden@suse.de
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+
+| `public_keys` | `issuers` | annotations | Behaviour |
+|-------------|-----------|-------------|-----------|
+| ✅ | ❌ | ❌ | policies must be signed by at least <threshold> keys |
+| ✅ | ✅ | ❌ | policies must be signed by at least <threshold> keys, **plus** they must be signed by one of the issuers |
+| ❌ | ✅ | ❌ | policies must be signed by one of the issuers |
+| ❌ | ✅ | ✅ | policies must be signed by one of the issuers, **plus** they must satisfy all the annotations |
+| ✅ | ❌ | ✅ | policies must be signed by at least <threshold> keys, **plus** they must satisfy all the annotations |
+| ✅ | ✅ | ✅ | policies must be signed by at least <threshold> keys, plus they must be signed by one of the issuers, **plus** they must satisfy all the annotations |
+| ❌ | ❌ | ✅ | this configuration is wrong, it should result in an error |
```
viccuad, line 344: With the `anyOf` approach, the last configuration would be possible (just, checking that there's any signature that marks the images as `prod` for example, regardless of the signer).

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+### Annotations
+
+This is an entirely optional section that applies both to key and keyless
+signatures.
+
+Cosign allows the signer to specify custom key-value attributes to embedded
+into the signature.
+
+```yaml
+annotations:
```
viccuad, line 317: I wonder if users are going to confuse these signing annotations with other type of annotations.

rfc/0003-policy-signing.md
viccuad: To me, the RBAC means that it's only useful for cluster admins, which diminishes returns.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+issuer.cosign.sigstore.dev: https://token.actions.githubusercontent.com
+```
+
+At verification time, we could look for a signature that satisfies the constrains provided inside of the policy metadata.
+
+Alterning the policy's metadata would lead to a different `.wasm` file, which would
+break the initial checks of sigstore's signature.
+
+We've discarded this approach because, among other things, we think it's safer to provide the verification constraints
+in an out-of-band way.
```
viccuad, line 427: You could add that it makes having reproducible builds of policy images more difficult, if not impossible.

rfc/0003-policy-signing.md
ereslibre: Not sure what do you mean with RBAC. Can you elaborate on that? My comment is not applicable to RBAC, but to ensure that labels can be provided per policy server and per policies. Then, a policy must have been signed with all the key=value required both in the policy-server where it's to be scheduled, and on the policy CR itself.

rfc/0003-policy-signing.md
viccuad: With `anyOf`, `allOf`, we may just drop this section.

ereslibre: I am also in favour of the `anyOf` and `allOf` approach in principle. I guess they are exclusive with each other, right? In that you either provide `anyOf` or `allOf`, and beneath the one that has been chosen the list of signatures. Right?

rfc/0003-policy-signing.md
viccuad: In my opinion the shortcomings already listed for this in the RFC make it ugly.

Also, if we have per-policy annotations/signatures, it seems only useful for users that can deploy their own ClusterAdmissionPolicies, but don't have access to the PolicyServer. 
I would be happier using the verification config only in PolicyServers, and providing NamespacedPolicyServers. 
If what we want is a tiered system of verification config, I would prefer allowing PolicyServers to read a list of Configmaps, and some of them from a different Namespace. That way one can use the ns permissions.

ereslibre: Let's ignore my comment for now. I was aiming at also having the same behaviors when we introduce the `AdmissionPolicy` namespaced resource. But I have to admit I don't have a strong story to show now, just a hunch that it would be useful to allow. It's fine to ignore it right now.

rfc/0003-policy-signing.md
viccuad: They aren't exclusive, you could provide both. And one can list `pub_keys`, `keyless tuples`, or `annotations` on both of them.

ereslibre: At first it looked a bit confusing, but if we make it clear that if both are provided then both should be met I think we are fine.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+| Start Date   | 21 Jan 2022                      |
+| Category     | Security                         |
+| RFC PR       | [#147](https://github.com/kubewarden/kubewarden-controller/pull/147)  |
+
+# Summary
+[summary]: #summary
+
+> Brief (one-paragraph) explanation of the feature.
+
+We want to provide a way to verify the integrity and authenticity of our policies, so that end-users can trust them.
```
kkaempf, line 13: This RFC should mention https://slsa.dev as the 'directing force' behind our attempts to improve security of the software supply chain.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+> - What is the expected outcome?
+>
+> Describe the problem you are trying to solve, and its constraints, without coupling them too closely to the solution you have in mind. If this RFC is not accepted, the motivation can be used to develop alternative solutions.
+
+Our policies are binary blobs, that makes hard to ensure the policy we're about to run
+matches with the source code tagged inside of a git repository.
+
+This is what we want to achieve with this proposal:
+
+* policy authors: have a way to sign their policies, to attest their provenance.
```
kkaempf, line 32: Hmm, policy authors are probably more interested in signing-off their source code commits.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+* end users: have a way to verify the integrity of a policy that they want to load inside of their cluster. Allow them to execute only policies they trust.
```
kkaempf, line 33: Wouldn't that be a policy in itself. End users shouldn't be bothered about the internals. They just "trust this source/author" and rely on tools/ui to honor this trust.

kkaempf: We also need to split "end user" into (at least two) sub-categories
- cluster administrators - they set policies and ultimately decide on the rules workloads (and policies) have to obey.
- cluster users - they want their workload to run and usually follow the install instructions says "download and run this binary". Policies, signing, etc. just get into their way.

kkaempf: At the end of the day, none of them (authors, admins, users) will care about signing. Authors want their software to be easily deployed, admins want less administrative work, and users want to run software. It's the CIOs and CSO's of this world, asking for secure software chains.

That said, signing must be unobtrusive, ideally invisible, to get accepted.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+
+### Execute only policies that have been approved by selected people
+
+> As a group of Kubernetes administrators,
+> we want to deploy in production only policies that we have approved.
+
+In that case, the Kubernetes administrators would:
+
+  * Download a pre-built Kubewarden policy
+  * Verify it's being signed by its author
```
kkaempf, line 53: Verification shouldn't be an extra step for the admin. `zypper` and the open build service (OBS) is a good role model here. The OBS does all the key generation and signing of repos, no extra tasks for software authors. Adding a repo is where trust is verified. Package installation from a trusted repo never asks about trust.

Policy signing should follow a similar model. Explicit verification when a policy registry gets added, implicit (hidden) verification when a policy is downloaded.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+  * Push it to an internal registry
+  * All the Kubernetes administrators would sign the policy
```
kkaempf, line 55: What's the purpose of this step ?

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+
+This is the simplest way of signing. The policy author uses his private key to sign
+the policy.
+
+The end user needs the public key of the author to verify the policy.
+
+### Keyless signing outside of GitHub Action
+
+The policy author signs the policy in keyless mode. This is the workflow:
+
```
kkaempf, line 93: This workflow is ok for an initial implementation but needs to be simplified (from an authors pov) in future iterations

kkaempf (changes requested):
We might need two RFCs, one covering the 'user interface' of policy signing (briefly covered here) and one the implementation details (the bulk of this rfc).

Let's focus on implementation details first and improve usability step-by-step as we learn.

rfc/0003-policy-signing.md
flavio: Good point, added to the next iteration

rfc/0003-policy-signing.md
flavio: Policy authors are interested about this topic if they want to implement something like SLSA

rfc/0003-policy-signing.md
flavio: Admission policies (like the ones of kubewarden), have great power. They can be abused to do nasty things.

Because of that, kubernetes administrators might want to ensure only vetted policies can be loaded. Signing a policy with a personal key is a strong way to vet a policy.

rfc/0003-policy-signing.md
flavio: JFYI: this is a workflow that is imposed by `cosign`. Keyless singning inside of GitHub actions is super transparent, I guess this is what policy authors will use the most.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,440 @@
...
+    * Even keeping the same tag (like `main` - which would be horrible from a security
+      POV), the URL could change if the Yaml file that invokes `cosign sign` is renamed.
+
+## Signing Kubewarden policies
+
+We want to sign our policies inside of the GitHub Action that takes care of pushing
+them to our container registry.
+We want to sign them in keyless mode, using the GitHub Action OIDC token.
+
+This is the most secure setup we can imagine.
```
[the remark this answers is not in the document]
flavio: Good point, I clarified

rfc/0003-policy-signing.md
flavio: added

rfc/0003-policy-signing.md
flavio: I would prefer to the keep what is described into the RFC, and just add at the end of the urls that users are providing.

To clarify:

```yaml
issuers:
  "https://token.actions.githubusercontent.com":
  - https://github.com/flavio/
  - https://github.com/kubewarden # -> this becomes https://github.com/kubewarden/
  "https://github.com/login/oauth":
  - john@acme.org
```

rfc/0003-policy-signing.md
flavio: I like this proposal, but I'm not convinced about adding `annotations` under `allOf` and `anyOf`. I would leave them as top level.

flavio: So, to summarize:

I would use this format:


```yaml
allOf: # all info listed here needs to be verified for accepting
  pub_keys:
    key-name-irrelevant: |
              -----BEGIN PUBLIC KEY-----
              MBFKHFDGHKIJH0CAQYIKoZIzj0DAQcDQgAEX0HFTtCfTtPmkx5p1RbDE6HJSGAVD
              BVDF6SKFSF87AASUspkQsN3FO4iyWodCy5j3o0CdIJD/KJHDJFHDFIu6sA==
              -----END PUBLIC KEY-----

anyOf: # at least 1 of all info listed here needs to be verified for accepting
  issuers:
    "https://token.actions.githubusercontent.com":
    - url: https://github.com/flavio/
    - url: https://github.com/kubewarden # <- this becomes https://github.com/kubewarden/
    "https://github.com/login/oauth":
    - email: john@acme.org
 
annotations:
 env: prod
 foo: bar
```

flavio:
@viccuad , @raulcabello , @jvanz and @ereslibre : I've updated the PR to include the new format of the `verification.yml` file.

I'm a bit tired, so please forgive the grammar errors I might have left around...

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+>
+> This section should cover architecture aspects and the rationale behind disruptive technical decisions (when applicable), as well as corner-cases and warnings.
+
+We're going to use [Sigstore](https://sigstore.dev) to sign and verify our policies.
+
+Sigstore allows policies to be signed either with a pub/private keypair or with
+[KEYLESS](https://github.com/sigstore/cosign/blob/main/KEYLESS.md)
+signing.
+
+## How Sigstore signature work
```
viccuad, line 74:
```suggestion
## How Sigstore signatures work
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+directives about Secure Software Supply Chain.
+
+# Motivation
+[motivation]: #motivation
+
+> - Why are we doing this?
+> - What use cases does it support?
+> - What is the expected outcome?
+>
+> Describe the problem you are trying to solve, and its constraints, without coupling them too closely to the solution you have in mind. If this RFC is not accepted, the motivation can be used to develop alternative solutions.
```
viccuad, line 27: I would drop all templated explanations, makes it shorter, and their are obvious.

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+[KEYLESS](https://github.com/sigstore/cosign/blob/main/KEYLESS.md)
+signing.
+
+## How Sigstore signature work
+
+This is a quick overview of the different kind of signatures that could be produced by Sigstore.
+
+Even though we will use just one method to sign our policies, we will have to support
+all of them in order to allow the use cases described above. We don't know how users and 3rd party
+developer will sign their policies.
```
viccuad, line 80:
```suggestion
developers will sign their policies.
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+```yaml
+apiVersion: v1
+
+allOf:
+  - kind: githubAction
+    owner: kubewarden   # mandatory
+    annotations:
+      env: prod
+
+anyOf: # 1+ are required to match
```
viccuad, line 219:
```suggestion
anyOf: # at least `anyOf.minimumMatches` are required to match
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+    subjectEqual: alice@example.com
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectEqual: https://github.com/flavio/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectUrlPrefix: https://github.com/flavio/
+  - kind: genericIssuer
+    issuer: https://token.actions.githubusercontent.com
+    subjectUrlPrefix: https://github.com/kubewarden # <- we add the ending `/` behind the scene
```
viccuad, line 242:
```suggestion
    subjectUrlPrefix: https://github.com/kubewarden # <- it will be post-fixed with `/` for security reasons
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+    annotations:
+      env: prod
+
+anyOf: # 1+ are required to match
+  minimumMatches: 2 # default is 1
+  signatures:
+  - kind: pubKey
+    owner: flavio # optional
+    key: .... # mandatory
+    annotations:
```
viccuad, line 225:
```suggestion
    annotations:  # optional
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+    key: .... # mandatory
+```
+
+The file is composed by two sections:
+
+  * `allOf`: all the `SignatureRequirement`s specified inside of this section are mandatory.
+    Each policy must be signed by all of the mentioned Signers
+  * `anyOf`: this is a list of `SignatureRequirement`s that are searched inside of the policy
+    signatures. The `minimumMatches` value (which has a default value of `1`)
+    defines the minumum numbers of verified signatures that each policy must have.
```
viccuad, line 257:
```suggestion
    defines the minimum numbers of verified signatures that each policy must have. This behavior can be used as a feature when migrating between signatures.
```

rfc/0003-policy-signing.md
```diff
@@ -0,0 +1,636 @@
...
+attribute inside of the certificate issued by Fulcio's PKI.
+
+The `SignatureRequirement` then performs a comparison operation
+abainst the `Subject` attribute of the certificate issued by Fulcio's PKI.
+These are the constrains we plan to offer since the first release:
+
+  * `subjectEqual`: the `Subject` inside of the signature must be
+    equal to the one specified inside of the `SignatureRequirement`
+  * `subjectUrlPrefix`: the value provided inside of the `SignatureRequirement`
+    is terminated with a `/` char, unless it already ends with that. Then this
```
viccuad, line 318:
```suggestion
    is terminated with a `/` char for security reasons, unless it already ends with that. Then this
```

viccuad (approved):
LGTM, just some minor rewordings and typos.

rfc/0003-policy-signing.md
flavio: Closing, this is addressed with the latest iteration of the doc

rfc/0003-policy-signing.md
flavio: Now obsolete, because of the changes done to `verification.yml` format

rfc/0003-policy-signing.md
flavio: addressed with the latest rev

rfc/0003-policy-signing.md
flavio: Now obsolete, because of the changes done to `verification.yml` format

rfc/0003-policy-signing.md
flavio: Now obsolete, because of the changes done to `verification.yml` format

flavio:
I think we can go ahead and merge this RFC. OFC I'll squash the history before doing that.

What do you think?

jvanz (approved):

PR #150, opened 2022-01-24
Turn ci.yml a callable workflow.

jvanz:
Updates the ci.yml adding the configuration to allow other workflows call it to run the tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_call:
   push:
     branches:
     - '*'


ereslibre (approved):
Oops thanks!

viccuad (changes requested):
I like the idea, but I'm not sure that reusing `make test` (the unit tests) is useful. In fact, they will run `make manifest`, etc. Is that what we want?

I would expect a new `make e2e-test`, that runs things similar to what there is right now in `e2e-tests/`.

ereslibre:
> I would expect a new make e2e-test, that runs things similar to what there is right now in e2e-tests/.

Yes, I think we should run e2e-tests as well. I don't think there's anything wrong with running `make test` though, just an extra check that needs to pass before we call a release, besides the e2e-tests (that as said, I think we should add).

Thanks for realizing about that!

viccuad:
Oh, from the conversation we had, I was thinking about reusing the workflow from the end-to-end-tests repo, not just reusing within this repo on the release workflow. Makes sense.

PR #174, opened 2022-02-21
Update policy signing RFC

flavio:
Update the RFC to reflect the final design of the yml file


viccuad (approved):
Thanks! forgot to do that after the merge!

viccuad (approved):

PR #176, opened 2022-02-23
[RFC] Policy reports: user stories

ereslibre:
This is the initial work to settle on the User Stories about the "Policy Reports" feature.

After we agree on the User Stories, we can work on being specific about the implementation on the RFC section that is specific to that.


rfc/0005-policy-reports.md
```diff
@@ -0,0 +1,110 @@
...
+|:-------------|:--------------------------------|
+| Feature Name | Policy reports                  |
+| Start Date   | Feb 22nd 2022                   |
+| Category     | Policy Enforcement              |
+| RFC PR       | [fill this in after opening PR] |
+
+# Summary
+[summary]: #summary
+
+> Brief (one-paragraph) explanation of the feature.
```
viccuad, line 11: Nit: I would drop the boilerplate code for filled sections.

rfc/0005-policy-reports.md
```diff
@@ -0,0 +1,110 @@
...
+
+> As a UI integrator, I can show rich stats about evaluations without
+> the need of doing any computations, just by consuming reports, which
+> have a well known, stable and versioned structure.
+
+> As a third party integrator, the interface to fetch Kubewarden stats
+> is the well known and versioned Report resource. I can use this
+> information from Kubewarden to aggregate to other security relevant
+> data about the cluster in a hollistic security-focused control
+> plane.
```
viccuad, line 74: Does the user here expect low latency or pull behaviour, for automated/machine readable usecases? Maybe it is worth it to clarify.
For pull behaviour, we could always support setting annotation on the Report, that would trigger the controller to recompute it.

viccuad (commented):
Thanks! User stories look good to me.
Do we want to expand to include searching by resource/CRUD operation (instead of searching by policy/set of policies)?

flavio (approved):
I'm fine with the current proposal. I would keep things simple, I see the risk of re implementing what prometheus is doing

raulcabello (approved):
LGTM

nunix:
For me, the use cases are OK. I would suggest to change/rephrase some of the use-cases in order to make it clear what's the goal (i.e. what is the whole stack?).
And while there's no impact in terms of requests handling when we are in monitor mode, it still (might) impact the requests/cluster performances (as discussed during the last meeting).

rfc/0005-policy-reports.md
ereslibre: The idea is that the report is updated by a component when requests arrive, or by buffering events and dumping them onto the report from time to time (e.g. every 30 seconds or 1 minute).

This resource is for informational purposes and for the consumer to have a general overview of what is happening/would happen with policies on the cluster.

ereslibre:
> I would suggest to change/rephrase some of the use-cases

Done.

> And while there's no impact in terms of requests handling when we are in monitor mode, it still (might) impact the requests/cluster performances (as discussed during the last meeting).

I think we can discuss this on the implementation side. I would suggest that we also update the report on `protect` mode (both in `monitor` and `protect` mode). This way, numbers are always up to date and can give you certain ideas on third-party integrators that might use this information to cross data or show a dashbard to the user. It really depends on how the implementation is driven, I think we can pull an implementation that won't impact Kubewarden's performance -- at some cost, slightly increased memory usage --. But I think we can discuss that later on if that's fine.

ereslibre:
> Do we want to expand the users stories to allow searching by resource/CRUD operation (in addition of searching by policy/set of policies)?

I think this is entirely up to the consumer of the reports. They are free to list/watch the reports custom resource and build their own indexes etc...

ereslibre:
If you are all fine with it, I propose to merge this RFC as is, and work on top of it later on.

viccuad (approved):

jvanz (approved):
LGTM! :D

Issue #179, opened 2022-02-25
RFC: have a unified version for the whole Kubewarden stack


flavio:
We would like to have a way to easily identify the version of Kubewarden a user is installing/running. Right now each component of kubewarden has its own different version number.

## Acceptance criteria

* Create a RFC that proposes a way to version Kubewarden that simplifies life of users, supporters and marketing

jvanz:
@ereslibre should we keep this in pending review? I saw you closing the PR, can we move this to TODO? Or are we keeping this on review to discuss the card in a call?

ereslibre:
Moved it to TODO, thank you.

PR #182, opened 2022-03-03
[RFC] Add Kubewarden versioning proposal

ereslibre:
Fixes https://github.com/kubewarden/kubewarden-controller/issues/179

### DIFF
--- a/rfc/0006-kubewarden-versioning.md
+++ b/rfc/0006-kubewarden-versioning.md
@@ -1,0 +1,203 @@
+|              |                                 |
+|:-------------|:--------------------------------|
+| Feature Name | Kubewarden versioning           |
+| Start Date   | Mar 3rd 2022                    |
+| Category     | Versioning                      |
+| RFC PR       | [fill this in after opening PR] |
+
+# Summary
+[summary]: #summary
+
+Kubewarden is composed of many core components that work together in
+order to build the final product. This is a proposal to harmonize
+Kubewarden versioning schema.
+
+# Motivation
+[motivation]: #motivation
+
+Having a way to specify what version of Kubewarden is being installed
+or executed would make it easier for the Kubewarden team to describe,
+market and support Kubewarden (as in user support).
+
+It would help users in understanding what version of Kubewarden they
+are about to install or execute.
+
+## Examples / User Stories
+[examples]: #examples
+
+> Users do not need to keep track of component versions and how they
+> work together. They just need to care about running a specific
+> Kubewarden version.
+
+> Users are able to easily understand that a new version of Kubewarden
+> is available, and that more than one component might be upgraded in
+> order to move to the new Kubewarden version.
+
+> A user has an issue with Kubewarden. Given its multiple component
+> nature, we would like to know what was the Kubewarden version that
+> got installed, and through what versions it was upgraded.
+
+# Detailed design
+[design]: #detailed-design
+
+## Many components, one experience
+
+Kubewarden core components, which are exposed to the user are:
+
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
+- kwctl
+
+As a second dimension, there are also policies and their respective
+SDK's. This proposal is going to ignore this dimension in purpose,
+because Kubewarden runtime components (`policy-server` and `kwctl`)
+should be compatible with policies previously built, even with older
+language SDK's.
+
+### Public interfaces
+
+Our public interfaces with users are as follows:
+
+- Helm Charts
+  - `values.yaml`
+- Kubewarden Controller
+  - CLI args
+  - Custom Resource Definitions
+    - Versioned
+- Policy Server
+  - CLI args
+  - Configuration
+    - sigstore related files
+    - Usually handled by the Kubewarden Controller. Regardless we
+      should treat this as public interface, given some users might
+      run the Policy Server independent of the Kubewarden Controller,
+      and might configure it with the configuration file.
+- kwctl
+  - CLI args
+  - Configuration
+    - `sources.yaml`
+    - `docker.json`
+    - sigstore related files
+
+There is also the `waPC` contract between the `policy-evaluator`
+(consumed by the Policy Server and kwctl) and Policies and SDK's.
+
+## Kubewarden toplevel version
+
+This proposal envisions using [`CalVer`](https://calver.org/) to
+version the Kubewarden project.
+
+Had Kubewarden be released today, it would be `Kubewarden
+22.03`. This makes it trivial to understand how old your version of
+Kubewarden is, and how much your installation is drifting from the
+latest releases.
+
+## Kubewarden component versions
+
+Every component in Kubewarden might have its own versioning. The
+Kubewarden maintainers are going to maintain a list of all releases.
+
+This list of releases looks as follows:
+
+```yaml
+"22.03":
+  - component: kubewarden-crds-chart
+    version: 0.1.1
+  - component: kubewarden-controller-chart
+    version: 0.3.6
+  - component: kubewarden-controller
+    version: v0.4.5
+  - component: policy-server
+    version: v0.2.6
+  - component: kwctl
+    version: v0.2.5
+"22.05":
+  - component: kubewarden-crds-chart
+    version: 0.2.0
+  - component: kubewarden-controller-chart
+    version: 0.4.0
+  - component: kubewarden-controller
+    version: v0.4.8
+  - component: policy-server
+    version: v0.3.0
+  - component: kwctl
+    version: v0.3.1
+...
+```
+
+As more components are relevant they can be added to the list as time
+grows.
+
+This map allows us to generate documentation and a single source of
+truth to refer to when a user has a problem, or when we are debugging
+an issue. It helps to understand if any component has drifted from the
+version we are expecting.
+
+The helm chart will be responsible of storing the "Kubewarden version"
+in a resource, like a ConfigMap. So it's easy to inspect it and
+understand what "Kubewarden version" a cluster is running.
+
+Had other external components such as `kwctl` need to synchronize with
+the cluster, they can inspect this ConfigMap and realize if they are
+compatible with the running version of Kubewarden on the cluster they
+are connecting to.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It's not a very common versioning scheme. I would consider reading
+  ["When to use
+  CalVer"](https://calver.org/#when-to-use-calver). Although
+  Kubewarden does not have a "large" scope, it's to some extent
+  "constantly changing", and this constant change would imply
+  harmonizing versions of all components continuously.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Have a regular Kubewarden toplevel version
+
+Having a regular --even semver based-- version on the toplevel project
+seems like an artificial construct. What is the meaning of patchlevel
+release in this case? What is Kubewarden 1.2.3 vs 1.2.4?
+
+Thus, ignoring patchlevel (e.g. just keeping 1.2 vs 1.3 vs 1.4, or
+even 1 vs 2 vs 3) is an option. But then, if it's just meaningless
+numbers, why not make the toplevel number meaningful, referring to the
+date when the project was released. This brings instant information to
+the user as opposed to version numbers.
+
+## Sync all component versions
+
+Given the nature of multiple components of the Kubewarden projects,
+scattered across different repositories, closely related to each
+other, enforcing a single synced version on all of them seem
+artificial and heavily increases releasing efforts, given a problem or
+feature in on one component forces to tag and release the rest of the
+components.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should day versions be included in the toplevel map?
+
+    To make things even easier, versions might start with
+    YEAR.MONTH. If there is the need to make a second global release
+    during the same month, then a day is appended: YEAR.MONTH.DAY1. If
+    a third release is needed during that month: YEAR.MONTH.DAY2.
+
+    This makes it easier for maintainability of upgrades and is easier
+    to describe what should the upgrade path be if there is the
+    necessity of forcefully sequence upgrades in a certain order.
+
+- Should complete versions of components (MAJOR.MINOR.PATCH) be
+  referenced by the versions map as the example shows?
+
+    The alternative would be that we use MAJOR.MINOR, and let PATCH be
+    automatically upgradeable (e.g. through the helm-chart, by releasing
+    patchlevel versions of the helm-chart that only bump pathlevel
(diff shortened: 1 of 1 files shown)


ereslibre:
Cannot request the review from the documentation group as a whole, GitHub won't let me.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+
+As a second dimension, there are also policies and their respective
+SDK's. This proposal is going to ignore this dimension in purpose,
+because Kubewarden runtime components (`policy-server` and `kwctl`)
+should be compatible with policies previously built, even with older
+language SDK's.
+
+### Public interfaces
+
+Our public interfaces with users are as follows:
```
kkaempf, line 60: The use of 'users' in this context might be too generic, as we have
- policy consumers (aka end users, who try to deploy something on k8s and could fail due to a policy)
- cluster admins (those who define which policies are enforced)
- policy authors

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+    - sigstore related files
+
+There is also the `waPC` contract between the `policy-evaluator`
+(consumed by the Policy Server and kwctl) and Policies and SDK's.
+
+## Kubewarden toplevel version
+
+This proposal envisions using [`CalVer`](https://calver.org/) to
+version the Kubewarden project.
+
```
kkaempf, line 90: I don't see how calver would support modifiers, like -pre or -alpha etc.

kkaempf: calver also doesn't show if a new version is a patch release, adds a new feature, or a breaking change.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+## Have a regular Kubewarden toplevel version
+
+Having a regular --even semver based-- version on the toplevel project
+seems like an artificial construct. What is the meaning of patchlevel
+release in this case? What is Kubewarden 1.2.3 vs 1.2.4?
+
```
kkaempf, line 164: Isn't semver used everywhere in the Kubernetes ecosystem ? Do we have (well known) projects _not_ using semver ?

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
+- kwctl
+
+As a second dimension, there are also policies and their respective
+SDK's. This proposal is going to ignore this dimension in purpose,
+because Kubewarden runtime components (`policy-server` and `kwctl`)
+should be compatible with policies previously built, even with older
```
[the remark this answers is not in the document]
ereslibre: Our promise is that you will be able to run policies after you have built them, no matter what.

We might have to break our promise at some point, but the idea is that won't have to.

rfc/0006-kubewarden-versioning.md
ereslibre: With calver we only release final versions. Is the "umbrella" toplevel Kubewarden version.

We can release -pre, -alpha of certain components if we desire.

rfc/0006-kubewarden-versioning.md
ereslibre: It is. I am not claiming otherwise. This proposal is not aiming to move all components to calver, only the toplevel Kubewarden version we refer to. "Do you have Kubewarden 22.03, or is it 22.05?"

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+  - component: kubewarden-controller-chart
+    version: 0.4.0
+  - component: kubewarden-controller
+    version: v0.4.8
+  - component: policy-server
+    version: v0.3.0
+  - component: kwctl
+    version: v0.3.1
+...
+```
```
raulcabello, line 127: If a user wants to install `v22.05` in this example, he need to look into this list and then install `kubewarden-crds-chart` `v0.2.0` and `kubewarden-controller-chart` `v0.4.0`, which then it would install the right version of the controller and policy-server, right? I'm thinking it might be easier for user if we also use `v22.05` for the helm charts. Then it would be clear which version they are installing, and the version could be check with a `helm ls`

ereslibre: We might have a "toplevel" Kubewarden chart. This chart might be autogenerated based on this umbrella version.

So that we have the global Kubewarden chart, that depends on the specific semver-versioned charts kubewarden-crds and kubewarden-controller charts. However, those charts are still semver versioned. Only the "toplevel" Kubewarden chart would be calver versioned.

raulcabello: I like the idea of the toplevel chart

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+[design]: #detailed-design
+
+## Many components, one experience
+
+Kubewarden core components, which are exposed to the user are:
+
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
+- kwctl
```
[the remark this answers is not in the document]
viccuad: I agree with Rafa here.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+version we are expecting.
+
+The helm chart will be responsible of storing the "Kubewarden version"
+in a resource, like a ConfigMap. So it's easy to inspect it and
+understand what "Kubewarden version" a cluster is running.
+
+Had other external components such as `kwctl` need to synchronize with
+the cluster, they can inspect this ConfigMap and realize if they are
+compatible with the running version of Kubewarden on the cluster they
+are connecting to.
```
viccuad, line 144: who creates that Configmap to track the Kubewarden version? (what happens if the user deploys the controller on its own, or policy-server, even if not officially supported? are we ok making that usecase more difficult?)

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Have a regular Kubewarden toplevel version
+
+Having a regular --even semver based-- version on the toplevel project
+seems like an artificial construct. What is the meaning of patchlevel
+release in this case? What is Kubewarden 1.2.3 vs 1.2.4?
+
+Thus, ignoring patchlevel (e.g. just keeping 1.2 vs 1.3 vs 1.4, or
+even 1 vs 2 vs 3) is an option. But then, if it's just meaningless
+numbers, why not make the toplevel number meaningful, referring to the
```
viccuad, line 167: to me with a user hat, having major, minor, patch would inform me that I can upgrade without headache and needing to reconfigure things.
No matter the changes in the policy-server image, if I can just bump it and rollout the deployment without backwards incompatible changes, that's a bump in minor or patch version. Same for other components. Now, if I want to benefit from new features (bump in minor), then, I may need to enable them by hand, or maybe the come enabled by default.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+    YEAR.MONTH. If there is the need to make a second global release
+    during the same month, then a day is appended: YEAR.MONTH.DAY1. If
+    a third release is needed during that month: YEAR.MONTH.DAY2.
+
+    This makes it easier for maintainability of upgrades and is easier
+    to describe what should the upgrade path be if there is the
+    necessity of forcefully sequence upgrades in a certain order.
+
+- Should complete versions of components (MAJOR.MINOR.PATCH) be
+  referenced by the versions map as the example shows?
```
viccuad, line 195: I vote for complete versions of components, i.e., if we release a new patch for controller, we bump the general version of Kubewarden, and bump the day version.

viccuad (commented):
Thanks for the RFC!

I'm still not fully ratifying, undecided on not using semver, and where to record the Kubewarden version (chart, Configmap). See comments.

rfc/0006-kubewarden-versioning.md
jvanz: I like the psychological effect of Calver. It's cool to see a 2022.03 release and easy to understand. But when I think in how the Kubewarden is developed, I see us like a feature oriented project. Mostly of the time, we are not attached to date of things. So, this leds us to question like: `Should day versions be included in the toplevel map`? (from the `Unresolved question` section). Where we need to find a solution for when we have multiple global releases in the same time period.

It's not so uncommon we releasing features which require a "global release" (let's assume global release as a release where multiple components should be updated together) is a short period of time (less then a month). Therefore, I'm wondering if a simple version like v1, v2, v3 or even v1.0, v2.0 and so on... does not make more sense for our project. Which is also easy to understand and have a similar psychological effect of: "Wow! This project is advancing fast. We have 5 global releases within a month".

rfc/0006-kubewarden-versioning.md
ereslibre: I think the umbrella helm chart (versioned with calver) can install a configmap with the full map of versions of all components, so if needed, some components can always assume that this configmap exists in order to check their own compatibility with the rest of the components if needed.

This configmap would be autogenerated.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,203 @@
...
+There is also the `waPC` contract between the `policy-evaluator`
+(consumed by the Policy Server and kwctl) and Policies and SDK's.
+
+## Kubewarden toplevel version
+
+This proposal envisions using [`CalVer`](https://calver.org/) to
+version the Kubewarden project.
+
+Had Kubewarden be released today, it would be `Kubewarden
+22.03`. This makes it trivial to understand how old your version of
```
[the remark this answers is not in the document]
ereslibre: > For the first, I would prefer vulnerability scans and bills of materials.

I don't think they are mutually excluyent. I agree, but I think the information calver gives you is always valuable, besides security scans and BOM's.

>  For the later, I am undecided if I prefer semver. If I were a customer, I would prefer to know that I can update without non-backwards compat changes between releases, for example only updating minor versions. And schedule major versions for a later moment, as by the nature of Kubewarden, it may very well mean pausing the cluster workload.

Given there are a number of really big dependencies on our product, I don't think semver is really meaningful. I think WRT Kubewarden users what matters is "what is my upgrade path?". And to document that upgrade path (and their pre-requirements) I don't think semver offers any advantage over calver (I'm always referring to the umbrella version).

rfc/0006-kubewarden-versioning.md
ereslibre: > No matter the changes in the policy-server image, if I can just bump it and rollout the deployment without backwards incompatible changes, that's a bump in minor or patch version

Yeah, that's why I started this RFC with the "public interface", that by the way, is not complete.

Our interface is also comprised by Rego -- their builtins --, the Kubernetes version we support, the client-go we use in the controller, what policy-evaluator might read from the cluster (again, kube-rs...).

I think our dependencies and public interfaces are so many, and with many different natures impacting different kind of users (policy developers, cluster administrators...), that I don't believe semver for the toplevel versioning is meaningful.

rfc/0006-kubewarden-versioning.md
ereslibre: Despite I believe calver is the optimal way to go for the umbrella version I think it has a problem: Kubewarden will stabilize, and the core components will not be updated so often (at some point, this will happen). When that happens, how do you feel about installing now a 21.03 version, today?

Is it stable? Or is it abandoned? This is the only thing that does not convince me about calver to be honest. I still think it's something we can overcome, the project can be active and posting blog posts, new policies, news...

viccuad: > Is it stable? Or is it abandoned? 
In that time , would we need to rebuild images because of CVEs, and then release a new calver? If yes, then, the calver version means "it is secure until this date, and potentially more if there hasn't been a new release version afterwards".

ereslibre: Images are semver based.

Calver is the umbrella helm chart version. If we release today, 22.03.

If we fix a CVE on M07onday, we release 22.03.07.

rfc/0006-kubewarden-versioning.md
viccuad: My main concern here is users not knowing if the upgrade of the helm charts are painless or not.
I realize now that the helm charts have their own semver, and that carries that info.

I'm happy with calver in that aspect, resolving this conversation!

rfc/0006-kubewarden-versioning.md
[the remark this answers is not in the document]
viccuad: My main concern here is users not knowing if the upgrade of the helm charts are painless or not.
I realize now that the helm charts have their own semver, and that carries that info.

I'm happy with calver in that aspect, resolving this conversation!

rfc/0006-kubewarden-versioning.md
viccuad: > I think the umbrella helm chart (versioned with calver) can install a configmap

I don't think having an umbrella helm chart is feasible (as subcharts all get flattened and installed at the same time, giving us the problem of the controller not being up when instantiating PolicyServer). Unless it is a hypper chart.

Then, the problem is how to install it. I suppose the controller would do it.

viccuad (approved):
Ratifying on my side. I think calver is good, provided that helm charts have semver (as its the most painful part of the stack to upgrade).

jvanz:
Just remember to push this RFC to the [rfc repository](https://github.com/kubewarden/rfc) as well.

flavio (commented):
I'm not a fan of CalVer, hence I'm biased. I also think that having to maintain a global source of truth would be a bit painful and, most important of all, would require extra hops while debugging user reports.

I would prefer to do something like that:

  * Use semver
  * Have a unique version number for all the public components of the kubewarden stack you mentioned inside of the RFC

For example, when we release kubewarden `1.23` we would tag a `1.23.0` release for `policy-server`, `kwct`, `kubewarden-controller`.
Each project could have some level of independence with patch releases. That means we could fix a regression inside of `kwctl` and just release `kwctl 1.23.1` without having to tag all the other components.

As for the helm charts, they already have the concept of application version. This would reflect the global version of the kubewarden stack (e.g. `1.23.0`). The helm chart version can be something completely different and independent, or we could decide to follow the same process of the whole kubewarden stack.

## Pros 

### Clear support statement rules

This approach would allow us to set clear support rules like kubernetes does. For example, we could state that kubewarden controller version `x.y.z` can support policy-server versions between `x.(y-2).0` and `x.(y).0`.

### Consistent with Kubernetes

This versioning process is the same used by Kubernetes. Given we are a Kubernetes project, our users are already familiar with how Kubernetes versioning works. This should make them feel at home.

PR #195, opened 2022-03-25
fix: Disable cert-manager ca injection patches in CRDs

viccuad:
## Description

- fix: Disable cert-manager ca injection patches in CRDs
  These are not needed; we only need them on the webhooks.
Leaving the patches files themselves, as they are automatically
generated.

- build: Add `make generate-crds` for final CRDs that go to Helm charts
 This allows further automation when releasing Helm charts.

## Test

Generated crds by manually running `make generate-crds`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ testbin/*
 *.swp
 *.swo
 *~
+
+# Manually generated crds. Directory is needed though
+generated-crds/*
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -17,9 +17,9 @@ patchesStrategicMerge:
 
 # [CERTMANAGER] To enable webhook, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-- patches/cainjection_in_clusteradmissionpolicies.yaml
-- patches/cainjection_in_policyservers.yaml
-- patches/cainjection_in_admissionpolicies.yaml
+# - patches/cainjection_in_clusteradmissionpolicies.yaml
+# - patches/cainjection_in_policyservers.yaml
+# - patches/cainjection_in_admissionpolicies.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
 # the following config is for teaching kustomize how to do kustomization for CRDs.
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -23,3 +23,6 @@ testbin/*
 *.swp
 *.swo
 *~
+
+# Manually generated crds. Directory is needed though
+generated-crds/*
```
viccuad, line 28: The folder is needed so kustomize saves them in several files:
```
kubewarden-controller (no-cainjection-crds *$%)$ tree generated-crds
generated-crds
├── apiextensions.k8s.io_v1_customresourcedefinition_admissionpolicies.policies.kubewarden.io.yaml
├── apiextensions.k8s.io_v1_customresourcedefinition_clusteradmissionpolicies.policies.kubewarden.io.yaml
└── apiextensions.k8s.io_v1_customresourcedefinition_policyservers.policies.kubewarden.io.yaml
```

I chose not to check-in the files, but we could them check-in too.

config/crd/kustomization.yaml
```diff
@@ -17,9 +17,9 @@ patchesStrategicMerge:
 
 # [CERTMANAGER] To enable webhook, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-- patches/cainjection_in_clusteradmissionpolicies.yaml
-- patches/cainjection_in_policyservers.yaml
-- patches/cainjection_in_admissionpolicies.yaml
+# - patches/cainjection_in_clusteradmissionpolicies.yaml
+# - patches/cainjection_in_policyservers.yaml
+# - patches/cainjection_in_admissionpolicies.yaml
```
ereslibre, line 22: I'm sorry because I know I commented this. I was wondering why our CRD's had this annotation, thinking it was some mistake on our side -- copy/paste etc.

However, I have created a new kubebuilder project and I can see the patch targeting the `CustomResourceDefinition`.

I still don't understand the need for that, have to check in order to give a review.

ereslibre: Given that they are required for cert-manager, I think we should leave them uncommented (sorry!).

They come with the direct templating from kubebuilder, so I would not touch them. It's required for the webhooks (the kubewarden ones -- clusteradmissionpolicy, admissionpolicy, policy-server) to be served over TLS with cert-manager.

ereslibre (approved):
Thanks @viccuad!

.gitignore
ereslibre: I think it's fine to not check them in.

PR #199, opened 2022-03-31
Pod permissions.

jvanz:
## Description

Adds permissions allowing the controller to get,list and watch pods in the cluster. This is necessary because of the latest changes in the controller where it watches pods.


ereslibre (approved):
LGTM. Thanks @jvanz!

We have to see how we can reduce the scope of the get/list/watch to the deployments namespace, so that we only get pod events we care about, but I think as RBAC permissions this is already good.

PR #201, opened 2022-04-04
Sets policy as isPolicyUniqueReachable when the latest policy pod is ready.

jvanz:
## Description

Updates the code used to check if a policy is unique reachable only when the pods from the last policy server deployment replicaset are ready.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -41,6 +41,15 @@ rules:
   - update
   - watch
 - apiGroups:
+  - apps
+  resources:
+  - deployments
+  - replicasets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ""
   resources:
   - configmaps
--- a/controllers/policy_utils.go
+++ b/controllers/policy_utils.go
@@ -130,6 +130,7 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
 			return ctrl.Result{}, errors.Wrap(err, "error reconciling validating webhook")
 		}
 	}
+	policy.SetStatus(v1alpha2.PolicyStatusActive)
 
 	apimeta.SetStatusCondition(
 		&policy.GetStatus().Conditions,
@@ -140,7 +141,6 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
 			Message: "The latest replica set is uniquely reachable",
 		},
 	)
-	policy.SetStatus(v1alpha2.PolicyStatusActive)
 
 	return ctrl.Result{}, nil
 }
--- a/controllers/policystatus_utils.go
+++ b/controllers/policystatus_utils.go
@@ -80,15 +80,20 @@ func SetPolicyConfigurationCondition(policyServerConfigMap *corev1.ConfigMap, po
 	}
 }
 
+func isLatestReplicaSetFromPolicyServerDeployment(replicaSet *appsv1.ReplicaSet, policyServerDeployment *appsv1.Deployment) bool {
+	return replicaSet.Annotations[constants.KubernetesRevisionAnnotation] == policyServerDeployment.Annotations[constants.KubernetesRevisionAnnotation] &&
+		replicaSet.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] == policyServerDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation]
+}
+
 func isPolicyUniquelyReachable(ctx context.Context, apiReader client.Reader, policyServerDeployment *appsv1.Deployment) bool {
 	replicaSets := appsv1.ReplicaSetList{}
-	if err := apiReader.List(ctx, &replicaSets, client.InNamespace(policyServerDeployment.Namespace)); err != nil {
+	if err := apiReader.List(ctx, &replicaSets, client.MatchingLabels{constants.PolicyServerLabelKey: policyServerDeployment.Labels[constants.PolicyServerLabelKey]}); err != nil {
 		return false
 	}
 	podTemplateHash := ""
-	for _, replicaSet := range replicaSets.Items {
-		if replicaSet.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] == policyServerDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] {
-			podTemplateHash = replicaSet.Labels[appsv1.DefaultDeploymentUniqueLabelKey]
+	for index := range replicaSets.Items {
+		if isLatestReplicaSetFromPolicyServerDeployment(&replicaSets.Items[index], policyServerDeployment) {
+			podTemplateHash = replicaSets.Items[index].Labels[appsv1.DefaultDeploymentUniqueLabelKey]
 			break
 		}
 	}
@@ -96,7 +101,7 @@ func isPolicyUniquelyReachable(ctx context.Context, apiReader client.Reader, pol
 		return false
 	}
 	pods := corev1.PodList{}
-	if err := apiReader.List(ctx, &pods, client.InNamespace(policyServerDeployment.Namespace)); err != nil {
+	if err := apiReader.List(ctx, &pods, client.MatchingLabels{constants.PolicyServerLabelKey: policyServerDeployment.Labels[constants.PolicyServerLabelKey]}); err != nil {
 		return false
 	}
 	if len(pods.Items) == 0 {
@@ -110,7 +115,6 @@ func isPolicyUniquelyReachable(ctx context.Context, apiReader client.Reader, pol
 			return false
 		}
 	}
-
 	return true
 }
 
--- a/internal/pkg/constants/constants.go
+++ b/internal/pkg/constants/constants.go
@@ -44,4 +44,7 @@ const (
 
 	// Finalizers
 	KubewardenFinalizer = "kubewarden"
+
+	// Kubernetes
+	KubernetesRevisionAnnotation = "deployment.kubernetes.io/revision"
 )
--- a/controllers/admissionpolicy_controller.go
+++ b/controllers/admissionpolicy_controller.go
@@ -40,6 +40,7 @@ import (
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=admissionpolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=admissionpolicies/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=apps,resources=replicasets;deployments,verbs=get;list;watch
 
 // AdmissionPolicyReconciler reconciles an AdmissionPolicy object
 type AdmissionPolicyReconciler struct {
--- a/controllers/clusteradmissionpolicy_controller.go
+++ b/controllers/clusteradmissionpolicy_controller.go
@@ -40,6 +40,7 @@ import (
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=clusteradmissionpolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=policies.kubewarden.io,resources=clusteradmissionpolicies/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=apps,resources=replicasets;deployments,verbs=get;list;watch
 
 // ClusterAdmissionPolicyReconciler reconciles a ClusterAdmissionPolicy object
 type ClusterAdmissionPolicyReconciler struct {


controllers/policy_utils.go
```diff
@@ -105,6 +105,16 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
...
+		// we need to check if the policy has status active because
+		// when installing a new policy, the first reconciliation loop may
+		// occur before the new replicaset creation. Which means that
+		// the comparisons in the isPolicyUniquelyReachable returns the policy
+		// as unique reachable. Which is true in that point in time.
+		// Therefore, the policy is set as "active". In the second
+		// reconciliation, we need to fix that.
+		if policy.GetStatus().PolicyStatus == v1alpha2.PolicyStatusActive {
+			policy.SetStatus(v1alpha2.PolicyStatusPending)
+		}
```
ereslibre, line 117: I think it would be better to avoid setting its status to active when no replica set exists yet, so that we don't have to maintain side fixes. For example, this change would imply that a user watching the policy will see it going from `active` to `pending`, being `active` for a very short period of time. I think we should make this impossible in the first place.

controllers/policystatus_utils.go
```diff
@@ -80,14 +80,23 @@ func SetPolicyConfigurationCondition(policyServerConfigMap *corev1.ConfigMap, po
...
 }
 
+// Checks if the given replica is owned by the given policy server deployment
+// and if it is the latest replica set from the deployment as well.
+func isReplicaSetFromPolicyServerDeployment(replicaSet *appsv1.ReplicaSet, policyServerDeployment *appsv1.Deployment) bool {
+	return replicaSet.Labels[constants.PolicyServerLabelKey] == policyServerDeployment.Labels[constants.PolicyServerLabelKey] &&
+		replicaSet.Annotations[constants.KubernetesRevisionAnnotation] == policyServerDeployment.Annotations[constants.KubernetesRevisionAnnotation] &&
+		replicaSet.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] == policyServerDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation]
+}
+
```
ereslibre, line 90: For this check I think we can just filter by the policy server label, given we are checking "ownership". So, something like:

```
func isReplicaSetFromPolicyServerDeployment(replicaSet *appsv1.ReplicaSet, policyServerDeployment *appsv1.Deployment) bool {
	return replicaSet.Labels[constants.PolicyServerLabelKey] == policyServerDeployment.Labels[constants.PolicyServerLabelKey]
}
```

Should be enough.

controllers/policystatus_utils.go
```diff
@@ -106,12 +115,11 @@ func isPolicyUniquelyReachable(ctx context.Context, apiReader client.Reader, pol
 		if pod.DeletionTimestamp != nil {
 			continue
 		}
-		if pod.Labels[appsv1.DefaultDeploymentUniqueLabelKey] != podTemplateHash || !isPodReady(pod) {
-			return false
+		if pod.Labels[appsv1.DefaultDeploymentUniqueLabelKey] == podTemplateHash && isPodReady(pod) {
+			return true
```
ereslibre, line 119: This is inverting the logic of the check. The idea of the `isUniquelyReachable` condition is that there is no possibility for a request to arrive to an old pod (e.g. with old policy settings, or old policy version), because no old pod exists.

So, the fact that we find a pod coming from the newest replicaset and that it's ready does not mean that other older pod from an older replicaset is not still around able to process requests.

I would change this code to follow this behavior. Furthermore, I think we can also change the policy status before:

- Move lines https://github.com/kubewarden/kubewarden-controller/blob/410204ee09fdcfa50f78296d79c8c887e7b3e508/controllers/policy_utils.go#L107-L118 to https://github.com/kubewarden/kubewarden-controller/blob/410204ee09fdcfa50f78296d79c8c887e7b3e508/controllers/policy_utils.go#L133.
-  Move https://github.com/kubewarden/kubewarden-controller/blob/410204ee09fdcfa50f78296d79c8c887e7b3e508/controllers/policy_utils.go#L143 right before the `isUniquelyReachable` block just moved in the previous point.

This would ensure that the policy is set to active soon, but that we only change the `isUniquelyRechable` condition at the very end, when we know that no old pods will receive any requests. Users are free to wait for the isUniquelyReachable condition or not. When waiting for it we know that no request can reach an older pod, but we also know that this condition takes a little more to transition to `True`.

So, what I propose is something like:

```go
func reconcilePolicy(ctx context.Context, client client.Client, reconciler admission.Reconciler, policy v1alpha2.Policy) (ctrl.Result, error) {
	if policy.GetPolicyServer() == "" {
		policy.SetStatus(v1alpha2.PolicyStatusUnscheduled)
		return ctrl.Result{}, nil
	}
	policyServer, err := getPolicyServer(ctx, client, policy)
	if err != nil {
		policy.SetStatus(v1alpha2.PolicyStatusScheduled)
		//lint:ignore nilerr set status to scheduled if policyServer can't be retrieved, and stop reconciling
		return ctrl.Result{}, nil
	}
	if policy.GetStatus().PolicyStatus != v1alpha2.PolicyStatusActive {
		policy.SetStatus(v1alpha2.PolicyStatusPending)
	}

	policyServerDeployment := appsv1.Deployment{}
	if err := reconciler.APIReader.Get(ctx, types.NamespacedName{Namespace: reconciler.DeploymentsNamespace, Name: naming.PolicyServerDeploymentNameForPolicyServerName(policy.GetPolicyServer())}, &policyServerDeployment); err != nil {
		if apierrors.IsNotFound(err) {
			return ctrl.Result{Requeue: true}, nil
		}
		return ctrl.Result{}, errors.Wrap(err, "could not read policy server Deployment")
	}

	secret := corev1.Secret{}
	if err := client.Get(ctx, types.NamespacedName{Namespace: reconciler.DeploymentsNamespace, Name: constants.PolicyServerCARootSecretName}, &secret); err != nil {
		return ctrl.Result{}, errors.Wrap(err, "cannot find policy server secret")
	}
	if policy.IsMutating() {
		if err := reconciler.ReconcileMutatingWebhookConfiguration(ctx, policy, &secret, policyServer.NameWithPrefix()); err != nil {
			return ctrl.Result{}, errors.Wrap(err, "error reconciling mutating webhook")
		}
	} else {
		if err := reconciler.ReconcileValidatingWebhookConfiguration(ctx, policy, &secret, policyServer.NameWithPrefix()); err != nil {
			return ctrl.Result{}, errors.Wrap(err, "error reconciling validating webhook")
		}
	}

	policy.SetStatus(v1alpha2.PolicyStatusActive)

	if !isPolicyUniquelyReachable(ctx, client, &policyServerDeployment) {
		apimeta.SetStatusCondition(
			&policy.GetStatus().Conditions,
			metav1.Condition{
				Type:    string(v1alpha2.PolicyUniquelyReachable),
				Status:  metav1.ConditionFalse,
				Reason:  "LatestReplicaSetIsNotUniquelyReachable",
				Message: "The latest replica set is not uniquely reachable",
			},
		)
		return ctrl.Result{Requeue: true, RequeueAfter: 2 * time.Second}, nil
	}

	apimeta.SetStatusCondition(
		&policy.GetStatus().Conditions,
		metav1.Condition{
			Type:    string(v1alpha2.PolicyUniquelyReachable),
			Status:  metav1.ConditionTrue,
			Reason:  "LatestReplicaSetIsUniquelyReachable",
			Message: "The latest replica set is uniquely reachable",
		},
	)

	return ctrl.Result{}, nil
}
```

ereslibre (changes requested):
Thanks @jvanz! Some comments.

controllers/policy_utils.go
raulcabello: yes, I agree. I would not set the status to active if the replica set is not created yet. Looking into the code it should return false if the replica set is not there https://github.com/jvanz/kubewarden-controller/blob/46827eaf2fa1664721d7f4a94b783966dbdcf24d/controllers/policystatus_utils.go#L93-L94 we should investigate why this is not working

jvanz: I agree that the `active` -> `pending` -> `active` is not nice. But the problem is that there is a `replicaset` when we install a policy. For example, the user installs the policy server with no policy yet. There is a replica set to the current policy server. Then, the user installs the policy. This triggers a reconciliation, and the policy status `reconciler` starts. When the reconcile loop checks if the `replicaset` is the latest it will returns true. Because the new `replicaset` used to deploy the policy is not there yet. Thus, the `replicaset` version will match with the version in the `deployment`. Returning the policy as unique reachable setting it as `active`. 

I could not find a way to detect that the current `replicaset`, even matching the `deployment` version, is not the latest `replicaset`.

ereslibre: > I could not find a way to detect that the current replicaset, even matching the deployment version, is not the latest replicaset.

A possible way would be to ignore replicasets with a `.metadata.creationTimestamp` lower than the policy`.metadata.creationTimestamp`. This way, we know that it's not a replica set we should compare against.

controllers/policystatus_utils.go
jvanz: But we can have more than one replica set for the same deployment. Kubernetes keep a history of the replicaset for some iterations. Do the list method remove those ones?  

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#revision-history-limit

jvanz: Never mind, the function does not need to find the latest `replicaset` in the end.

apis/v1alpha2/policy.go
```diff
@@ -39,9 +39,6 @@ const (
 type PolicyConditionType string
 
 const (
-	// PolicyActive represents the condition of the Policy admission
-	// webhook been registered
-	PolicyActive PolicyConditionType = "PolicyActive"
```
ereslibre, line 44: Please keep this one around. I forgot to add the condition set to `True` on the refactor. This is the condition our users can wait for when they create a new policy, this condition will be true when we register the webhooks.

I think we can do that on a follow up PR, but I would not remove this condition.

controllers/policy_utils.go
jvanz: Never mind, I'm mixing the meanings of the status `active` and the condition `isPolicyUniquelyReachable`.

apis/v1alpha2/policy.go
jvanz: Okay

controllers/policystatus_utils.go
jvanz: This change break the tests. But I'll find a solution for that and push the code for review again.

jvanz: I've pushed a new version of the `isPolicyUniquelyReachable` function. Now, I do not invert the logic of it. But I filter the replicaset used in the comparison. AFAICS, it fixed the original problem and it does not require changes in the tests. 

Let me know what you think

controllers/policy_utils.go
```diff
@@ -140,8 +140,8 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
 			Message: "The latest replica set is uniquely reachable",
 		},
 	)
-	policy.SetStatus(v1alpha2.PolicyStatusActive)
 
+	policy.SetStatus(v1alpha2.PolicyStatusActive)
```
ereslibre, line 144: I think it would be better to put the policy in active state right after the webhooks have been reconciled

controllers/policystatus_utils.go
```diff
@@ -96,7 +102,7 @@ func isPolicyUniquelyReachable(ctx context.Context, apiReader client.Reader, pol
 		return false
 	}
 	pods := corev1.PodList{}
-	if err := apiReader.List(ctx, &pods, client.InNamespace(policyServerDeployment.Namespace)); err != nil {
+	if err := apiReader.List(ctx, &pods, client.MatchingLabels{constants.PolicyServerLabelKey: policyServerDeployment.Labels[constants.PolicyServerLabelKey]}); err != nil {
```
ereslibre, line 105: 💯

internal/pkg/constants/constants.go
```diff
@@ -44,4 +44,7 @@ const (
 
 	// Finalizers
 	KubewardenFinalizer = "kubewarden"
+
+	// Kubernetes
+	KubernetesRevisionAnnotation = "deployment.kubernetes.io/revision"
```
ereslibre, line 49: You can use this one: https://github.com/kubernetes/kubernetes/blob/2da1558909043ddc41d7a36ab93672869474ec55/pkg/controller/deployment/util/deployment_util.go#L48

ereslibre (changes requested):
Thanks @jvanz!

controllers/policystatus_utils.go
```diff
@@ -80,23 +80,28 @@ func SetPolicyConfigurationCondition(policyServerConfigMap *corev1.ConfigMap, po
...
 	replicaSets := appsv1.ReplicaSetList{}
-	if err := apiReader.List(ctx, &replicaSets, client.InNamespace(policyServerDeployment.Namespace)); err != nil {
+	if err := apiReader.List(ctx, &replicaSets, client.MatchingLabels{constants.PolicyServerLabelKey: policyServerDeployment.Labels[constants.PolicyServerLabelKey]}); err != nil {
 		return false
 	}
 	podTemplateHash := ""
-	for _, replicaSet := range replicaSets.Items {
-		if replicaSet.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] == policyServerDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] {
-			podTemplateHash = replicaSet.Labels[appsv1.DefaultDeploymentUniqueLabelKey]
+	for index := range replicaSets.Items {
```
ereslibre, line 94: Why not loop with two vars as before, ignoring the index?

config/rbac/role.yaml
```diff
@@ -40,6 +40,15 @@ rules:
...
   - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  - replicasets
+  verbs:
+  - get
+  - list
+  - watch
```
raulcabello, line 51: please don't forget to add this to the helm chart as well

raulcabello (commented):
I agree with @ereslibre comments. Apart from that it looks good to me. Thanks!

viccuad (approved):
LGTM, besides Rafa's comments too

controllers/policystatus_utils.go
jvanz: Because our linter complains about that: ` Error: G601: Implicit memory aliasing in for loop. (gosec)`

https://github.com/kubewarden/kubewarden-controller/runs/5838848906?check_suite_focus=true

config/rbac/role.yaml
jvanz: :+1:

controllers/policystatus_utils.go
ereslibre: Phew... good. Thanks :)

internal/pkg/constants/constants.go
ereslibre: You are right. I checked this by grepping on `kubernetes/staging`, that includes all repos that end up in separate repos in the Github `kubernetes/xyz` and there is not such `KubernetesRevisionAnnotation` constant we can use, only from the main tree, that should not be consumed directly.

Sorry for the bad pointer.

ereslibre (approved):
LGTM, thanks a lot @jvanz for finding the issue, fixing it and having a lot of patience with the reviews!

PR #202, opened 2022-04-06
Add PolicyActive condition.

jvanz:
Updates the policy reconciliation loop to set the PolicyActive condition to true when the policy status is updated to "active".

## Test
You can test this manually or using the e2e test repository. 

If you want to test it using e2e tests. You need to follow some steps:

```shell
# upload the container image somewhere
# update the makefile setting the container name and tag to be used
 make create-k8s-cluster install-kubewarden basic-e2e-test
```

### DIFF
--- a/controllers/policy_utils.go
+++ b/controllers/policy_utils.go
@@ -82,6 +82,15 @@ func startReconciling(ctx context.Context, client client.Client, reconciler admi
 }
 
 func reconcilePolicy(ctx context.Context, client client.Client, reconciler admission.Reconciler, policy v1alpha2.Policy) (ctrl.Result, error) {
+	apimeta.SetStatusCondition(
+		&policy.GetStatus().Conditions,
+		metav1.Condition{
+			Type:    string(v1alpha2.PolicyActive),
+			Status:  metav1.ConditionFalse,
+			Reason:  "PolicyActive",
+			Message: "The policy webhook has not been created",
+		},
+	)
 	if policy.GetPolicyServer() == "" {
 		policy.SetStatus(v1alpha2.PolicyStatusUnscheduled)
 		return ctrl.Result{}, nil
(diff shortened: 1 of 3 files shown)


controllers/policy_utils.go
```diff
@@ -144,6 +144,18 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
...
+	policy.SetStatus(v1alpha2.PolicyStatusActive)
+	apimeta.SetStatusCondition(
+		&policy.GetStatus().Conditions,
+		metav1.Condition{
+			Type:    string(v1alpha2.PolicyActive),
+			Status:  metav1.ConditionTrue,
+			Reason:  "PolicyActive",
+			Message: "The policy webhooks are created",
+		},
+	)
```
ereslibre, line 157: This is good. I would also set this same condition to `ConditionFalse` at the beginning of the reconciliation loop.

This means that if we early exit because of an error, the condition on the policy is updated always, and it's set to `False`.

ereslibre:
I would also take advantage of this change to move this block:

https://github.com/kubewarden/kubewarden-controller/blob/8c8434c2fa25a88edf559de9c2bee2c4465770a5/controllers/policy_utils.go#L120-L133

before this block:

https://github.com/kubewarden/kubewarden-controller/blob/8c8434c2fa25a88edf559de9c2bee2c4465770a5/controllers/policy_utils.go#L107-L118

so that we don't wait for "uniquely reachable" in order to install the webhooks. The main reason is that we don't perform any kind of follow up later regarding the status of the pods to remove the webhook (which is fine, and I don't think we should do it). So I think we should not check this before we install the webhook either.

As a user, I can do:

- Install policy
- Wait for policy condition `PolicyActive` if I want -- this means that webhooks are reconciled, my policy is "active" in that the apiserver is starting to send requests to the webhook

When I change a policy setting, whatever that is:

- Change policy
- Wait for policy condition `IsUniquelyReachable` if I want

controllers/policy_utils.go
```diff
@@ -82,6 +82,15 @@ func startReconciling(ctx context.Context, client client.Client, reconciler admi
 }
 
 func reconcilePolicy(ctx context.Context, client client.Client, reconciler admission.Reconciler, policy v1alpha2.Policy) (ctrl.Result, error) {
+	apimeta.SetStatusCondition(
+		&policy.GetStatus().Conditions,
+		metav1.Condition{
+			Type:    string(v1alpha2.PolicyActive),
+			Status:  metav1.ConditionFalse,
+			Reason:  "PolicyActive",
+			Message: "The policy webhooks are created",
```
ereslibre, line 91:
```suggestion
			Message: "The policy webhook has not been created",
```

controllers/policy_utils.go
```diff
@@ -144,6 +153,18 @@ func reconcilePolicy(ctx context.Context, client client.Client, reconciler admis
...
 }
+func setPolicyAsActive(policy v1alpha2.Policy) {
+	policy.SetStatus(v1alpha2.PolicyStatusActive)
+	apimeta.SetStatusCondition(
+		&policy.GetStatus().Conditions,
+		metav1.Condition{
+			Type:    string(v1alpha2.PolicyActive),
+			Status:  metav1.ConditionTrue,
+			Reason:  "PolicyActive",
+			Message: "The policy webhooks are created",
```
ereslibre, line 164:
```suggestion
			Message: "The policy webhook has been created",
```

ereslibre (approved):
LGTM, a minor wording suggestion. Thanks @jvanz

controllers/policy_utils.go
ereslibre: The message on this one has to be in negative form given the condition is `False`.

ereslibre (approved):

PR #206, opened 2022-04-17
Extract common policy properties to a separate type

alexander-demichev:
## Description

Extract common policy properties to a separate type, this will make maintenance of similar fields easier.

## Test


internal/pkg/admission/testing_common.go
```diff
@@ -3,29 +3,39 @@ package admission
 import (
 	"time"
 
-	v1alpha2 "github.com/kubewarden/kubewarden-controller/apis/v1alpha2"
+	kubewardenv1 "github.com/kubewarden/kubewarden-controller/apis/v1alpha2"
```
ereslibre, line 6:
```suggestion
	kubewardenv1alpha2 "github.com/kubewarden/kubewarden-controller/apis/v1alpha2"
```

Just to be explicit about the import. It applies to other imports named as `kubewardenv1` when the package imported was `v1alpha2`. It's a minor detail and probably a matter of taste.

ereslibre (changes requested):
Thanks @alexander-demichev! This is a really nice contribution!

Unit tests are not passing, you can check by running `make unit-tests` in the root of the project.

The only thing missing in terms of contents is to regenerate the CRD ref docs, you can do that by installing `crd-ref-docs` tool (https://github.com/elastic/crd-ref-docs) and running the `generate` target in the `Makefile` at `docs/crds`. You should see that `docs/crds/README.asciidoc` has changes afterwards. I can take care of adding the commit to your PR, as you prefer.

internal/pkg/admission/testing_common.go
alexander-demichev: `kubewardenv1` is a better choice in this case because when you promote the API version to v1alpha3 or beta, you'll only have to change import and avoid making changes in code where API objects are used. It's a common practice in kube-related projects, I can open a follow-up PR and change all imports to `kubewardenv1`.

ereslibre: In order to properly achieve support for multiple versions would be to either:

a) Refer to the same version of the API in the code. Have the spike model translate other versions to the central one. The code only uses one version (the one you decide as the "main" one).
b) Refer to an "internal type" not directly marshaled/unmarshaled. With this, the code always refers to this internal type, and there is a way to always convert different versions to this internal type.

Thinking about the future, I think that what you are aiming for is a). With multiple supported versions I think b) is slightly better because you never change your code (even if you decide that your "reference version" is something else), you always refer to the internal type in your code and never change it, you maintain the transformations from external types (v1alpha1, v1alpha2, v1beta1, v1...) to the internal type.

I think `kubewardenv1` might lead to think that it's the `v1` version, as in `corev1` or `appsv1`.

When the time comes for multiple versions I think we will have to make the call of choosing between something like a) or something like b) and then update.

ereslibre: @alexander-demichev Ok, let's do that for now and use `kubewardenv1` as you suggest. I must admit I'm not a big fan because if we push to a `v1alpha3` all imports still need to be changed. It's true that the rest of the source code would still use `v1`.

I would prefer to not use versioned types in internal code though, only at the edges.

ereslibre (approved):
LGTM

viccuad (approved):
LGTM!

PR #212, opened 2022-05-06
Add `always-accept-admission-reviews-on-deployments-namespace` flag

ereslibre:
Depends on: https://github.com/kubewarden/policy-server/pull/247

This allows the user to configure the controller in a way that when it
reconciles the policy-server deployment, the policy-server will be
configured in a way to always accept requests targeting the namespace
where policy-servers are running (the provided
`--deployments-namespace` argument in the controller)


jvanz (approved):

flavio (approved):
LGTM, I think we can go ahead and merge it even without having the new policy-server tagged. We are setting an environment variable, hence there's no harm

PR #216, opened 2022-05-12
Fix the policy server env var name.

jvanz:
Updates the env var name used to define the namespace which the policy server should always accept the request.

This PR is necessary after a change done in the review process of the https://github.com/kubewarden/policy-server/pull/247


flavio (approved):
LGTM

jvanz:
Related to https://github.com/kubewarden/kubewarden-controller/pull/212/files

PR #219, opened 2022-05-13
Skip Kubewarden namespace in the webhook configuration.

jvanz:
Updates the reconciler to add a namespace selector in the webhooks configuration to skip request from the Kubewarden's namespace. Therefore, avoiding issues during the policy server rollout due request being rejected by the old policy server instance.

### DIFF
--- a/internal/pkg/admission/mutating-webhook.go
+++ b/internal/pkg/admission/mutating-webhook.go
@@ -95,7 +95,7 @@ func (r *Reconciler) mutatingWebhookConfiguration(
 				Rules:                   policy.GetRules(),
 				FailurePolicy:           policy.GetFailurePolicy(),
 				MatchPolicy:             policy.GetMatchPolicy(),
-				NamespaceSelector:       policy.GetNamespaceSelector(),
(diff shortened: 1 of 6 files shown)


jvanz:
The previous change adding an environment variable with the namespace to be skipped does not work. During the rollout request will be send to the old policy server instance asking to validate a policy, the policy which the rollout suppose to deploy, which is not running yet.

jvanz:
Important note: this fix does not suppose to be a final solution. It's just a short-term fix to allow users to use Kubewarden without big problems. We still need to come up with a better solution

raulcabello (commented):
This approach is fine for `ClusterAdmissionPolicies`. However I think for `AdmissionPolicies` we should just reject the policy if it is created in the same ns. If we just ignore the namespace where the AdmissionPolicy is created, this policy will never be called. In this case I'd prefer to not allow the creation.
We can check the ns in the existing validating webhook https://github.com/kubewarden/kubewarden-controller/blob/main/apis/v1alpha2/admissionpolicy_webhook.go#L66

raulcabello (approved):
It looks like the other approach it’s not that easy. We can improve it later. Thanks!

PR #231, opened 2022-05-18
feat: sign container image

### LINKED ISSUE #230: Feature Request: Sign kubewarden-controller image with Sigstore

Artifacts are not signed, therefore we can't verify if they come from a trusted source

### Solution you'd like

Sign kubewarden-controller image using keyless in github actions

raulcabello:
Use cosign keyless to sign the kubewarden-controller image. 

I tested in my fork. You can see the [github actions](https://github.com/raulcabello/kubewarden-controller/actions) and you can verify the signature with:
```
COSIGN_EXPERIMENTAL=1 cosign verify ghcr.io/raulcabello/kubewarden-controller:latest | jq
```

Fix #230

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -33,6 +33,6 @@ jobs:
       - name: Sign the images
         run: |
           cosign sign \
-            ${{needs.build.outputs.repository}}:${{needs.build.outputs.tag}}
+            ${{needs.build.outputs.repository}}@${{needs.build.outputs.digest}}
         env:
           COSIGN_EXPERIMENTAL: 1
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -14,3 +14,25 @@ jobs:
     uses: kubewarden/kubewarden-controller/.github/workflows/container-image.yml@main
     with:
       push-image: true
+
+  sign:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      id-token: write
+    needs: build
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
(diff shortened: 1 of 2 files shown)


viccuad (approved):
LGTM! As with the policy-server, I would add a `cosign verify` after the sign, even if it only verifies that a signature was made, and not the signer.

jvanz (approved):

raulcabello:
draft until we use digest like we do in https://github.com/kubewarden/policy-server/pull/256

raulcabello:
This can now be reviewed. We are now signing the digest, you can verify it [here](https://github.com/raulcabello/kubewarden-controller/runs/6552207293?check_suite_focus=true)

@viccuad as I mentioned in the other PR I think there is no need to call verify. I think we can assume the image was successfully signed if `cosign sign` did not exit with an error.

flavio (approved):
awesome! :clap:

flavio:
> @viccuad as I mentioned in the other PR I think there is no need to call verify. I think we can assume the image was successfully signed if `cosign sign` did not exit with an error.

I agree, I don't think we need to call `cosign verify`

flavio:
@raulcabello I think this is having problems signings `latest`. The signature for latest is not found inside of your ghcr. I think something is wrong, see: https://github.com/raulcabello/kubewarden-controller/runs/6553336932?check_suite_focus=true

raulcabello:
@flavio it's not working because it is using the workflow from the [kubewarden ](https://github.com/raulcabello/kubewarden-controller/blob/main/.github/workflows/container-build.yml#L14) org, and the changes are not merged yet.

[here](https://github.com/raulcabello/kubewarden-controller/runs/6551847304?check_suite_focus=true) you can see how it signed latest using the workflow from my fork

flavio:
@raulcabello go ahead and merge it :+1:

PR #234, opened 2022-05-23
Check the policy server container image in all containers. 

### LINKED ISSUE #178: Improve the logic to trigger a rollout when the policy-server image changes

### Current Behavior

We should do a rollout of the `policy-server` Deployment if the image has changed. Currently we are checking the image for the first container [Containers[0]](https://github.com/kubewarden/kubewarden-controller/blob/30d36df36827f27a4641b37b6caa663c20bd6404/internal/pkg/admission/policy-server-deployment.go#L123)
There might be the case that our policy-server container is not always the first container. For example if some webhook adds a sidecar container. If our `policy-server` is not the first container, `policy-server` image updates won't work.

### Expected Behavior

Changes to `policy-server` image should always trigger a rollout regardless the order and number of containers in the pod

### Steps To Reproduce

1. Add an additional container in the first position to the `policy-server` deployment template spec  
2. Change  `policy-server` image
3. No rollout happens

### Environment

```markdown
All
```

### Anything else?

_No response_

jvanz:
## Description
If the container of the policy server is not in the first position of the deployment container list, the policy server rollout does not start. This commit changes this behaviour by looking for the policy server container in any position of the container list.

Fix #178

### DIFF
--- a/internal/pkg/admission/policy-server-deployment.go
+++ b/internal/pkg/admission/policy-server-deployment.go
@@ -103,7 +103,11 @@ func (r *Reconciler) updatePolicyServerDeployment(ctx context.Context, policySer
 		return fmt.Errorf("cannot retrieve existing policy-server Deployment: %w", err)
 	}
 
-	if shouldUpdatePolicyServerDeployment(originalDeployment, newDeployment) {
+	shouldUpdate, err := shouldUpdatePolicyServerDeployment(policyServer, originalDeployment, newDeployment)
+	if err != nil {
+		return fmt.Errorf("cannot check if it is necessary to update the policy server deployment: %w", err)
+	}
+	if shouldUpdate {
 		patch := originalDeployment.DeepCopy()
 		patch.Spec.Replicas = newDeployment.Spec.Replicas
 		patch.Spec.Template = newDeployment.Spec.Template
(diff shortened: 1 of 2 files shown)


internal/pkg/admission/policy-server-deployment.go
```diff
@@ -118,9 +118,31 @@ func (r *Reconciler) updatePolicyServerDeployment(ctx context.Context, policySer
...
+	for containerIndex := range deployment.Spec.Template.Spec.Containers {
+		container := &deployment.Spec.Template.Spec.Containers[containerIndex]
+		if container.Name == policyServer.NameWithPrefix() {
+			return container.Image, nil
+		}
+	}
+	return "", fmt.Errorf("cannot find policy server container")
+}
+
+func ifPolicyServerImageChanged(policyServer *v1alpha2.PolicyServer, originalDeployment *appsv1.Deployment, newDeployment *appsv1.Deployment) bool {
```
flavio, line 131: Nit, I would name this function `isPolicyServerImageChanged` not `if`

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -118,9 +118,31 @@ func (r *Reconciler) updatePolicyServerDeployment(ctx context.Context, policySer
...
+	var oldImage, newImage string
+	var err error
+	if oldImage, err = getPolicyServerImageFromDeployment(policyServer, originalDeployment); err != nil {
+		return false
+	}
+	if newImage, err = getPolicyServerImageFromDeployment(policyServer, newDeployment); err != nil {
+		return false
+	}
```
flavio, line 139: I would not hide these errors. Instead, I would:

  * Change the signature of the function to return `(string, error)`
  * Change these `return` to be `return "", false`

flavio (changes requested):
Good catch, I left some minor notes

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -125,3 +135,67 @@ func createDeployment(replicasInt int, serviceAccount, imagePullSecret, image st
...
+
+func insertContainer(deployment *appsv1.Deployment) {
+	container := corev1.Container{
+		Name:  "container0",
+		Image: "container0image:latest",
+	}
+	containers := []corev1.Container{container}
+	for index := range deployment.Spec.Template.Spec.Containers {
+		containers = append(containers, deployment.Spec.Template.Spec.Containers[index])
+	}
```
raulcabello, line 147:
```suggestion
		containers = append(containers, deployment.Spec.Template.Spec.Containers...)
```
nit: we don't need the fork to add the containers to the slice

raulcabello (commented):
I tested locally, and it is working fine! I agree with Flavio's comments, and I left one minor comment too. Thanks!

internal/pkg/admission/policy-server-deployment.go
jvanz: I've changed to code to propagate the error. Thus, I believe that  reconciler can handle it better. Please, let me know what you think.

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -118,13 +122,39 @@ func (r *Reconciler) updatePolicyServerDeployment(ctx context.Context, policySer
 	return nil
 }
 
-func shouldUpdatePolicyServerDeployment(originalDeployment *appsv1.Deployment, newDeployment *appsv1.Deployment) bool {
+func getPolicyServerImageFromDeployment(policyServer *v1alpha2.PolicyServer, deployment *appsv1.Deployment) (string, error) {
+	for containerIndex := range deployment.Spec.Template.Spec.Containers {
+		container := &deployment.Spec.Template.Spec.Containers[containerIndex]
+		if container.Name == policyServer.NameWithPrefix() {
+			return container.Image, nil
```
viccuad, line 129: doesn't this for just return the first policy-server image, as did the behaviour before this PR?

raulcabello: It's iterating through all the containers, so it will return the right image. Before we were assuming the policy server was always the first container [Containers[0]](https://github.com/kubewarden/kubewarden-controller/blob/30d36df36827f27a4641b37b6caa663c20bd6404/internal/pkg/admission/policy-server-deployment.go#L123)

viccuad: doh, true, we just need "a" policy-server container image. Thanks!

viccuad (approved):

flavio (approved):

Issue #242, opened 2022-05-31
Dont mark pod as uniquely reachable if there is a terminating pod without the latest configmap version


raulcabello:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We skip the terminating pods from the isUniquelyReachable condition. Therefore pods are mark as active and uniquely reachable when there still some terminating pods that might be still get called.

### Expected Behavior

A pod should be UniquelyReachable once all pods have the latest configmap version (including terminating pods)

PR #243, opened 2022-05-31
feat: don't mark policy as uniquely reachable until all terminating pods have disappeared.

raulcabello:
## Description

Don't skip pods with deletion timestamp in the isUniquelyReachable method

Fix https://github.com/kubewarden/kubewarden-controller/issues/242

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -14,7 +14,7 @@ jobs:
   run-e2e-tests:
     name: "Tests"
     needs: [build]
-    uses: kubewarden/kubewarden-end-to-end-tests/.github/workflows/e2e-tests.yml@v0.2.1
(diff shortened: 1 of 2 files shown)


ereslibre (approved):
I honestly don't recall why we were skipping deleted pods. However, I think this is safer when it comes to setting the uniquely reachable condition.

The reasoning in terms of maintainability is that we don't need to keep track of endpoints etc to determine if a policy is uniquely reachable when a rollout is happening, but just to check that there are no old pods around, just the latest one that matches the configuration version we are expecting, so that it's not possible for requests to reach other pods with an older configuration for a given policy server.

Thanks @raulcabello!

viccuad (approved):
LGTM, many thanks!

jvanz (approved):

viccuad:
Tested it here locally with `kubewarden-controller:latest` now that is released, and all fine!

PR #246, opened 2022-06-02
feat: Generate Bill Of Materials (BOM) file.

### LINKED ISSUE #228: Provide Software Bill Of Materials
We should generate a SBOM for the kubewarden controller.

Requirements:

  * [x] Each controller release has its own SBOM file generated
  * [x] The SBOM is signed using cosign
  * [x] The SBOM is attached to the GH Release as an artifact

jvanz:
Updates the Github workflow used in our CI pipelines to generate the BOM files and attach it to the container image and add it as an asset in the Github release.

Fix #228

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -49,6 +49,18 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
       -
+        name: Install Golang
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.17'
+      -
+        name: Install the bom command
+        shell: bash
+        run: go install sigs.k8s.io/bom/cmd/bom@v0.2.2
+      -
+        name: Install Cosign
+        uses: sigstore/cosign-installer@main
+      -
         name: Retrieve tag name
         if: ${{ startsWith(github.ref, 'refs/heads/') }}
         run: |
@@ -84,6 +96,17 @@ jobs:
           outputs: type=docker,dest=/tmp/kubewarden-controller-image-${{ env.TAG_NAME }}.tar
           tags: |
             ghcr.io/${{github.repository_owner}}/kubewarden-controller:${{ env.TAG_NAME }}
+      -
+        name: Create SBOM file
+        shell: bash
+        run: |
+          bom generate -n https://kubewarden.io/kubewarden.spdx -o kubewarden-controller.spdx .
+      -
+        name: Attach SBOM file in the container image
+        shell: bash
+        run: |
+          set -e
+          cosign attach sbom --sbom kubewarden-controller.spdx "ghcr.io/${{github.repository_owner}}/kubewarden-controller@${{ steps.build-image.outputs.digest }}"
       -
         name: Upload container image to use in other jobs
         if: ${{ inputs.push-image == false }}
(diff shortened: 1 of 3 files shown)


.github/workflows/container-image.yml
```diff
@@ -84,6 +96,16 @@ jobs:
...
+      -
+        name: Create SBOM file
+        shell: bash
+        run: |
+          bom generate -n https://kubewarden.io/kubewarden.spdx -o kubewarden-controller.spdx .
+      -
+        name: Attach SBOM file in the container image
+        shell: bash
+        run: |
+          cosign attach sbom --sbom kubewarden-controller.spdx ghcr.io/${{github.repository_owner}}/kubewarden-controller:${{ env.TAG_NAME }}
```
raulcabello, line 108: I think this is just attaching the sbom, but it does not sign it. This is the output I see:
```
cosign attach sbom --sbom kubewarden-controller.spdx raulcabm/sigstore:cosign
WARNING: Attaching SBOMs this way does not sign them. If you want to sign them, use 'cosign attest -predicate kubewarden-controller.spdx -key <key path>' or 'cosign sign -key <key path> <sbom image>'.
```

raulcabello: Can we use the digest instead of the tag? This is how we do it for signing https://github.com/kubewarden/kubewarden-controller/blob/main/.github/workflows/container-build.yml#L36

raulcabello (changes requested):

.github/workflows/container-image.yml
jvanz: Oh, for some reason the signing step is not committed. I did it already. Let me check what happened

jvanz: I've updated the PR pushing the missing steps. I'll update the PR to use digest next.

.github/workflows/container-image.yml
```diff
@@ -84,6 +96,17 @@ jobs:
...
+        name: Create SBOM file
+        shell: bash
+        run: |
+          bom generate -n https://kubewarden.io/kubewarden.spdx -o kubewarden-controller.spdx .
+      -
+        name: Attach SBOM file in the container image
+        shell: bash
+        run: |
+          set -e
+          cosign attach sbom --sbom kubewarden-controller.spdx "ghcr.io/${{github.repository_owner}}/kubewarden-controller:${{ env.TAG_NAME }}"
```
raulcabello, line 109: Can we use the digest from the output of `build-push-action` instead of the tag?

raulcabello: I see you've just done it. Thanks!

jvanz: Yes, I was working on that since your first review. Please take a look again. I just pushed the change.

raulcabello (approved):
LGTM, thanks!

.github/workflows/release.yml
```diff
@@ -21,3 +39,13 @@ jobs:
           release_name: Release kubewarden-controller ${{ github.ref }}
           draft: false
           prerelease: ${{ contains(github.ref, '-alpha') || contains(github.ref, '-beta') || contains(github.ref, '-rc') }}
+
+      - name: Upload SBOM file as release asset
```
flavio, line 43: Should we add another step to sign it?

flavio (approved):
LGTM, I left a minor comment that can be ignored

jvanz:
@kubewarden/kubewarden-developers,  I'm asking you for another round of review because I've changed the action used to create the release. The [action](https://github.com/actions/create-release) from the Github team is not maintained anymore.

raulcabello (approved):
LGTM

PR #247, opened 2022-06-03
ClusterAdmissionPolicy, AdmissionPolicy and PolicyServer V1 crds

### LINKED ISSUE #245: Feature Request: Create v1 CRDS
### Description

Create v1 for PolicyServer, ClusterAdmissionPolicy and AdmissionPolicy. We should keep v1alpha2, so upgrades from previous installation still work. There will be a conversion from v1alpha2 to v1.

v1 is a copy of v1alpha2, it just changes the version

Follow the steps from [kubebuilder documentation](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html)

raulcabello:
Create new v1 for the three CRDs. We keep v1alpha2, but we mark v1 as the one that should be stored in etcd. No conversion needed since the schema is identical.

Fix #245


viccuad (approved):
Some comments. A priori it looks good, but it's a difficult PR to review (thanks for committing the automated commands separate from the rest!) and I haven't tested locally.

viccuad (changes requested):
Whoops.. I just wanted to make a review with comments, without explicit approval. I mistakenly left "Approve". I can't retract my approval, so now I'm marking as "Request changes" just in case..

PROJECT
```diff
@@ -19,7 +18,6 @@ resources:
     webhookVersion: v1
 - api:
     crdVersion: v1
-    namespaced: false
```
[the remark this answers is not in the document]
raulcabello: yeah it looks like it's false if its not there because it was removed by kubebuilder. Nevertheless I added it [here](https://github.com/kubewarden/kubewarden-controller/pull/247/commits/84e2dd165f0d02ea77d1116f6ab40212740a3d57) since I also think it should be there

config/crd/bases/policies.kubewarden.io_policyservers.yaml
```diff
@@ -1,9 +1,10 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.4.1
```
ereslibre, line 7: I would prefer to generate this manifests with controller-tools at 0.8.0 at least (05 Jan 2022). 0.4.1 goes back to 17 Nov 2020.

config/crd/kustomization.yaml
```diff
@@ -20,6 +20,9 @@ patchesStrategicMerge:
 # - patches/cainjection_in_clusteradmissionpolicies.yaml
 # - patches/cainjection_in_policyservers.yaml
 # - patches/cainjection_in_admissionpolicies.yaml
+#- patches/cainjection_in_admissionpolicies.yaml
+#- patches/cainjection_in_clusteradmissionpolicies.yaml
+#- patches/cainjection_in_policyservers.yaml
```
ereslibre, line 25: This is duplicated

controllers/admissionpolicy_controller.go
```diff
@@ -66,7 +66,7 @@ func (r *AdmissionPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 // SetupWithManager sets up the controller with the Manager.
 func (r *AdmissionPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	err := ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha2.AdmissionPolicy{}).
+		For(&policiesv1.AdmissionPolicy{}).
```
ereslibre, line 69: I haven't tested it yet, I will later today. But I suspect if we only do this, existing `v1alpha2` resources are not going to be reconciled by this controller as it is. Note: if this is an issue, this applies to `ClusterAdmissionPolicy` and `PolicyServer` resources as well in their controllers.

What passive operation is going to "upgrade" the version of the existing objects? The idea in this kind of migration is that it happens passively: an object is read in `v1alpha2`, and the component writes it back after the reconciliation in the default stored version: `v1`, effectively migrating the resource. More info: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version.

So, for this PR, I think we have to also watch `v1alpha2` resources. By updating them in regular reconcile loops, they will get migrated to `v1`.

-- Aside, unrelated to this PR strictly:

For `1.1.0` we will have to:
- Remove the watch for `v1alpha2` resources on the controller.
- Remove the existing `v1alpha2` resources that have been ignored since `1.0`.
- Remove the `v1alpha2` CRD's.

At that point in time, only `v1` resources will exist and will be allowed.

-- /Aside

ereslibre (commented):
Thanks @raulcabello.

I have to really test if my assumption is correct in that no `v1alpha2` resources are going to be listed as it is, hence no migration would happen.

Some other comments as well. I will place another review after this first pass, but checking if my assumptions are correct.

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
```diff
@@ -1,9 +1,10 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.4.1
```
flavio, line 7: It looks like you used an older version of kubebuilder... This gives me doubts about the code that has been automatically generated. Am I a fool?

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
```diff
@@ -348,6 +350,340 @@ spec:
     storage: true
     subresources:
       status: {}
+  - additionalPrinterColumns:
```
flavio, line 353: Like this whole section, it didn't exist before. However this doesn't seem to be dangerous

main.go
```diff
@@ -182,7 +184,7 @@ func main() {
 func webhooks() []webhookwrapper.WebhookRegistrator {
 	return []webhookwrapper.WebhookRegistrator{
 		{
-			Registrator: (&v1alpha2.PolicyServer{}).SetupWebhookWithManager,
+			Registrator: (&policiesv1.PolicyServer{}).SetupWebhookWithManager,
```
flavio, line 187: shouldn't we still be looking after v1alpha2 policies?

controllers/admissionpolicy_controller.go
raulcabello: I also thought about that, however it seems to work fine in the local test I did. I created a PolicyServer and a ClusterAdmission  with v1alpha2 using an old version of the controller. Then I updated to the latest version of the controller and the controller could see changes in the the v1alpha2 resources (eg when I deleted the PolicyServer resource everything was removed)

I believe this is because we watch the v1 and k8s returned the v1alpha2 wth the v1 apiVersion:

```
When you read an object, you specify the version as part of the path. If you specify a version that is different from the object's persisted version, Kubernetes returns the object to you at the version you requested, but the persisted object is neither changed on disk, nor converted in any way (other than changing the apiVersion string) while serving the request. You can request an object at any version that is currently served.
```

more info https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects

but please do a local test and let me know the outcome, I might be wrong and this is important to get it right. Thanks!

main.go
raulcabello: this seems to be the same as https://github.com/kubewarden/kubewarden-controller/pull/247#discussion_r890870189

config/crd/bases/policies.kubewarden.io_policyservers.yaml
raulcabello: done in https://github.com/kubewarden/kubewarden-controller/pull/247/commits/f71b653c1b969cecaae9a1c1c1cbdc3e3e88c73f (I had an old version in my bin folder and make controller-gen did nothing until I removed it first)

apis/policies/v1/admissionpolicy_webhook.go
```diff
@@ -57,7 +56,7 @@ func (r *AdmissionPolicy) Default() {
 	}
 }
 
-//+kubebuilder:webhook:path=/validate-policies-kubewarden-io-v1alpha2-admissionpolicy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=admissionpolicies,verbs=create;update,versions=v1alpha2,name=vadmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
+//+kubebuilder:webhook:path=/validate-policies-kubewarden-io-v1-admissionpolicy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=admissionpolicies,verbs=create;update,versions=v1alpha2;v1,name=vadmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
```
flavio, line 59: This line still has a `v1alpah2` occurrence inside of it. Is that a leftover or is it intentional?

apis/policies/v1/clusteradmissionpolicy_webhook.go
```diff
@@ -46,7 +46,7 @@ func (r *ClusterAdmissionPolicy) SetupWebhookWithManager(mgr ctrl.Manager) error
 	return nil
 }
 
-//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1alpha2-clusteradmissionpolicy,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpolicies,verbs=create;update,versions=v1alpha2,name=mclusteradmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
+//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-clusteradmissionpolicy,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpolicies,verbs=create;update,versions=v1;v1alpha2,name=mclusteradmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
```
flavio, line 49: Same here, `v1alpha2` is still mentioned

apis/policies/v1/clusteradmissionpolicy_webhook.go
```diff
@@ -61,7 +61,7 @@ func (r *ClusterAdmissionPolicy) Default() {
 	}
 }
 
-//+kubebuilder:webhook:path=/validate-policies-kubewarden-io-v1alpha2-clusteradmissionpolicy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpolicies,verbs=create;update,versions=v1alpha2,name=vclusteradmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
+//+kubebuilder:webhook:path=/validate-policies-kubewarden-io-v1-clusteradmissionpolicy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpolicies,verbs=create;update,versions=v1alpha2;v1,name=vclusteradmissionpolicy.kb.io,admissionReviewVersions={v1,v1beta1}
```
flavio, line 64: Here too, there's another `v1alpha2`

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
jvanz: This commit is not PR branch. You need to rebase it or cherry pick it.

jvanz: image

controllers/admissionpolicy_controller.go
jvanz: I've tested the [upgrade process](https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/27) from an old Helm chart (using `v1alpha2` crd version) to latest one (using `v1` crd version). In my tests, all the `v1alpha2` running objects are migrated to the `v1` and others `v1alpha2` objects installed after the upgrade are stored as `v1` objects.

apis/policies/v1/admissionpolicy_webhook.go
raulcabello: I left it intentionally because I think Webhook should watch both versions (old versions stored in etcd are still v1alpha2, even though it will be transformed to v1 when you fetch them)
To be honest it might work fine with just v1, but I haven’t tested it

jvanz (approved):

config/crd/kustomization.yaml
raulcabello: fixed by @jvanz in https://github.com/kubewarden/kubewarden-controller/pull/247/commits/ff013d4ea058525b9f5f8dc156fb3541b058b9ac

ereslibre (approved):

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
raulcabello: yes it was already [there](https://github.com/kubewarden/kubewarden-controller/blob/main/config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml#L18) 
Now it is twice, one for v1alpha2 and another for v1

apis/policies/v1/admissionpolicy_webhook.go
raulcabello: removed [here](https://github.com/kubewarden/kubewarden-controller/pull/247/commits/84a0ba1a5d5a925b1cafd7e40e4ce9892dc1a62e) I did some manual test and it is working fine

apis/policies/v1/clusteradmissionpolicy_webhook.go
raulcabello: fixed in https://github.com/kubewarden/kubewarden-controller/pull/247/commits/84a0ba1a5d5a925b1cafd7e40e4ce9892dc1a62e

apis/policies/v1/clusteradmissionpolicy_webhook.go
raulcabello: fixed https://github.com/kubewarden/kubewarden-controller/pull/247/commits/84a0ba1a5d5a925b1cafd7e40e4ce9892dc1a62e

viccuad (approved):

apis/policies/v1/clusteradmissionpolicy_webhook.go
flavio: same here, sorry

apis/policies/v1/admissionpolicy_webhook.go
raulcabello: I removed them and it seems to be fine. It looks like it will fetch and transform v1alpha2 to v1 if we add the webhook just for v1

apis/policies/v1/clusteradmissionpolicy_webhook.go
raulcabello: commit reverted https://github.com/kubewarden/kubewarden-controller/pull/247/commits/348e840158cb069e358534e89051c10e56b10ec1

PR #250, opened 2022-06-15
remove v1alpha2 from the webhooks

raulcabello:
We are converting v1alpha2 resources into v1 using the none conversion strategy
If we add both apiVersions v1 and v1alpha2 to the webhooks we can't create v1alpha2 resources

Removed in the helm charts too https://github.com/kubewarden/helm-charts/pull/103


viccuad (approved):

viccuad:
Makes sense. Merging to consume the controller:latest image locally.

PR #254, opened 2022-07-21
chore: Update sample policies

viccuad:
## Test

To test this pull request, you can run the following commands:
Tested locally by instantiating them and seeing they come up.

PR #255, opened 2022-07-21
Configuration to disable SBOM generation.

jvanz:
Run running the E2E tests is not necessary generate the SBOM file. Therefore, the workflow used to build the container and generate the SBOM file now has a configuration to disable the SBOM file generation.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -11,6 +11,7 @@ jobs:
     uses: kubewarden/kubewarden-controller/.github/workflows/container-image.yml@main
     with:
       push-image: false
+      generate-sbom: false
   run-e2e-tests:
     name: "Tests"
     needs: [build]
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -8,9 +8,10 @@ on:
 jobs:
   build:
     name: "Build"
-    uses: kubewarden/kubewarden-controller/.github/workflows/container-image.yml@main
+    uses: jvanz/kubewarden-controller/.github/workflows/container-image.yml@main
```
viccuad, line 11: this would need to be changed back (I suppose it was for tests)

viccuad (approved):
LGTM, besides comment. I hit the same golangci-lint warning on my PR, fixed it there: https://github.com/kubewarden/kubewarden-controller/pull/254

.github/workflows/e2e-tests.yml
jvanz: Ops... thanks for the heads up

PR #256, opened 2022-08-02
Update policy server security context.

### LINKED ISSUE #203: Reduce permissions on deployed policy-server
When the controller deploys a `policy-server` it's doing so with many permissions. It's being deployed in privileged mode by default for example.

Reduce the permissions in the security context of the policy-server pod so that it only requires what it needs.

jvanz:
Updates the code where the policy server is created to configure the securityContext of the Policy Server container. The process cannot run in a privileged container, cannot run with root user, can run on a read only file system and all the capabilities are dropped.

Fix #203

### DIFF
--- a/internal/pkg/admission/policy-server-deployment.go
+++ b/internal/pkg/admission/policy-server-deployment.go
@@ -330,8 +330,19 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
 		)
 	}
 	enableReadOnlyFilesystem := true
+	privileged := false
+	runAsNonRoot := true
+	allowPrivilegeEscalation := false
+	capabilities := corev1.Capabilities{
+		Add:  []corev1.Capability{},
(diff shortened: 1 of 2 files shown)


raulcabello (approved):
LGTM, just make sure to fix the golangci-lint issue. Thanks!

flavio (commented):
LGTM, the golang-ci issues have to be fixed as reported by @raulcabello.

Important question: have you run this change through the e2e tests? I wonder if we have to do some changes to our container image of the policy-server to make it work.

viccuad (approved):
LGTM, yet I have the same question as Flavio.

jvanz:
> Important question: have you run this change through the e2e tests? I wonder if we have to do some changes to our container image of the policy-server to make it work.

Yes. I have. The only failing tests are the secure supply chain ones. But they are failing even with the released container image. Thus, it is not related to this change. I believe it is a test issue.

PR #257, opened 2022-08-04
Use main e2e tests branch instead of tag

kravciak:
## Description

Second part of https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/32#discussion_r934788987 for switching to main branch

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -15,7 +15,7 @@ jobs:
   run-e2e-tests:
     name: "Tests"
     needs: [build]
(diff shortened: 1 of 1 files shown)


viccuad (approved):

viccuad:
I'm ok with this change for now, if we want to iterate on them. But IMO we should tag the e2e tests together with the releases; the e2e tests of `current-version + 1` may not work with `current-version` because they test new features. Or viceversa if there are breaking changes. 

I would prefer to have some logic to automatically run the corresponding e2e tests tag for each version in the future.

raulcabello (approved):

PR #259, opened 2022-08-24
ci: Use actions/create-release

viccuad:
This is the offical GH action for releases, and we use it in kwctl and
others.

With this, I hope that release naming will be fixed to
  `Release kubewarden-controller v1.1.0`
from
  `Release kubewarden-controller refs/tags/v1.1.0`

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -22,4 +22,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: latest
+          version: v1.47.2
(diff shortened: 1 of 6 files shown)


jvanz (commented):
I do know if this is a good idea. The official Github action is not maintained anymore and the current actions used is one of Gihub replacement.

jvanz:
Afaics, we can continue to use to the the atual action. But we need to change a configuration. By default the actions uses the `github.ref` as the [release name](https://github.com/softprops/action-gh-release#-customizing). I believe that if we change it to [`GITHUB_REF_NAME`](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) the issue goes away.

viccuad:
True, good point! amended the PR, now it uses `GITHUB_REF_NAME`. Also, needed to pint golangci in GHA, or it will fail.

jvanz (approved):

PR #272, opened 2022-09-09
feat: Create SBOM of container image

viccuad:
## Description

Partial fix of https://github.com/kubewarden/kubewarden-controller/issues/261

As it stands on this commit, the functionality of creating SBOMs added, but never
shipped in a release.

One would need to download them from the `release.yml` worklow and add
them to the release, as done in policy-server.

This isn't as trivial as in policy-server, as we use a 3rd workflow in
`container-build.yml` to trigger `container-image.yml`, and that makes
it difficult (if not impossible) to download them. I have tried several incantations of workflows and failed.

The `container-image.yml` workflow is parametrized to build the
kubewarden-controller container image as a tar, to be consumed by the
e2e tests. This is done so we don't polute the registry with container
images for each branch/PR/main commit. Maybe we should revisit this and
find a different approach.

The `container-image.yml` workflow is being consumed in `e2e-tests.yml`, which passes the output
to
https://github.com/kubewarden/kubewarden-end-to-end-tests/blob/main/.github/workflows/e2e-tests.yml.
Then, it is
passed to kubewarden/setup-kubewarden-cluster-action@v1.1.0.

## Test

To test this pr on your fork,

1. delete fake tag: `git tag -d v1.2.0-rc1 && git push --delete yourremote v1.2.0-rc1`
2. delete old releases manually in your forks' release page
3. push branch to your own fork
2. create fake tag: `git tag -a v1.2.0-rc1 -m "v1.2.0-rc1" -s && git push yourremote main v1.2.0-rc1`
4. watch workflow runs

Example of a workflow run with SBOM artifacts: https://github.com/viccuad/kubewarden-controller/actions/runs/3023713161

### Potential improvement

Missing:
- Depend on ci.yml for workflows that should
- Currently we only sign production images. But the way we do it via `container-build.yml` is clumsy. Maybe it should be a conditional on `container-image.yml` when triggered by a tag.
- Remove `container-build.yml`, only use `container-image.yml`. Depend on `container-image.yml` for `e2e-tests.yml`.
- The tricky one: Release should depend on `container-image.yml` succeeding (so we have the SBOM artifacts in GHA) with something like:
  ```
  name: Release kubewarden-controller
  on:
    workflow_run:
      workflows: ["Build container image"]
      types:
        - completed
      branches:
        - "main"
    push:
      tags:
      - 'v*'

  jobs:
   release:
      name: Create release
      runs-on: ubuntu-latest
      if: ${{ github.event.workflow_run.conclusion == 'success' }}
  ```
  But then, I can't get release to trigger on tags..

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -24,8 +24,6 @@ on:
         description: "Image digest"
         value: ${{ jobs.build.outputs.digest }}
 
-
-
 jobs:
   build:
     name: Build container image
@@ -35,6 +33,10 @@ jobs:
       tag: ${{ steps.setoutput.outputs.tag }}
       artifact: ${{ steps.setoutput.outputs.artifact }}
       digest: ${{ steps.setoutput.outputs.digest }}
+    permissions:
+      # needed for cosign:
+      packages: write
+      id-token: write
     steps:
       -
         name: Checkout code
@@ -59,6 +61,7 @@ jobs:
           go-version: '1.17'
       -
         name: Install the bom command
+        if: ${{ inputs.generate-sbom == true }}
         uses: kubewarden/github-actions/kubernetes-bom-installer@v1
       -
         name: Install Cosign
@@ -87,10 +90,13 @@ jobs:
           tags: |
             ghcr.io/${{github.repository_owner}}/kubewarden-controller:${{ env.TAG_NAME }}
       -
+        name: Build linux/amd64 container image into .tar for e2e tests
+        # This step is only triggered from e2e tests, e2e tests consume the
+        # controller image as a tar.
+        #
         # Only build amd64 because buildx does not allow multiple platforms when
         # exporting the image to a tarball. As we use this only for end-to-end tests
         # and they run on amd64 arch, let's skip the arm64 build for now.
-        name: Build linux/amd64 container image
         if: ${{ inputs.push-image == false }}
         uses: docker/build-push-action@v3
         with:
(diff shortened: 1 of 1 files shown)


raulcabello (commented):
perhaps I'm missing something, but why do we want to add the container image sbom file to the release? We already attach (and sign) the sbom to the OCI registry with cosign. Users can retrieve sbom with `cosign download sbom`, this will also verify sbom was properly signed.

.github/workflows/container-image.yml
```diff
@@ -101,25 +107,85 @@ jobs:
...
+          COSIGN_EXPERIMENTAL: 1
+      -
+        name: Attach SBOM file of build in the container image
         if: ${{ inputs.generate-sbom == true }}
         shell: bash
         run: |
           set -e
-          cosign attach sbom --sbom kubewarden-controller.spdx "ghcr.io/${{github.repository_owner}}/kubewarden-controller@${{ steps.build-image.outputs.digest }}"
+          cosign attach sbom --sbom kubewarden-controller-container-image-sboms/kubewarden-controller.spdx \
+            "ghcr.io/${{github.repository_owner}}/kubewarden-controller@${{ steps.build-image.outputs.digest }}"
```
raulcabello, line 180: we don't sign the sbom we attach. We should probably sign it after this or use in-toto attestations
more info https://docs.sigstore.dev/cosign/other_types/#sboms-software-bill-of-materials

raulcabello:
It is very complicated to add the sbom to the release with the current approach as @viccuad mentioned. I run the container creation inside of the release workflow [here](https://github.com/kubewarden/kubewarden-controller/pull/276) then it is easy to add the sbom to the release.
I think it make sense to run the container-build inside of the release workflow, so we don't create a release until the container is created.

.github/workflows/container-image.yml
raulcabello: I saw we are signing it in a different step, sorry about the noise

flavio (approved):
LGTM, waiting for @viccuad to give his approval once he's back from vacation

flavio (changes requested):
I take my approval back, https://github.com/kubewarden/kubewarden-controller/pull/276 seems a better solution

viccuad:
Superseded by https://github.com/kubewarden/kubewarden-controller/pull/276.

PR #276, opened 2022-09-15
Run container-build inside of the release workflow and add container image sbom

raulcabello:
Run container-build inside of the release workflow, so the release is created after the container image is created.
Release will not be created if container creation failed.

Add container image sbom to the release

[This](https://github.com/raulcabello/kubewarden-controller/releases/tag/vsbomtest18) is how a release looks like

Fix https://github.com/kubewarden/kubewarden-controller/issues/261

Replaces https://github.com/kubewarden/kubewarden-controller/pull/272

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -2,6 +2,10 @@ name: Build container image every change.
 
 on:
   workflow_call:
+    outputs:
+      digest:
+        description: "Container image digest"
+        value: ${{jobs.build.outputs.digest}}
   push:
     branches:
       - "*"
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,11 +1,10 @@
 name: Build container image every change.
 
 on:
+  workflow_call:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


flavio (approved):
I think I prefer this approach over https://github.com/kubewarden/kubewarden-controller/pull/272

viccuad (approved):
Approving this.

With this in, we have:
- The SBOM of the build attached to the container image.
- The SBOM of the container image shipped in the release.

This simplifies needing to juggle things to upload artifacts or the sbom of the build so they are shipped in the GH release, simplifying things.

Still, the SBOM of the build is not signed, and the signature is not attached to the container image. I think we should provide those, too.

raulcabello:
@viccuad I also thought the sbom was not signed. But then I found out it is signed later in the [container-build workflow](https://github.com/kubewarden/kubewarden-controller/blob/main/.github/workflows/container-build.yml#L40-L47)

viccuad:
Oh, true. A bit convoluted (and may need changes for SLSA), but it's fine. Let's get this merged!

PR #279, opened 2022-09-20
Add OpenSSF best practices badge

raulcabello:
Fix https://github.com/kubewarden/kubewarden.io/issues/128

Please also review the answers in https://bestpractices.coreinfrastructure.org/projects/6502. Feel free to modify them


viccuad (approved):
LGTM!

viccuad:
I would consider the left over cryptographic practices to be met. But it feels so weird to mark them as met for the foreseeable, without an external audit..

PR #281, opened 2022-09-21
docs: Clarify `spec.module` defaults to `registry://`

viccuad:
## Description

Clarify `spec.module` defaults to `registry://`


flavio (approved):
LGTM, I think we have to regenerate the docs of our CRDs, don't we?

jvanz (approved):

viccuad:
> LGTM, I think we have to regenerate the docs of our CRDs, don't we?

we will need to, but we need a tagged release first. I peruse https://doc.crds.dev/github.com/kubewarden/kubewarden-controller from time to time..

PR #310, opened 2022-10-04
Add E2E tests badge

kravciak:
## Description

Add badge for e2e tests.


viccuad (approved):
Nice! Thanks!

jvanz (changes requested):
I'm fine with the changes. But I would say that we must fix the DCO issue. As a CNCF project we need to follow the processes in place. 

@kravciak , please, check the feedback from the DCO bot to know how to fix the problem. It seems a simple fix. After that, we can merge the change.

jvanz:
Thank you! :rocket:

PR #320, opened 2022-10-19
Move policies to pkg folder

raulcabello:
We need to import code from inside of the api folder (e.g. policies structs, policy interface,..) in the `audit-scanner`. In order to do this the api folder must be placed inside the pkg folder.
This will allow us to reuse all this code in the `audit-scanner`


jvanz (approved):
LGTM!

PR #321, opened 2022-10-21
v1.3.0 rc1

flavio:
Create a RC1 for the new kubewarden release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 ## [Unreleased]
 
 
+<a name="v1.3.0-rc1"></a>
+## [v1.3.0-rc1] - 2022-10-11
+### Bug Fixes
+- Function `RecordPolicyCount` should pass context
+
+### Code Refactoring
+- Allow defer functions to run
+- Changes for `go.opentelemetry.io/otel/sdk` 0.32.1
+
+### Pull Requests
+- Merge pull request [#304](https://github.com/kubewarden/kubewarden-controller/issues/304) from viccuad/bump-1.19
+- Merge pull request [#311](https://github.com/kubewarden/kubewarden-controller/issues/311) from kravciak/main
+
+
 <a name="v1.1.1"></a>
 ## [v1.1.1] - 2022-09-22
 ### Bug Fixes


CHANGELOG.md
```diff
@@ -174,207 +188,37 @@
 
 <a name="v0.4.4"></a>
 ## [v0.4.4] - 2022-01-20
-### Bug Fixes
```
raulcabello, line 177: should these messages from previous releases not be added?

Makefile
```diff
@@ -117,6 +117,13 @@ generate-crds: $(KUSTOMIZE) manifests kustomize ## generate final crds with kust
 	mkdir -p generated-crds
 	$(KUSTOMIZE) build config/crd -o generated-crds # If -o points to a folder, kustomize saves them as several files instead of 1
 
+.PHONY: tag
+tag:
+	@git tag "${TAG}" || (echo "Tag ${TAG} already exists. If you want to retag, delete it manually and re-run this command" && exit 1)
+	@git-chglog --output CHANGELOG.md
+	@git commit -m 'Update CHANGELOG.md' -- CHANGELOG.md
+	@git tag -f "${TAG}"
+
```
jvanz, line 126: Thanks for that!

jvanz: Don't we want to sign tags?

jvanz (approved):

CHANGELOG.md
flavio: ouch, this is the changelog generator messing around... What should we do? The changelog should not change over the time... is there a way to tell it to ignore these old things or should we just remove them once, to avoid this issue from happening again?

Makefile
viccuad: agree, it should be `git tag -s -a "${TAG}" -f "${TAG}` or similar.

CHANGELOG.md
viccuad: I vote for only staging the last version, and the bottom of the file with the compare links.

Makefile
```diff
@@ -117,6 +117,13 @@ generate-crds: $(KUSTOMIZE) manifests kustomize ## generate final crds with kust
 	mkdir -p generated-crds
 	$(KUSTOMIZE) build config/crd -o generated-crds # If -o points to a folder, kustomize saves them as several files instead of 1
 
+.PHONY: tag
+tag:
+	@git tag "${TAG}" || (echo "Tag ${TAG} already exists. If you want to retag, delete it manually and re-run this command" && exit 1)
+	@git-chglog --output CHANGELOG.md
+	@git commit -m 'Update CHANGELOG.md' -- CHANGELOG.md
+	@git tag -f "${TAG}"
```
viccuad, line 125: While generating chglog + tagging is nice, this means that the tag will be created for a code that is still to be reviewed and merged.

It's still also dependent on how people push, if only the code, or they are also pushing the tag.

Also, the PR of updating the changelog will not be included in the generated changelog. Which means that on next generation, it will pop-up (as it does now).

If we are happy with tag creation in unmerged code, we should change the github actions release workflow so it only runs when there's a tag, and that tag is merged in main, to not release unreviewed/unmerged code by mistake.

Makefile
flavio: added

Makefile
flavio: I temporarily removed the `make tag` target. We can add it back once we sort out the issue with the changelog generator

CHANGELOG.md
flavio: I cleaned up the changelog to ensure only new data is added, old parts of the file are not changed

jvanz (approved):

PR #332, opened 2022-10-27
Tag v1.3.0

flavio:
Everything is ready!


viccuad (approved):
Approving. Missing to stage the link lines at the end of the changelog, but since for the next version we may not have one.. I don't care :>

PR #335, opened 2022-11-03
feat: add permission to each job

raulcabello:
Build image gh action was broken when we introduced permissions: read-all at the top level of all files. 
This PR sets the permissions at job level instead, which fixes the issue. This should also be ok with CLOMonitor. We won't get the maximun score for the token permission (10) but it should give us more than 4 that is required for the check to be green. [More info](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions)
Everything worked fine when I created a release in my fork with [this changes](https://github.com/raulcabello/kubewarden-controller/actions/runs/3384112138)

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -10,14 +10,12 @@ on:
     branches:
       - "*"
 
-# Declare default permissions as read only.
-permissions: read-all
-
-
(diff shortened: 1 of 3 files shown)


flavio (approved):
LGTM

Issue #326, opened 2022-10-24
Policy-server-default pod is not recreated after deletion


kravciak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Default policy server pod is not recreated (deleted it manually) when there are clusteradmissionpolicies running.

### Expected Behavior

Pod is recreated when I delete it manually. It should not depend on existing clusteradmissionpolicies.

### Steps To Reproduce

Create kubewarden cluster with recommendedPolicies.enabled=True and delete policy-server pod.
```
# Install fresh kubewarden (crds, controller, defaults)
helm install --create-namespace -n kubewarden kubewarden-crds kubewarden/kubewarden-crds
helm install --wait -n kubewarden kubewarden-controller kubewarden/kubewarden-controller
helm install --wait -n kubewarden kubewarden-defaults kubewarden/kubewarden-defaults \
    --set recommendedPolicies.enabled=True

# Delete policyserver pod, it won't be recreated
kubectl delete pod -n kubewarden policy-server-default-86c5787d45-ldrpn
kubectl get pods -n kubewarden
    kubewarden-controller-759774675b-prwk7   1/1     Running   0          5m23s

# Delete cluster admission policies and policyserver pod will start again
kubectl delete clusteradmissionpolicies --all
kubectl get pods -n kubewarden
    kubewarden-controller-759774675b-prwk7   1/1     Running   0          6m29s
    policy-server-default-d8d6bdd9c-cnm4p    1/1     Running   0          54s
```

flavio:
@kravciak thanks for submitting this issue.

Some questions:
  * Does your cluster have some policy CR defined?
  * What is the status of the Deployment that controls the default policy server after the Pod is deleted? It should be its duty to automatically reconcile...
  * What happens if, after the Policy Server `default` pod is deleted, I create a policy? Is the Pod recreated?
  * If your cluster has no policies defined, what happens when, given a working cluster, you delete the Pod?

kravciak:
> Does your cluster have some policy CR defined?

Just defaults that come with installation, I did not crate any policies. Existing clusteradmissionpolicies went from active into pending state when I deleted policyserver pod.
```
ε k get customresourcedefinitions | grep kubewarden
policyservers.policies.kubewarden.io              2022-10-24T13:35:34Z
admissionpolicies.policies.kubewarden.io          2022-10-24T13:35:34Z
[...]

PR #338, opened 2022-11-04
fix: exclude kubewarden namespace in webhooks

raulcabello:
There is a problem when the PolicyServer pod is deleted. It can't be recreated because the webhook endpoint points to the PolicyServer which is not available. Error message:

```
Error creating: Internal error occurred: failed calling webhook "clusterwide-do-not-share-host-paths.kubewarden.admission": failed to call webhook: Post "https://policy-server-default.kubewarden.svc:8443/validate/clusterwide-do-not-share-host-paths?timeout=10s": dial tcp 10.100.59.167:8443: connect: connection refused
```

This will prevent any Pod from being created in the cluster. This PR excludes in the webhooks the namespace where Kubewarden was deployed.

Updates will work without doing any extra change. Controller will trigger a reconciliation for each policy the first time it starts. It will find a difference [here](https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/mutating-webhook.go#L50) between the old and new webhook, and then apply the changes with the namespace selector.

Fix https://github.com/kubewarden/kubewarden-controller/issues/326

Signed-off-by: raulcabello <raul.cabello@suse.com>

### DIFF
--- a/apis/policies/v1/clusteradmissionpolicy_types.go
+++ b/apis/policies/v1/clusteradmissionpolicy_types.go
@@ -151,7 +151,7 @@ func (r *ClusterAdmissionPolicy) GetRules() []admissionregistrationv1.RuleWithOp
 	return r.Spec.Rules
 }
 
-func (r *ClusterAdmissionPolicy) GetNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
+func (r *ClusterAdmissionPolicy) GetUpdatedNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
 	// exclude namespace where kubewarden was deployed
 	if r.Spec.NamespaceSelector != nil {
 		r.Spec.NamespaceSelector.MatchExpressions = append(r.Spec.NamespaceSelector.MatchExpressions, metav1.LabelSelectorRequirement{
--- a/apis/policies/v1/clusteradmissionpolicy_types.go
+++ b/apis/policies/v1/clusteradmissionpolicy_types.go
@@ -151,7 +151,24 @@ func (r *ClusterAdmissionPolicy) GetRules() []admissionregistrationv1.RuleWithOp
 	return r.Spec.Rules
 }
 
-func (r *ClusterAdmissionPolicy) GetNamespaceSelector() *metav1.LabelSelector {
+func (r *ClusterAdmissionPolicy) GetNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
+	// exclude namespace where kubewarden was deployed
+	if r.Spec.NamespaceSelector != nil {
+		r.Spec.NamespaceSelector.MatchExpressions = append(r.Spec.NamespaceSelector.MatchExpressions, metav1.LabelSelectorRequirement{
+			Key:      "kubernetes.io/metadata.name",
+			Operator: "NotIn",
+			Values:   []string{deploymentNamespace},
+		})
+	} else {
+		r.Spec.NamespaceSelector = &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{{
+				Key:      "kubernetes.io/metadata.name",
+				Operator: "NotIn",
+				Values:   []string{deploymentNamespace},
+			}},
+		}
+	}
+
 	return r.Spec.NamespaceSelector
 }
 
--- a/apis/policies/v1/admissionpolicy_types.go
+++ b/apis/policies/v1/admissionpolicy_types.go
@@ -115,7 +115,7 @@ func (r *AdmissionPolicy) GetMatchPolicy() *admissionregistrationv1.MatchPolicyT
 }
 
 // GetNamespaceSelector returns the namespace of the AdmissionPolicy since it is the only namespace we want the policy to be applied to.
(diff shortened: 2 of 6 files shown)


apis/policies/v1/clusteradmissionpolicy_types.go
```diff
@@ -151,7 +151,24 @@ func (r *ClusterAdmissionPolicy) GetRules() []admissionregistrationv1.RuleWithOp
 	return r.Spec.Rules
 }
 
-func (r *ClusterAdmissionPolicy) GetNamespaceSelector() *metav1.LabelSelector {
+func (r *ClusterAdmissionPolicy) GetNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
+	// exclude namespace where kubewarden was deployed
+	if r.Spec.NamespaceSelector != nil {
+		r.Spec.NamespaceSelector.MatchExpressions = append(r.Spec.NamespaceSelector.MatchExpressions, metav1.LabelSelectorRequirement{
```
flavio, line 157: I would look at all the existing Namespace selectors, and ensure this one that you're about to add, doesn't exist yet

flavio (commented):
Overall everything is ok, on top of the comment I left:

* Would it be possible to add a unit test?
* As pointed out by Victor, can this upgrade previously created webhook configurations to have this namespace selector?

apis/policies/v1/clusteradmissionpolicy_types.go
viccuad: Raul pointed out on the PR description that we are already updating the webhook with a deepEqual comparison [here](https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/mutating-webhook.go#L50). So I think we are good to go.

apis/policies/v1/clusteradmissionpolicy_types.go
```diff
@@ -151,7 +151,24 @@ func (r *ClusterAdmissionPolicy) GetRules() []admissionregistrationv1.RuleWithOp
 	return r.Spec.Rules
 }
 
-func (r *ClusterAdmissionPolicy) GetNamespaceSelector() *metav1.LabelSelector {
+func (r *ClusterAdmissionPolicy) GetNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
```
viccuad, line 154: Since we are changing the namespace selector, 

```suggestion
func (r *ClusterAdmissionPolicy) ReconcileNamespaceSelector(deploymentNamespace string) *metav1.LabelSelector {
```

Or something similar. Granted, that this function is only called in `ReconcileMutatingWebhookConfiguration()` and `UpdateMutatingWebhook()`, but..

raulcabello: It doesn't reconcile the namespace selectore inside the method, so I wouldn't add it to the name. This function belongs to the Policy interface, and it will also be called from the [background audit check](https://github.com/kubewarden/audit-scanner/blob/main/internal/policies/fetcher.go#L104) and maybe more places in the future.

apis/policies/v1/clusteradmissionpolicy_types.go
raulcabello: unit tests added here https://github.com/kubewarden/kubewarden-controller/pull/338/commits/1c12617050c0bc15600d6588b41241c96e706833

@flavio I think it is better to add a new `MatchExpression` excluding the kubewarden namespace. Looking to existing ones is not that easy because we also need to look to all the different operations. From k8s [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements)

> All of the requirements, from both matchLabels and matchExpressions are ANDed together -- they must all be satisfied in order to match.

So it doesn't really matter is this namespace also exists in a different `MatchExpression`

flavio: Thanks for the clarification

flavio (approved):

apis/policies/v1/clusteradmissionpolicy_types.go
viccuad: how about `GetAndUpdateNamespaceSelector`? I'm bound to forget that a getter actually adds a NamespaceSelector and bumps the object.

Feel free to merge as it is, if you still think is overkill.

raulcabello: I doesn't really update the selector, just returns a modified version of it. The implementation is different from [ClusterAdmissionPolicy](https://github.com/raulcabello/kubewarden-controller/blob/controller-bug/apis/policies/v1/clusteradmissionpolicy_types.go#L154) and [AdmissionPolicy](https://github.com/kubewarden/kubewarden-controller/blob/main/apis/policies/v1/admissionpolicy_types.go#L118).
What about `GetUpdatedNamespaceSelector`? or `GetModifiedNamespaceSelector`?

viccuad: both are fine! `GetUpdatedNamespaceSelector` sounds good, thanks!

raulcabello: changed https://github.com/kubewarden/kubewarden-controller/pull/338/commits/5404bf4333d15f7133b199806aa1a847afb6f1fa

PR #358, opened 2022-11-30
Enable release drafter

### LINKED ISSUE #351: Automatic changelog generation
In order to automatically release `kubewarden-controller` is necessary to update the CI/CD pipeline to generate and publish the change log automatically. This issue should track this change and document what is necessary and how to do it.

jvanz:
## Description

Enable the release drafter action to generate the release notes automatically when PR are merged into the main branch.

With this changes the release notes will be written in the Github release page by the release-drafter. The release drafter uses the labels in the PR to define which kind changes are being pushed in the PR. As well as which version should be updated (major, minor or patch). 

Fix #351


jvanz:
The steps to publish the release and upload assets can be shared with other repositories. But I would leave to do that when adding these steps there are well. I'll use this PR as an opportunity to get your feedback in the current approach.

.github/release-drafter.yml
```diff
@@ -0,0 +1,50 @@
...
+  - duplicate
+  - invalid
+  - later
+  - wontfix
+  - question
+  - skip-changelog
+
+change-template: '- $TITLE (#$NUMBER)'
+change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
+tag-template: 'v$RESOLVED_VERSION'
```
viccuad, line 23: Looks good!

I have to say I am uncomfortable with setting a version programmatically already.
I would like first to have a linter so all PRs follow the conventional commits style, so we are sure to never forget a possible breaking change.

Even with that, I would be more comfortable producing some releases via release-drafter yet keeping tags and version bumps manually.

jvanz: I've changed the configuration to keep the next release with the "Draft" name. The release workflow will update the release name with the tag pushed by the maintainers.

viccuad (commented):
So, I read more about release-drafter, and I wonder if marking the PRs via PR labels is a good fit for us. At the very least, we would need to make them compatible with the labels that we are migrating to. Only org maintainers can set labels, which is good as we know if a change is a breaking one.

Another option could be to use something like https://github.com/zeke/semantic-pull-requests (deprecated, but see list of alternatives). semantic-pull-requests does:
- expect either the PR title with a [conventionalcommit](https://www.conventionalcommits.org/en/v1.0.0/) prefix, or at least 1 commit of the PR. With that, it can know if to bump major version (backwards incompatible `fix!`, `feat!`, `refactor!`..), minor version (backwards compatible `feat`), or minor `fix`).
This approach of linting PR titles and/or commits is good for external contributors, as org maintainers can always edit the PR title. For editing the commit messages, one needs to ask maintainers to rebase (cumbersome), or to squash the PR after approval.

I personally favour linting for "conventional commit prefix" PR titles because they can be edited easier.

jvanz:
@viccuad, I've fine using conventional commit standard. We can do the same thing using release-drafter. It has a feature called [autolabeler](https://github.com/release-drafter/release-drafter#autolabeler). Which is able to tag PR based title, body, files and branch. Actually, I've already started to use it in this PR. See here:

```yaml
autolabeler:
  - label: 'area/dependencies'
    title: 'chore(deps)'
  - label: 'kind/enhancement'
    title: 'feat'
```
This configuration tells release-drafter to tag a PR with a title staring with `chore(deps)`  with the `area/dependencies` label. As well as tagging a PR with title starting with `feat` with the tag `kind/enhancement`. Is not this what do you want?

We can use the same feature to mark a PR with breaking changes. Furthermore, we can then show breaking changes in a special section of the release note, if we want to.

viccuad:
if we go that way, we have to indeed add support for breaking changes, as they are the only reason by a version bump is a "major" and not a "minor" nor "fix". I wonder how release-drafter can decide the specific semantic version to use, if it doesn't detect those breaking changes.

jvanz:
> if we go that way, we have to indeed add support for breaking changes, as they are the only reason by a version bump is a "major" and not a "minor" nor "fix". I wonder how release-drafter can decide the specific semantic version to use, if it doesn't detect those breaking changes.

In the latest changes, the release-drafter will not define the version. I've addressed your previous comment and configure it to create the draft release with the name "Draft". The release workflow will change the release name using the tag name. Therefore, the release-drafter will not define the next version. 

If you want the release-drafter to have a breaking changes section, we only need to have a tag for that. Then, we add a new configuration section, something like:

```yaml
categories:
  - title: 'Breaking changes'
    labels:
      - 'breaking'
      - 'major'
```

This label (`breaking` or `major`) can be added manually or if we want to use autolabeler. If we decided to use autolabeler, we can define that a PR with `!` is a breaking change. For example, `feat!: ...` or `fix!:...` is a breaking change.

viccuad:
I would be happy with support for breaking changes, to correctly bump major versions. I suppose that means support for `feat!:`, `fix!:`, `refactor!:`, but also `feat(webhooks)!: ` and the like.

.github/release-drafter.yml
```diff
@@ -0,0 +1,54 @@
...
+change-template: '- $TITLE (#$NUMBER)'
+change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
+name-template: 'Draft'
+template: |
+  $CHANGES
+
+autolabeler:
+    # Tag any PR with "!" in the subject as major update. In other words, breaking change
+  - label: 'major'
+    title: '/.*!:.*/'
```
jvanz, line 33: @viccuad, I've added this configuration to tag any PR with `!` in the title subject/scope as  major update.

jvanz:
@kubewarden/kubewarden-developers , this is an example of the release notes generated from the latest configuration:
image

Are you happy with it? Do you want to remove the emojis?

.github/release-drafter.yml
viccuad: perfect! many thanks! :green_heart:

viccuad (approved):
I'm happy with the emojis, approving!

.github/release-drafter.yml
jvanz: @viccuad, please double check the other autolabeler rules for the other types of changes.

.github/release-drafter.yml
```diff
@@ -0,0 +1,61 @@
+categories:
+  - title: '⚠️  Breaking changes'
+    labels:
+      - 'major'
```
viccuad, line 4:
```suggestion
      - 'kind/major'
```
I wonder if it should be so

jvanz: I have the same question. But the `kind/*` labels should represent the "kind" or "type" of the change. I'm not 100% sure if this is the case. Maybe should we use `size/major`, 

@kkaempf , what do you think?

viccuad: suggestion: `kind/breaking-change`

kkaempf: `size` is like effort estimation, which in reality never really works.
Issues are either "fully defined and actionable" or "unclear, needs refinement".
The former are normal issues, the latter epics. :wink:

jvanz: Done. Now the labels used to set a PR as "major" update can be `kind/breaking-change` or `kind/major`.

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,42 @@
+name: Release Drafter
+
+on:
+  workflow_dispatch:
+  push:
+    # branches to consider in the event; optional, defaults to all
+    branches:
+      - main
+  # pull_request event is required only for autolabeler
+  pull_request:
```
olblak, line 10: This workflow seems to be only for release-drafter so we could remove all mention to autolabeler and only run this workflow on dispatch event or on main branch change

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,42 @@
...
+  update_release_draft:
+    permissions:
+      # write permission is required to create a github release
+      contents: write
+      # write permission is required for autolabeler
+      # otherwise, read permission is required at least
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      # (Optional) GitHub Enterprise requires GHE_HOST variable set
```
olblak, line 30: I guess you can remove this comment :)

.github/workflows/release-drafter.yml
olblak: I got confused by https://github.com/github/issue-labeler

PR #371, opened 2022-12-23
Use WORKFLOW_PAT token to trigger CI.

jvanz:
## Description

Updates the release.yaml workflow file to use the WORKFLOW_PAT token to trigger the Helm chart update. This token is already used for that in past CI jobs.


jvanz:
I'm merging this because I'm pretty confidence this is a change with no issue. And considering that most of the team is on vacations the review will take some time.

PR #385, opened 2023-01-24
add security configuration in the policy server definition

jvanz:
## Description

Fix #XXX

## Test


internal/pkg/admission/policy-server-deployment.go
```diff
@@ -329,22 +329,9 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
...
 		)
 	}
-	enableReadOnlyFilesystem := true
-	privileged := false
-	runAsNonRoot := true
-	allowPrivilegeEscalation := false
-	capabilities := corev1.Capabilities{
-		Add:  []corev1.Capability{},
-		Drop: []corev1.Capability{"all"},
+	if policyServer.Spec.Security.Container != nil {
```
flavio, line 332: I would prefer to not drop the explicit defaults we had before. The code should be something like that:

```go
if policyServer.Spec.Security.Container != nil {
  // use whatever is defined inside of the CRD
} else {
  // the previous code
}
```

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +210,187 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
 	}
 }
 
-func TestPolicyServerSecurityContext(t *testing.T) {
+func TestPolicyServerContainerSecurityContext(t *testing.T) {
```
flavio, line 213: nitpicking

```suggestion
func TestPolicyServerWithContainerSecurityContext(t *testing.T) {
```

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +210,187 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
 	if deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities == nil ||
 		len(deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Add) > 0 ||
 		len(deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Drop) != 1 ||
 		deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Drop[0] != "all" {
-		t.Error("Policy server container should drop all capabilities")
+		t.Error("Policy server container Capabilities diverge from the expected value")
+	}
+}
+
+func TestPolicyServerPodSecurityContext(t *testing.T) {
```
flavio, line 266: nitpicking:

```suggestion
func TestPolicyServerWithPodSecurityContext(t *testing.T) {
```

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +210,187 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
+		if *deployment.Spec.Template.Spec.SecurityContext.RunAsUser != user {
+			t.Error("Pod RunAsUser diverges from the expected value")
+		}
+		if *deployment.Spec.Template.Spec.SecurityContext.RunAsGroup != group {
+			t.Error("Pod RunAsGroup diverges from the expected value")
+		}
+	}
+}
+
+func TestPolicyServerWithNoSecurityContext(t *testing.T) {
```
flavio, line 302: nitpicking: 

```suggestion
func TestPolicyServerWithoutSecurityContext(t *testing.T) {
```

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +210,187 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
+	}
+	policyServer := &policiesv1.PolicyServer{
+		Spec: policiesv1.PolicyServerSpec{
+			Image:    "image",
+			Security: policiesv1.PolicyServerSecurity{},
+		},
+	}
+	deployment := reconciler.deployment("v1", policyServer)
+
+	if deployment.Spec.Template.Spec.SecurityContext != nil {
```
flavio, line 315: Based on a previous comment, I think we should set a good default when the PolicyServer CRD doesn't provide guidance. This test should make sure our defaults are picked.

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +210,187 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
+		t.Error("Pod securityContext should not be defined ")
+	}
+
+	if deployment.Spec.Template.Spec.Containers[0].SecurityContext != nil {
+		t.Error("Container securityContext should not be defined ")
+	}
+}
+
+func TestPolicyServerPodAndContainerSecurityContext(t *testing.T) {
```
flavio, line 324: nitpicking:

```suggestion
func TestPolicyServerWithPodAndContainerSecurityContext(t *testing.T) {
```

apis/policies/v1/policyserver_types_test.go
```diff
@@ -0,0 +1,57 @@
...
+	privileged := true
+	containerSecurity := corev1.SecurityContext{
+		Privileged:   &privileged,
+		Capabilities: &capabilities,
+	}
+	podSecurity := corev1.PodSecurityContext{
+		RunAsUser:  &user,
+		RunAsGroup: &group,
+	}
+	_ = PolicyServerSpec{
```
flavio, line 42: This is a bit strange test, is it checking that we can define an object like the one described inside of `policyserver_types.go`?

If that's the case, it feels a bit like testing the code generation capabilities of kubebuilder. I don't think this is worth the effort, but I'm not against leaving this code around

apis/policies/v1/policyserver_types.go
```diff
@@ -21,6 +21,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Security configuration to be used in the Policy Server workload
```
viccuad, line 24: Nit:

```suggestion
// PolicyServerSecurity defines securityContext configuration to be used in the Policy Server workload
```

(godoc style)

apis/policies/v1/policyserver_types.go
```diff
@@ -65,6 +75,13 @@ type PolicyServerSpec struct {
 	// key named verification-config in the Configmap.
 	// +optional
 	VerificationConfig string `json:"verificationConfig,omitempty"`
+
+	// Security configuration to be used in the Policy Server workload.
+	// The field allows different configurations for the pod and containers.
+	// This configuration will not be used in containers added by other
+	// controllers (e.g. telemetry sidecars)
+	// +optional
+	Security PolicyServerSecurity `json:"security,omitempty"`
```
viccuad, line 84: Nitpicking, and annoying maybe this late in the process:

```suggestion
	SecurityContexts PolicyServerSecurity `json:"securityContexts,omitempty"`
```

This way looking at `PolicyServer.spec.securityContexts.container` & `PolicyServer.spec.securityContexts.pod` is self-explainable.

apis/policies/v1/policyserver_types_test.go
viccuad: Agree here.

internal/pkg/admission/policy-server-deployment.go
viccuad: I would also like this.

apis/policies/v1/policyserver_types_test.go
jvanz: File removed. ;)

internal/pkg/admission/policy-server-deployment_test.go
jvanz: Done

config/crd/bases/policies.kubewarden.io_policyservers.yaml
```diff
@@ -173,7 +173,7 @@ spec:
                 description: Replicas is the number of desired replicas.
                 format: int32
                 type: integer
-              security:
+              securityContexts:
```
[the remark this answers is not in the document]
jvanz: I'm fine with that. But I think that @viccuad suggested to use the plural word because we have 2 `securityContext` in this field. One for pod and another for container.

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -329,8 +329,25 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
...
+			Drop: []corev1.Capability{"all"},
+		}
+		admissionContainerSecurityContext := corev1.SecurityContext{
+			ReadOnlyRootFilesystem:   &enableReadOnlyFilesystem,
+			Privileged:               &privileged,
+			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+			Capabilities:             &capabilities,
+			RunAsNonRoot:             &runAsNonRoot,
+		}
+		admissionContainer.SecurityContext = &admissionContainerSecurityContext
```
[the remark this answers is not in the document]
jvanz: Done.

jvanz:
@flavio @viccuad I've added a default `securityContext` for the pod as well. It will only set the `RunAsNonRoot` to `true`.

config/crd/bases/policies.kubewarden.io_policyservers.yaml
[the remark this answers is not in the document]
jvanz: Should I still change to `securityContext`?

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +211,247 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
+		t.Error("Policy server container should have capabilities defined")
+	} else {
+		if len(deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Add) > 0 {
+			t.Error("Policy server container should not have 'Add' capabilities defined")
+		}
+		if len(deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Drop) != 1 ||
+			deployment.Spec.Template.Spec.Containers[0].SecurityContext.Capabilities.Drop[0] != dropCapabilityAll {
+			t.Error("Policy server container Capabilities should have only one 'All' drop capability")
+		}
+	}
```
flavio, line 338: Here we should compare the actual values with the expected ones, where the expected ones are returned by the `defaultPodSecurityContext` function (if I remember the name of the function properly)

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -210,38 +211,247 @@ func TestIfPolicyServerImageChanged(t *testing.T) {
...
+	if *securityContext.Privileged != false {
+		t.Error("Policy server container Privileged diverge from the expected value")
+	}
+	if *securityContext.RunAsNonRoot != true {
+		t.Error("Policy server container RunAsNonRoot diverges from the expected value")
+	}
+	if *securityContext.AllowPrivilegeEscalation != false {
+		t.Error("Policy server container AllowPrivilegeEscalation diverge from the expected value")
+	}
```
flavio, line 437: Same here, we should use the value returned by the `defaultContainerSecurityContext` as expected values. Maybe we can even use some k8s library that does a deep comparison of the two objects

config/crd/bases/policies.kubewarden.io_policyservers.yaml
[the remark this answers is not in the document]
flavio: No sorry, it's my fault. The new name is perfect because of the two different security context objects being hold by this struct.

internal/pkg/admission/policy-server-deployment_test.go
jvanz: Done

internal/pkg/admission/policy-server-deployment_test.go
flavio: Fine

flavio (approved):

flavio:
I think this can be merged, once rebased. I'm fine with the change

PR #405, opened 2023-02-27
contextAwareResources field in the policies.yml.

### LINKED ISSUE #401: Policy Server configuration: take `contextAwareResources` field into consideration
> **Note:** this issue depends on https://github.com/kubewarden/kubewarden-controller/issues/400

When generating the `policies.yml` file of a Policy Server instance, the controller has to take into account the freshly added `contextAwareResources` attribute.

The configuration snippet for a `ClusterAdmissionPolicy` that has `contextAwareResources` details, will look like that:

```yaml
namespace-validate:
  url: file:///tmp/namespace-validate-policy.wasm
  allowedToMutate: false
  contextAwareResources:
  - apiVersion: v1
    kind: Pod
  - apiVersion: apps/v1
    kind: Deployment
  - apiVersion: networking.k8s.io/v1
    Kind: Ingress
  settings:
    valid_namespace: valid
```

There's a new `contextAwareResources` attribute that is basically a dump of the `contextAwareResources` field of the `ClusterAdmissionPolicy` Custom Resource

jvanz:
Adds the contextAwareResources field in the policies.yml file. This field is added when a ClusterAdmissionPolicy is configured to has access to resource in the cluster setting the `contextAwareResources` field in the policy definition.

Fix #401

### DIFF
--- a/internal/pkg/admission/policy-server-configmap_test.go
+++ b/internal/pkg/admission/policy-server-configmap_test.go
@@ -64,7 +64,7 @@ func TestArePoliciesEqual(t *testing.T) {
 }
 
 func TestCreatePoliciesMap(t *testing.T) {
-	reconciler, validationPolicy, mutatingPolicy := createReconciler()
+	reconciler, validationPolicy, mutatingPolicy, contextAwarePolicy := createReconciler()
 
 	var policies PolicyConfigEntryMap
 	clusterAdmissionPolicies := []policiesv1.Policy{}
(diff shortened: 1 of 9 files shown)


internal/pkg/admission/policy-server-configmap_test.go
```diff
@@ -73,9 +73,9 @@ func TestCreatePoliciesMap(t *testing.T) {
 		t.Error("Empty ClusterAdmissionPolicyList should generate empty policies map")
 	}
 
-	clusterAdmissionPolicies = []policiesv1.Policy{&validationPolicy, &mutatingPolicy}
+	clusterAdmissionPolicies = []policiesv1.Policy{&validationPolicy, &mutatingPolicy, &contextAwarePolicy}
 	policies = reconciler.createPoliciesMap(clusterAdmissionPolicies)
-	if len(policies) != 2 {
+	if len(policies) != 3 {
 		t.Error("Policy map must has 2 entries")
```
flavio, line 79: nitpicking: "Policy must **have** 2 entries"

flavio (approved):
LGTM. So there's no need to write code for the reconciliation, everything is handled automatically.

Like, if I have a policy with 0 contextAwareResources, and then I update it to have access to `v1.pods`, this is going to cause the Policy Server to be redeployed. Am I right?

jvanz:
> Like, if I have a policy with 0 contextAwareResources, and then I update it to have access to `v1.pods`, this is going to cause the Policy Server to be redeployed. Am I right?

Yes. That's right.

PR #421, opened 2023-03-31
fix: allow Policy Server to consume policies from authenticated registries

flavio:
This PR allows Policy Server to pull policies from registries that are using authentication.

While working on this fix I had to introduce 2 other smaller fixes.

## chore(deps): update go

This was done just to ensure a more recent version of Go is being used. I also ran `go mod tidy` which performed some little cleanups

## fix: policy-server pull from authenticated registries

After the 1.5.0 release, Policy Server changed the source code that parses the docker credentials.
The new code comes from a 3rd party Rust crate: docker_credentials.

The library looks for for a file called`docker.config` file under the `~/.docker` directory.

The Kubernetes secret holding the docker credentials was mounted to `/home/kubewarden/docker/.dockerconfigjson`, causing Policy Server to not be able to find the credentials. Hence, pulling from a registry using authorization resulted in an anonymous pull, which then lead to a pull error.

This code changes the Deployment of Policy Server to mount the secret under the expected location: `/home/kubewarden/.docker/config.json`.

## fix: allow kubectl debug to be used against Policy Server

This fix was made necessary because I needed to debug what was going wrong with the Policy Server deployment.

After the 1.5.0 release, some changes have been made to the security context of the Pod running Policy Server. By default the Pod had a securityContext that prevented the creation of containers whose entrypoint was executed by the root user.

This restriction makes impossible to use kubectl debug against the policy-server container. That's because the sidecar debug container requires to run as root in oder to perform debug operations.

This commit removes the default securityContext added to the Policy Server Pod. We leave the ability to the end user to specify his own securityContext to be applied to the Policy Server Pod.

Note well: this change does not make the Policy Server deployment less secure. As a matter of fact, the Policy Server Pod is made by 1 single container: the one running the actual Policy Server. This container has its own securityContext which already enforces the root user to NOT be used.

The securityContext we previously added to the Pod was affecting the sidecars being added to the Pod. Potentially breaking them.


flavio:
@kravciak this is a bad regression that we introduced. Unfortunately there's no easy way to test that in isolation inside of `policy-server`. We need to extend the e2e test to cover pulling images from a registry that uses authentication.

Can you please file an issue to keep track of that?

jvanz (approved):

PR #426, opened 2023-04-06
fix: Only watch on kubewarden ns for namespaced resources

viccuad:
## Description

fix: Only watch on kubewarden ns for namespaced resources.
This should also improve resource consumption on clusters with a lot of Pods, as we would not be triggering with their changes.
This introduces no changes for cluster-scope resources.

## Test

With a controller with this, use rolebinding instead of clusterrolebinding for `kubewarden-controller-manager-cluster-role`.
Everything should still work fine, and the controller logs shouldn't have:
```
*v1.Pod: pods is forbidden: User "system:serviceaccount:kubewarden:kubewarden-controller" cannot list resource "pods" in API group "" at the cluster scope
```

### Potential improvement

See https://togithub.com/kubernetes-sigs/controller-runtime/blob/main/designs/use-selectors-at-cache.md
See https://togithub.com/kubernetes-sigs/controller-runtime/issues/244

### DIFF
--- a/main.go
+++ b/main.go
@@ -134,6 +134,12 @@ func main() {
 			LeaderElection:         enableLeaderElection,
 			LeaderElectionID:       "a4ddbf36.kubewarden.io",
 			ClientDisableCacheFor:  []client.Object{&corev1.ConfigMap{}, &appsv1.Deployment{}},
+			// Watch for kubewarden namespace on namespaced resources:
+			// Note: If a namespace is specified, controllers can still Watch for a
+			// cluster-scoped resource (in our case, ClusterAdmissionPolicies,
+			// webhooks, etc). For namespaced resources, the cache will only
+			// hold objects from the desired namespace.
+			Namespace: deploymentsNamespace,
 		},
 		setupLog,
 		environment.developmentMode,


main.go
```diff
@@ -134,6 +134,12 @@ func main() {
 			LeaderElection:         enableLeaderElection,
 			LeaderElectionID:       "a4ddbf36.kubewarden.io",
 			ClientDisableCacheFor:  []client.Object{&corev1.ConfigMap{}, &appsv1.Deployment{}},
+			// Watch for kubewarden namespace on namespaced resources:
```
viccuad, line 137: (randomly opening thread here)

So, I think this change is succint. It may be not transparent enough, I suppose. But we should never need to look outside of our namespace, besides for cluster-scoped resources.

Another option is to configure the Manager being used in `SetupWithManager()` , for both the clusteradmissionpolicy_controller.go and admissionpolicy_controller.go. E.g, here:

https://github.com/kubewarden/kubewarden-controller/blob/50d6c05f92f85e68bc50c5403b70aa0cbbbc2a3f/controllers/clusteradmissionpolicy_controller.go#L69-L73

Something like:
```go
namespace := deploymentsNamespace;
options := ctrl.Options{
    Namespace: cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
}
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), options)
// and then keep using it:
err := ctrl.NewControllerManagedBy(mgr)
(...)
```

In addition, it may be possible to move away from using a broad `Watches()`, and filter for Pods with out own labels. This would improve efficiency on clusters where there's a lot of Pods. But it shouldn't be an issue anymore, as we will be looking only into the kubewarden namespace anyways.

viccuad:
I haven't tested this locally yet.

flavio:
I prefer this change, compared to the one you propose inside of the comment.

jvanz (approved):
LGTM. I've tested the change locally. I've built a controller container image and updated the Helm charts to use `Role` and `RoleBinding`. When using the latest stable version image I can see the error message from the PR description. Then, I've tested the container image with this changes, the error disappear and  policy server runs fine.

viccuad:
thanks for testing @jvanz!

PR #433, opened 2023-04-20
fix: controller failed to update policy server service.

### LINKED ISSUE #432: Default policy server does not create metrics port on edit
When I edit existing default policyServer and enable telemetry it should create metrics port.
This does not happen, metrics port is created only when I enable telemetry during initial chart installation.

All grafana fields are then filled with zeroes because of this.

### Current behavior
Enabling telemetry on existing policyServer does not create metrics port:
```yaml
# Clusters -> local -> Instaled Apps -> rancher-kubewarden-defaults -> Edit
policyServer:
  telemetry:
    enabled: false -> true

# kubectl get svc -n cattle-kubewarden-system policy-server-default -o yaml | yq -r '.spec.ports'
- name: policy-server
  port: 8443
  protocol: TCP
  targetPort: 8443
```

### What works (workaround)
- delete existing policy server (Instaled Apps -> rancher-kubewarden-defaults -> delete)
- create default policy server again, enable telemetry `DURING` installation

```yaml
# kubectl get svc -n cattle-kubewarden-system policy-server-default -o yaml | yq -r '.spec.ports'
- name: policy-server
  port: 8443
  protocol: TCP
  targetPort: 8443
- name: metrics
  port: 8080
  protocol: TCP
  targetPort: 8080
```

### Steps to reproduce:
- install opentelemetry
- install kubewarden + default policyserver from Rancher UI
- install monitoring from `Rancher UI -> Cluster tools` (change `prometheus -> additionalServiceMonitors` as described [here](https://docs.kubewarden.io/operator-manual/ui-extension/metrics#1-within-the-cluster-explorer-click-on-cluster-tools-in-the-side-navigation), use `cattle-kubewarden-namespace`
- edit rancher-kubewarden-controller app, set `telemetry: enabled: false -> true`
- edit rancher-kubewarden-defaults app, set `policyServer: telemetry: enabled: false -> true`

jvanz:
## Description

The controller is not able to update the policy server service because it always call the create function in the reconciliation loop. Therefore, if the user edits the policy server enabling telemetry, the service will not be updated adding the metric port in the service.

To fix that, this change adds a call to the update function when the creation failed due an "already exists" error.

Fix #432

### DIFF
--- a/internal/pkg/admission/policy-server-service.go
+++ b/internal/pkg/admission/policy-server-service.go
@@ -35,8 +35,13 @@ func init() {
 }
 
 func (r *Reconciler) reconcilePolicyServerService(ctx context.Context, policyServer *policiesv1.PolicyServer) error {
-	err := r.Client.Create(ctx, r.service(policyServer))
-	if err == nil || apierrors.IsAlreadyExists(err) {
+	service := r.service(policyServer)
+	err := r.Client.Create(ctx, service)
+
+	if err != nil && apierrors.IsAlreadyExists(err) {
+		err = r.Client.Update(ctx, service)
+	}
+	if err == nil {
 		return nil
 	}
 	return fmt.Errorf("cannot reconcile policy-server service: %w", err)


viccuad (approved):
LGTM!

PR #442, opened 2023-05-02
Set telemetry configuration in policy server deployment.

jvanz:
## Description

Updates the telemetry configuration in all the policy servers by default if the Kubewarden controller has telemetry enabled.

Related to https://github.com/kubewarden/ui/issues/287

### DIFF
--- a/internal/pkg/admission/policy-server-deployment.go
+++ b/internal/pkg/admission/policy-server-deployment.go
@@ -226,6 +226,18 @@ func (r *Reconciler) adaptDeploymentSettingsForPolicyServer(policyServerDeployme
 			},
 		)
 	}
+	if r.MetricsEnabled {
+		policyServerDeployment.Annotations[constants.OptelInjectAnnotation] = "true"
+	}
+}
+
+func envVarsContainVariable(envVars []corev1.EnvVar, envVarName string) int {
+	for i, envvar := range envVars {
+		if envvar.Name == envVarName {
+			return i
+		}
+	}
+	return -1
 }
 
 func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv1.PolicyServer) *appsv1.Deployment {
@@ -344,6 +356,20 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
 	templateAnnotations := policyServer.Spec.Annotations
 	if templateAnnotations == nil {
 		templateAnnotations = make(map[string]string)
+	}
+	if r.MetricsEnabled {
+		envvar := corev1.EnvVar{Name: constants.PolicyServerEnableMetricsEnvVar, Value: "1"}
+		if index := envVarsContainVariable(admissionContainer.Env, constants.PolicyServerEnableMetricsEnvVar); index >= 0 {
+			admissionContainer.Env[index] = envvar
+		} else {
+			admissionContainer.Env = append(admissionContainer.Env, envvar)
+		}
+		logFmtEnvVar := corev1.EnvVar{Name: constants.PolicyServerLogFmtEnvVar, Value: "otlp"}
+		if index := envVarsContainVariable(admissionContainer.Env, constants.PolicyServerLogFmtEnvVar); index >= 0 {
+			admissionContainer.Env[index] = logFmtEnvVar
+		} else {
+			admissionContainer.Env = append(admissionContainer.Env, logFmtEnvVar)
+		}
 	}
 
 	policyServerDeployment := &appsv1.Deployment{
--- a/internal/pkg/admission/policy-server-deployment_test.go
+++ b/internal/pkg/admission/policy-server-deployment_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	policiesv1 "github.com/kubewarden/kubewarden-controller/apis/policies/v1"
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+		},
+	}
+	deployment := reconciler.deployment("v1", policyServer)
+	hasMetricEnvvar := false
+	hasLogFmtEnvvar := false
+	for _, envvar := range deployment.Spec.Template.Spec.Containers[0].Env {
+		if envvar.Name == "KUBEWARDEN_ENABLE_METRICS" {
+			hasMetricEnvvar = true
+			if envvar.Value != "1" {
+				t.Error("Missing KUBEWARDEN_ENABLE_METRICS value")
```
viccuad, line 489: If the variable is set regardless of value, it is enabled. I would reword to:

```suggestion
				t.Error("Present but not reconciled KUBEWARDEN_ENABLE_METRICS value")
```

(same for LOG_FMT)

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -226,6 +226,19 @@ func (r *Reconciler) adaptDeploymentSettingsForPolicyServer(policyServerDeployme
 			},
 		)
 	}
+	if r.MetricsEnabled {
+		policyServerDeployment.Annotations[constants.OptelInjectAnnotation] = "true"
+	}
+}
+
+func envVarsContaintsVariable(container *corev1.Container, envVarName string) (bool, int) {
```
viccuad, line 234:
```suggestion
func envVarsContainVariable(container *corev1.Container, envVarName string) (bool, int) {
```

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+	for _, envvar := range deployment.Spec.Template.Spec.Containers[0].Env {
+		if envvar.Name == "KUBEWARDEN_ENABLE_METRICS" {
+			hasMetricEnvvar = true
+		}
+		if envvar.Name == "KUBEWARDEN_LOG_FMT" {
+			hasLogFmtEnvvar = true
+		}
+	}
+	if hasMetricEnvvar {
+		t.Error("KUBEWARDEN_ENABLE_METRICS should not define environment variable")
```
viccuad, line 539:
```suggestion
		t.Error("KUBEWARDEN_ENABLE_METRICS should not be set")
```

(same for the other messages)

viccuad (approved):
LGTM! just some nits

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -226,6 +226,19 @@ func (r *Reconciler) adaptDeploymentSettingsForPolicyServer(policyServerDeployme
...
+}
+
+func envVarsContaintVariable(container *corev1.Container, envVarName string) (bool, int) {
+	for i, envvar := range container.Env {
+		if envvar.Name == envVarName {
+			return true, i
+		}
+	}
+	return false, -1
+
```
flavio, line 241: nitpicking...

* The name of the function has a typo. I think you meant to call it `envVarsContainsVariable`
* I don't like the return type of the function, I see two options:
  *  return `int`: it would be `-1` if the item has not been found, otherwise it would return the index position
  * return `(bool, string)`: instead of returning the index to the value, just return the value

I also wonder if it would make sense to make the function more generic by changing the signature to use that as input: `(values []EnvVar, key string)`

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -345,6 +358,23 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
...
 		templateAnnotations = make(map[string]string)
 	}
+	if r.MetricsEnabled {
+		envvar := corev1.EnvVar{Name: constants.PolicyServerEnableMetricsEnvVar, Value: "1"}
+		logFmtEnvVar := corev1.EnvVar{Name: constants.PolicyServerLogFmtEnvVar, Value: "otlp"}
+
+		if hasEvnvar, index := envVarsContaintVariable(&admissionContainer, constants.PolicyServerEnableMetricsEnvVar); hasEvnvar {
+			admissionContainer.Env[index] = envvar
+		} else {
+			admissionContainer.Env = append(admissionContainer.Env, envvar)
```
flavio, line 368: I would move the definition of `envvar` inside of this block

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -345,6 +358,23 @@ func (r *Reconciler) deployment(configMapVersion string, policyServer *policiesv
...
+		}
+
+		if hasEvnvar, index := envVarsContaintVariable(&admissionContainer, constants.PolicyServerLogFmtEnvVar); hasEvnvar {
+			admissionContainer.Env[index] = logFmtEnvVar
+		} else {
+			admissionContainer.Env = append(admissionContainer.Env, logFmtEnvVar)
```
flavio, line 374: I would move the definition of `logFmtEnvVar` into this block

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+		Spec: policiesv1.PolicyServerSpec{
+			Image: "image",
+			Env:   []corev1.EnvVar{{Name: "KUBEWARDEN_ENABLE_METRICS", Value: "0"}, {Name: "KUBEWARDEN_LOG_FMT", Value: "invalid"}},
+		},
+	}
+	deployment := reconciler.deployment("v1", policyServer)
+	hasMetricEnvvar := false
+	hasLogFmtEnvvar := false
+	for _, envvar := range deployment.Spec.Template.Spec.Containers[0].Env {
+		if envvar.Name == "KUBEWARDEN_ENABLE_METRICS" {
```
flavio, line 486: nitpicking, I think we have a constant for that

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+			hasMetricEnvvar = true
+			if envvar.Value != "1" {
+				t.Error("Present but not reconciled KUBEWARDEN_ENABLE_METRICS value")
+			}
+		}
+		if envvar.Name == "KUBEWARDEN_LOG_FMT" {
```
flavio, line 492: nitpicking, I think we have a constant for that

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+	if !hasMetricEnvvar {
+		t.Error("Missing KUBEWARDEN_ENABLE_METRICS environment variable")
+	}
+	if !hasLogFmtEnvvar {
+		t.Error("Missing KUBEWARDEN_LOG_FMT environment variable")
+	}
+
+	value, hasAnnotation := deployment.Annotations["sidecar.opentelemetry.io/inject"]
+	if !hasAnnotation {
+		t.Error("Missing OPTEL annotation")
```
flavio, line 508: I think OPTEL is a typo, I think it should be OTEL

internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -432,3 +432,118 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
...
+	}
+	if value != "true" {
+		t.Error("OPTEL annotation invalid value")
```
flavio, line 511: I think OPTEL is a typo

flavio (changes requested):

internal/pkg/admission/policy-server-deployment.go
jvanz: What do you mean? Duplicate the `envvar` definition on line 366 and 368?

internal/pkg/admission/policy-server-deployment.go
flavio: You're right, `contain` is correct. However the name currently is "envVarsContain**t**Variable", note the highlighted "t". It should be `envVarsContainVariable`

internal/pkg/admission/policy-server-deployment.go
flavio: Scratch it... I don't know what I saw :facepalm:

internal/pkg/admission/policy-server-deployment.go
flavio: Same here... sorry about the noise

flavio (approved):
LGTM, there's just the typo in the function name to be fixed

PR #446, opened 2023-05-05
fix: otel annotations should be in the pod spec.

jvanz:
## Description

The controller was adding the OTEL annotation to inject the collector container in the deploy annotations but it should be added in the pod spec inside the deployment spec. This commit fixes that and the tests to validate this behavior.


viccuad (approved):
LGTM!

flavio (approved):
Nice catch!

PR #449, opened 2023-05-10
fix: missing metrics port in the policy server service.

jvanz:
## Description

In the latest changes in the controller we remove the need to define the OTEL environment variables in the policy server definition. This causes a problem because the controller uses this information to detect if it should adds the metrics port in the policy server service.

This commit fixes that by checking the flag in the controller to know if metrics are enabled or not.

Related to https://github.com/kubewarden/kubewarden-controller/pull/442


flavio (approved):
LGTM

viccuad (approved):
LGTM

viccuad:
rebased as needed by github, and merging, to get the image.

Issue #411, opened 2023-03-03
Update to cosign v2


viccuad:
[Cosign v2 was released](https://blog.sigstore.dev/cosign-2-0-released/), and, as expected by semver, is not backwards compatible with v1.

## Acceptance criteria

- [x] Make sure all GitHub Actions from kubewarden/github-actions are consuming cosign v2, so, using [GHA cosign-installer 3.*](https://github.com/sigstore/cosign-installer/releases/tag/v3.0.0), and they are correctly signing images.
- [x] https://github.com/kubewarden/docs/pull/175
- [x] Make sure all repos that don't come from kubewarden/github-actions (e.g: GHA in kubewarden-controller, kwctl, audit-scanner, rancher-kubectl-builder, etc) are using GHA cosign-installer 3.*, and they are correctly signing things.

flavio:
Starting from this week, there's even a v3 installer out. I think we should update during the 1.7.0 RC phase

jvanz:
Everything is using the latest cosign and cosign installer action version. Closing the issue

PR #477, opened 2023-06-30
fix: updates cosign to v2.

jvanz:
## Description

Updates the CI scripts updating the cosign-installer to v3. Thus, the cosign version now is v2. Which has breaking changes requiring some changes on how the command is called and removing the need of using environment variable to enable keyless signature.

Related to https://github.com/kubewarden/kubewarden-controller/issues/411


viccuad (approved):
LGTM!

PR #485, opened 2023-07-20
fix: GH action, reuse actions from the same checkout

flavio:
Prior to this commit, some GH actions reused workflows that came from the `main` branch, regardless of the branch (or tag) triggering the execution.

This can be a bit dangerous and have side effects. This commit ensures everything stays in sync.


jvanz (approved):
Thanks for that! That will also simplify when we need to test CI on forks

PR #487, opened 2023-07-25
New SBOM generation process

flavio:
Use syft to produce bill or materials of the container image contents and the application itself.

Ensure each architecture-specific build has its own SBOM, signed and attached to the specific image.

Also, attach to the GH release the SBOM of each container image being built.

## GH actions

A preview of the results produced by the this new approach can be found here:

* Push against `main`: [GH run](https://github.com/flavio/kubewarden-controller/actions/runs/5654765398). The run now has a `sbom` artifact, this contains the spdx files and all their signatures
* New tag:
  - [GH run](https://github.com/flavio/kubewarden-controller/actions/runs/5655027453). The run failed because it couldn't perform the last release step: triggering the update of helm charts; this fine, since the tag happened inside of a fork
  - [GH release](https://github.com/flavio/kubewarden-controller/releases/tag/v1.7.1). We now have assets for both the arm64 and amd64 builds

## Container images

Prior to this commit the SBOM was only generated for the x86_64 platform and was attached to the multi-architecture container image index manifest.

Prior to this change the SBOM was attached to the image index manifest:

```console
cosign download sbom --output-file sbom.spdx ghcr.io/kubewarden/kubewarden-controller:v1.6.0
WARNING: Downloading SBOMs this way does not ensure its authenticity. If you want to ensure a tamper-proof SBOM, download it using 'cosign download attestation <image uri>' or verify its signature using 'cosign verify --key <key path> --attachment sbom <image uri>'.
Found SBOM of media type: text/spdx
```

After this PR the SBOM is no longer attached to the image index:

```console
WARNING: Downloading SBOMs this way does not ensure its authenticity. If you want to ensure a tamper-proof SBOM, download it using 'cosign download attestation <image uri>' or verify its signature using 'cosign verify --key <key path> --attachment sbom <image uri>'.

This multiarch image does not have an SBOM attached at the index level.
Try using --platform with one of the following architectures:
linux/amd64, linux/arm64, unknown/unknown, unknown/unknown

Error: no SBOM found attached to image index
main.go:74: error during command execution: no SBOM found attached to image index
```

Instead, they are attached to the actual container images.

This is the amd64 SBOM:

```console
cosign download sbom --output-file amd64-sbom.spdx ghcr.io/flavio/kubewarden-controller@$(crane digest --platform linux/amd64 ghcr.io/flavio/kubewarden-controller:v1.7.1)
WARNING: Downloading SBOMs this way does not ensure its authenticity. If you want to ensure a tamper-proof SBOM, download it using 'cosign download attestation <image uri>' or verify its signature using 'cosign verify --key <key path> --attachment sbom <image uri>'.
Found SBOM of media type: text/spdx
```

While this one is the arm64 one:

```console
cosign download sbom --output-file arm64-sbom.spdx ghcr.io/flavio/kubewarden-controller@$(crane digest --platform linux/arm64 ghcr.io/flavio/kubewarden-controller:v1.7.1)
WARNING: Downloading SBOMs this way does not ensure its authenticity. If you want to ensure a tamper-proof SBOM, download it using 'cosign download attestation <image uri>' or verify its signature using 'cosign verify --key <key path> --attachment sbom <image uri>'.
Found SBOM of media type: text/spdx
```

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -6,6 +6,7 @@ on:
       digest:
         description: "Container image digest"
         value: ${{jobs.build.outputs.digest}}
+
   push:
     branches:
       - "main"
@@ -20,29 +21,12 @@ jobs:
     with:
       push-image: true
 
-  sign:
-    runs-on: ubuntu-latest
+  sbom:
+    name: SBOM
+    needs: build
+    uses: ./.github/workflows/sbom.yml
     permissions:
       packages: write
       id-token: write
-    needs: build
-    steps:
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v2
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - uses: sigstore/cosign-installer@v3
-
-      - name: Sign the images
-        run: |
-          cosign sign --yes \
-            ${{needs.build.outputs.repository}}@${{needs.build.outputs.digest}}
-
-      - name: Sign the SBOM
-        run: |
-          tag=$(echo '${{needs.build.outputs.digest}}' | sed 's/:/-/g')
-          cosign sign --yes \
-            "${{needs.build.outputs.repository}}:$tag.sbom"
+    with:
+      image-digest: ${{ needs.build.outputs.digest }}
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,4 +1,4 @@
-name: Build container image every change.
+name: Build container image and SBOMs
 
 on:
   workflow_call:
@@ -14,15 +14,16 @@ on:
 
 jobs:
   build:
-    name: Build
+    name: Build container image
     uses: ./.github/workflows/container-image.yml
     permissions:
       packages: write
+      id-token: write
     with:
       push-image: true
 
   sbom:
-    name: SBOM
+    name: Generate SBOM and push them to OCI registry
     needs: build
     uses: ./.github/workflows/sbom.yml
     permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,32 +12,21 @@ jobs:
     uses: ./.github/workflows/ci.yml
     permissions: read-all
 
-  container-build:
+  build:
+    name: Build container
     uses: ./.github/workflows/container-build.yml
     permissions:
       id-token: write
       packages: write
-  release:
-    permissions:
-      id-token: write
-      contents: write
-    name: Create release
+
+  crds:
+    name: Build CRDs
     runs-on: ubuntu-latest
-    needs:
-      - ci
-      - container-build
     steps:
       - name: Install Golang
         uses: actions/setup-go@v4
         with:
           go-version: '1.19'
-
-      - name: Install the bom command
-        shell: bash
-        run: go install sigs.k8s.io/bom/cmd/bom@v0.2.2
-
-      - name: Install cosign
-        uses: sigstore/cosign-installer@v3
 
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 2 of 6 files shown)


flavio:
@viccuad iterating over the controller was faster than doing that against policy-server. Once merged I'll propagate these changes made to the GH workflows to the new build proposal I created against the policy-server repo

.github/workflows/container-build.yml
```diff
@@ -6,6 +6,7 @@ on:
       digest:
         description: "Container image digest"
         value: ${{jobs.build.outputs.digest}}
+
```
viccuad, line 9: (opening here, because GH doesn't allow to comment in non-changed line)

Nit: this workflow  name could be better `name: Build container image and generate SBOMs on every change` for clarity

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,77 @@
+name: Build container image
```
viccuad, line 1:
```suggestion
name: Generate SBOMs
```

(or similar)

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,77 @@
...
+            ghcr.io/${{ github.repository_owner }}/kubewarden-controller@${{ env.PLATFORM_DIGEST }}
+
+      - name: Sign BOM file
+        run: |
+          cosign sign-blob --yes \
+            --output-certificate kubewarden-controller-sbom-${{ matrix.arch }}.spdx.cert \
+            --output-signature kubewarden-controller-sbom-${{ matrix.arch }}.spdx.sig \
+            kubewarden-controller-sbom-${{ matrix.arch }}.spdx
+
+      - name: Attach SBOM file in the container image
```
viccuad, line 65: out of curiosity:
the sboms attached to the container images are not signed, and cosign suggest to use a specific command to download the attestation or verify the signature. I suppose that doesn't work as cosign doesn't know how to do that, yet, no?

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,77 @@
...
+      packages: write
+      id-token: write
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3
+
+      - name: Install the syft command
+        #TODO: change origin
```
viccuad, line 27: (we shouldn't forget these)

viccuad (approved):
Looks good, I like it! Quite an improvement!

.github/workflows/release.yml
```diff
@@ -100,13 +94,33 @@ jobs:
...
             let fs = require('fs');
-            let files = ['kubewarden-controller-sbom.spdx', 'kubewarden-controller-sbom.spdx.cert', 'kubewarden-controller-sbom.spdx.sig', "CRDS.tar.gz"]
+            let files = [
+              'kubewarden-controller-sbom-amd64.spdx',
+              'kubewarden-controller-sbom-amd64.spdx.cert',
+              'kubewarden-controller-sbom-amd64.spdx.sig',
+              'kubewarden-controller-sbom-arm64.spdx',
+              'kubewarden-controller-sbom-arm64.spdx.cert',
+              'kubewarden-controller-sbom-arm64.spdx.sig',
+              "CRDS.tar.gz"]
```
viccuad, line 123: (opening thread here randomly)

Unrelated to this PR, but,
What happens if the release fails on triggering the helm-charts repo?
We can retrigger the job, without rebuilding and overwriting the release, correct?
I wouldn't like to release something and later overwrite the same release version with different artifacts.

flavio: We could try to rerun only the failed job, that's an option. Otherwise I fear we will have to either manually create the PR that updates the helm chart or tag a new patch release

viccuad: I'm happy just re-triggering the job that is in your fork as it is, to see what happens.

flavio:
@viccuad please take a look at the last commit I just pushed. It addresses your feedback, plus it brings back image signing

viccuad (approved):
LGTM, thanks for the changes!

PR #488, opened 2023-07-25
test: enable dependabot

flavio:
Configure dependabot to keep GH actions updated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  open-pull-requests-limit: 10


viccuad (approved):

viccuad:
Wait, renovate does detect github-actions, see https://github.com/kubewarden/kubewarden-controller/issues/159

viccuad (changes requested):
Wait, renovate does detect github-actions, see: https://github.com/kubewarden/kubewarden-controller/issues/159

I think we just need to trigger it there in the dashboard. It would have happened weekly anyways.

flavio:
Could it be that dependabot is the one that handles the nice `shasum # v1.2` syntax while renovate doesn't?

viccuad:
That's it! For example, on sigstore-rs, we have the nice `shasum # v1.2` and it only has dependabot enabled.

I'm in for dealing with github-actions with dependabot, and disable renovatebot for them. For example with https://docs.renovatebot.com/configuration-options/#ignoredeps.

viccuad:
Renovatebot seems to do the same, but maybe one needs to do it once for it to keep with it:
https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

flavio:
Nice finding! Then we could configure renovatebot to use the nice syntax it provides

flavio:
Closing in favor of https://github.com/kubewarden/kubewarden-controller/pull/489

PR #491, opened 2023-07-26
ci: Fix e2e tests, no need for `id-token` write perms for them

viccuad:
## Description

This removes the `id-token` permission from container-image.yml, by separating the image signing into sign-image.yml.

This makes container-image.yml reusable as a workflow. If not, one gets:
```
The nested job 'build' is requesting 'id-token: write', but is only allowed 'id-token: none'
```

This happens on the e2e-tests.yml workflow without this change.

## Test

To be tested on my CI fork run. Notice the separation of the container signing into its own job.
Run of build image:
https://github.com/viccuad/kubewarden-controller/actions/runs/5667913267

Run of end-to-end-tests, without the error now:
https://github.com/viccuad/kubewarden-controller/actions/runs/5667954405
(they fail on the e2e tests proper, but not on the build image)

No changes to release job.

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,4 +1,4 @@
-name: Build container image and SBOMs
+name: Build container image, sign it, and generate SBOMs
 
 on:
   workflow_call:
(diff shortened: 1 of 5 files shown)


flavio (approved):
Nice refactor, LGTM

viccuad:
Fixed some mistakes, now it is ready for review.

flavio (approved):

viccuad:
fixed DCO on last commit, was a renamed fixup! commit..

PR #502, opened 2023-08-09
feat: CA controller

### LINKED ISSUE #493: Move generation of CA to the main reconciliation loop
## Is your feature request related to a problem?

Currently the kubewarden-controller generates certificates for each Policy Server instance declared by the user. All these certificates are signed by a self-signed CA.

The CA data is stored inside of a Secret which is generated the 1st time a `PolicyServer` resource is created.

The audit-scanner needs to make https requests against the Policy Servers defined inside of the cluster. Because  of that, we mount the certificate of this CA into the audit scanner Cron Jobs. In this way we can ensure a secure communication between the audit-scanner and the Policy Servers.

### The problem

The audit-scanner Cron Job is currently created by the `kubewarden-controller` helm chart; this happens as soon as the helm chart is installed. However, in the beginning there's no Policy Server defined, hence there's no root CA yet, hence there's no Secret containing the CA information. Kubernetes will immediately schedule the execution of the Cron Job, which will fail for as long as the 1st Policy Server is created.

This leaves a better impression to the end users, they might think their installation is broken.

## Solution you'd like

We should move the generation of the CA outside of the reconciliation loop that monitors Policy Server resources. This should be done inside of the main control loop.

By doing that we will ensure the CA secret is created almost immediately.

This is not enough to solve the issue, but it's a starting point. I'll file another issue to keep track of what has to be done.

## Alternatives you've considered

We have thought about other solutions, but unfortunately there's no other way to go. We could generate the CA via helm, but we don't want to go this way. We prefer to keep full control over the CA, particularly about how often this is automatically renewed.

### Changing the name and purpose

While moving the code, we would like to rename the CA. Right now the CA is explicitly about "Policy Server", both in terms of the name of the secret and its x509 details.

We would like to make this CA more "generic", make it just about Kubewarden (the project). This would allow us to leverage this CA later on, when getting rid of cert-manager. We could in fact use this CA to sign all the certificates used by the Kubewarden stack:
* Policy Server (already done)
* Kubewarden Controller Webhook (to be done - currently handled via cert-manager)

### Relation with 1.7.0 release candidates

We are about to tag 1.7.0-rc2. We don't want this change to be part of RC2 because it's too invasive. RC2 will cope with the limitations described at the top of the issue.

This will be part of RC3

### 

TODO:
- [x] https://github.com/kubewarden/kubewarden-controller/issues/547
- [ ] https://github.com/kubewarden/kubewarden-controller/issues/544
- [ ] https://github.com/kubewarden/kubewarden-controller/issues/545
- [ ] https://github.com/kubewarden/kubewarden-controller/issues/546
- [ ] https://github.com/kubewarden/docs/issues/252
- [ ] https://github.com/kubewarden/helm-charts/issues/321

Nice to have:

- [x] https://github.com/kubewarden/kubewarden-controller/issues/548

jvanz:
## Description
    
Adds new controller to reconcile Kubewarden root CA and updates the reconciler to proper update the policy server CA when necessary.

Fix #493

### DIFF
--- a/controllers/ca_controller.go
+++ b/controllers/ca_controller.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/admission"
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/admissionregistration"
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// This controller reconcile the root ca used by the Kubewarden stack
+type KubewardenCAReconciler struct {
+	client.Client
+	Log        logr.Logger
+	Scheme     *runtime.Scheme
+	Reconciler admission.Reconciler
+}
+
+func (r *KubewardenCAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	CASecret, initialized, err := r.Reconciler.FetchOrInitializeRootCASecret(ctx, admissionregistration.GenerateCA, admissionregistration.PemEncodeCertificate)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to create root CA")
+	}
+	// if we have a new kubewarden root ca it is necessary to update the policy server secrets.
+	if initialized {
+		r.Log.Info("Root CA initialized")
+		err := r.Reconciler.UpdateAllPolicyServerSecrets(ctx, CASecret)
+		if err != nil {
+			return ctrl.Result{}, errors.Wrap(err, "failed to recreate policy servers secrets")
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *KubewardenCAReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Secret{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			// CA controller watches only root ca secret.
+			// policy server secret should be managed by the policy server controller
+			secret, ok := object.(*corev1.Secret)
+			if !ok {
+				r.Log.Info("object is not type of corev1.Secret: %+v", secret)
+				return []ctrl.Request{}
+			}
+			if secret.Name == constants.KubewardenCARootSecretName {
+				return []ctrl.Request{
+					{
+						NamespacedName: client.ObjectKey{
+							Name: secret.Name,
+						},
+					},
+				}
+			}
+			return []ctrl.Request{}
+		})).Complete(r)
+	return errors.Wrap(err, "failed enrolling controller with manager")
+}
--- a/controllers/policyserver_controller.go
+++ b/controllers/policyserver_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -60,6 +61,15 @@ type PolicyServerReconciler struct {
 //+kubebuilder:rbac:namespace=kubewarden,groups=core,resources=pods,verbs=get;list;watch
 
 func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_, err := r.Reconciler.FetchKubewardenCARootSecret(ctx)
+	if err != nil {
+		r.Log.Info("cannot reconcile policy server: missing root CA. Rescheduling...")
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: time.Second * 5,
+		}, nil
+	}
+
 	var policyServer policiesv1.PolicyServer
 	if err := r.Get(ctx, req.NamespacedName, &policyServer); err != nil {
 		if apierrors.IsNotFound(err) {
@@ -80,6 +90,12 @@ func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileResult, reconcileErr := r.reconcile(ctx, &policyServer, policies)
 
 	if err := r.Client.Status().Update(ctx, &policyServer); err != nil {
+		if apierrors.IsConflict(err) {
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second * 5,
+			}, nil
+		}
 		return ctrl.Result{}, fmt.Errorf("update policy server status error: %w", err)
 	}
 
@@ -187,6 +203,28 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				{
 					NamespacedName: client.ObjectKey{
 						Name: policy.Spec.PolicyServer,
+					},
+				},
+			}
+		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			// This watch ensures that if the policy server CA
+			// secret is deleted, it will be recreated.
+			secret, ok := object.(*corev1.Secret)
+			if !ok {
+				r.Log.Info("object is not type of corev1.Secret: %+v", secret)
+				return []ctrl.Request{}
+			}
+			// the controller just care about policy server secrets. It ignore any other secret.
+			policyServerName, ok := secret.Labels[constants.PolicyServerLabelKey]
+			if !ok {
+				return []ctrl.Request{}
+			}
+
+			return []ctrl.Request{
+				{
+					NamespacedName: client.ObjectKey{
+						Name: policyServerName,
 					},
 				},
 			}
--- a/internal/pkg/admission/policy-server-ca-secret.go
+++ b/internal/pkg/admission/policy-server-ca-secret.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"fmt"
 
+	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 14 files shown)


controllers/policyserver_controller.go
```diff
@@ -191,6 +207,28 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			}
 		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			// This watch ensures that if the policy server CA
```
viccuad, line 211: If this is the watch for PolicyServer CA secrets, are we missing a watch for the Kubewarden CA secret?
Could it be that without that second watch we only create the Kubewarden CA secret on controller initalization? If so, is that enough?

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -19,30 +20,33 @@ type generateCAFunc = func() (*admissionregistration.CA, error)
...
+	ctx context.Context,
+	policyServer *policiesv1.PolicyServer,
+) (*corev1.Secret, error) {
 	policyServerSecret := corev1.Secret{}
 	err := r.Client.Get(
 		ctx,
 		client.ObjectKey{
 			Namespace: r.DeploymentsNamespace,
-			Name:      policyServerName},
+			Name:      policyServer.NameWithPrefix()},
```
viccuad, line 48: Same here WRT to `NameWithPrefix()`.

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -53,39 +57,66 @@ func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context,
...
 	servingCert, servingKey, err := generateCert(
 		admissionregCA.CaCert,
-		fmt.Sprintf("%s.%s.svc", policyServerName, r.DeploymentsNamespace),
-		[]string{fmt.Sprintf("%s.%s.svc", policyServerName, r.DeploymentsNamespace)},
+		fmt.Sprintf("%s.%s.svc", policyServer.NameWithPrefix(), r.DeploymentsNamespace),
+		[]string{fmt.Sprintf("%s.%s.svc", policyServer.NameWithPrefix(), r.DeploymentsNamespace)},
 		admissionregCA.CaPrivateKey)
 	if err != nil {
-		return nil, fmt.Errorf("cannot generate policy-server %s certificate: %w", policyServerName, err)
+		return nil, fmt.Errorf("cannot generate policy-server %s certificate: %w", policyServer.NameWithPrefix(), err)
```
viccuad, line 95: same here WRT to `NameWithPrefix()` in that other Sprintf and the Errorf.

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -53,39 +57,66 @@ func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context,
...
 	}
 	secretContents := map[string]string{
 		constants.PolicyServerTLSCert: string(servingCert),
 		constants.PolicyServerTLSKey:  string(servingKey),
 	}
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      policyServerName,
+			Name:      policyServer.NameWithPrefix(),
```
viccuad, line 103: Same here WRT to `NameWithPrefix()`.

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -97,27 +128,59 @@ func extractCaFromSecret(caSecret *corev1.Secret) (*admissionregistration.CA, er
 	return &admissionregistration.CA{CaCert: caCert, CaPrivateKey: caPrivateKey}, nil
 }
 
-func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Context, generateCA generateCAFunc, pemEncodeCertificate pemEncodeCertificateFunc) (*corev1.Secret, error) {
-	policyServerSecret := corev1.Secret{}
+// FetchOrInitializeRootCASecret checks if the root CA secret exists and
+// creates it if it does not.  It returns the secret containing the root CA, a
+// boolean flag to signal if the secret has been initialize now and an error in
```
viccuad, line 133: I like the bool `initialized` return. It paves the way for refreshing the CA out of the controller if users need.

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -128,16 +191,96 @@ func (r *Reconciler) buildPolicyServerCARootSecret(generateCA generateCAFunc, pe
...
 			Namespace: r.DeploymentsNamespace,
 		},
 		Data: secretContents,
 		Type: corev1.SecretTypeOpaque,
 	}, nil
 }
+
+// ReconcilePolicyServerSecret reconcile the policy server secret when necessary.
+//
+// policyServer is the Policy Server is a reference to the policy server ca secret which to be reconciled
```
viccuad, line 210: I think it this line may mean this, but I'm not sure:

```suggestion
// policyServerSecret is a reference to the policy server ca secret which to be reconciled
```

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -3,7 +3,7 @@ package admission
 import (
 	"bytes"
 	"context"
-	"crypto/rsa"
+	// "crypto/rsa"
```
viccuad, line 6: woops, leftover.

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -55,7 +59,7 @@ func TestFetchOrInitializePolicyServerCARootSecret(t *testing.T) {
 	for _, test := range tests {
 		ttest := test // ensure tt is correctly scoped when used in function literal
 		t.Run(ttest.name, func(t *testing.T) {
-			secret, err := ttest.r.fetchOrInitializePolicyServerCARootSecret(context.Background(), generateCAFunc, pemEncodeCertificateFunc)
+			secret, _, err := ttest.r.FetchOrInitializeRootCASecret(context.Background(), generateCAFunc, pemEncodeCertificateFunc)
```
viccuad, line 62: it would be nice to add a test for the returned `initialized` boolean, as the reconciler depends on it.

internal/pkg/constants/constants.go
```diff
@@ -9,12 +9,12 @@ var (
...
 	// PolicyServer CA Secret
-	PolicyServerTLSCert                  = "policy-server-cert"
-	PolicyServerTLSKey                   = "policy-server-key"
-	PolicyServerCARootSecretName         = "policy-server-root-ca"
-	PolicyServerCARootPemName            = "policy-server-root-ca-pem"
-	PolicyServerCARootCACert             = "policy-server-root-ca-cert"
-	PolicyServerCARootPrivateKeyCertName = "policy-server-root-ca-privatekey-cert"
+	PolicyServerTLSCert                = "policy-server-cert"
+	PolicyServerTLSKey                 = "policy-server-key"
+	KubewardenCARootSecretName         = "kubewarden-root-ca"
```
viccuad, line 14: I like it, better.
For audit-scanner args, this means we need to change the what secret we mount and the name of that secret in the helm-charts .

For helm upgrades, it may mean we need to recreate the policy-servers somehow, so they use the new secret. And we should also delete the now unused `policy-server-root-ca` in a post-install hook.

viccuad (changes requested):
Looking good! I have some small changes and clarifications for my benefit.

controllers/ca_controller.go
```diff
@@ -0,0 +1,84 @@
...
+	client.Client
+	Log        logr.Logger
+	Scheme     *runtime.Scheme
+	Reconciler admission.Reconciler
+}
+
+func (r *KubewardenCAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	CASecret, initialized, err := r.Reconciler.FetchOrInitializeRootCASecret(ctx, admissionregistration.GenerateCA, admissionregistration.PemEncodeCertificate)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to create root CA")
```
flavio, line 47: Technically we are not creating the CA inside of the etcd store yet. A better error message could be: `failed to fetch root CA`

controllers/ca_controller.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	// if we have a new kubewarden root ca it is necessary to update the policy server secrets.
+	if initialized {
+		r.Log.Info("Root CA initialized")
+		err := r.Reconciler.UpdateAllPolicyServerSecrets(ctx, CASecret)
```
flavio, line 52: I don't know if this method is doing that, but this is a list of steps that have to be done when the CA is generated/changed:

* Regenerate all the certificates used by the policy servers
* Update the contents of the Secret used by the policy server: add the new certificate
* Rollout the policy servers to ensure they are restarted using the new certificates
* Update all the validating/mutating webhook configurations generated by KW: there's the old CA bundled in there, they have to consume the new one

controllers/policyserver_controller.go
```diff
@@ -80,6 +90,12 @@ func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileResult, reconcileErr := r.reconcile(ctx, &policyServer, policies)
 
 	if err := r.Client.Status().Update(ctx, &policyServer); err != nil {
+		if apierrors.IsConflict(err) {
```
flavio, line 93: Why can we run into a conflict now? Is that because the policy server could be changed by the new CA reconciliation loop too?

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -19,30 +20,33 @@ type generateCAFunc = func() (*admissionregistration.CA, error)
 type pemEncodeCertificateFunc = func(certificate []byte) ([]byte, error)
 type generateCertFunc = func(ca []byte, commonName string, extraSANs []string, CAPrivateKey *rsa.PrivateKey) ([]byte, []byte, error)
 
+// reconcileCASecret reconcile the given secret.
+// It updates the secret if it already exists. Otherwise, it creates the secret.
```
flavio, line 24: I'm still going through the PR, why should we call this method? Regenerating the CA is pretty expensive, because lots of things have to be recreated and rolled out. This should be done only when the CA expires

internal/pkg/admission/policy-server-ca-secret.go
flavio:
```suggestion
// boolean flag to signal if the secret has been initialized now and an error in
```

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -97,27 +128,59 @@ func extractCaFromSecret(caSecret *corev1.Secret) (*admissionregistration.CA, er
...
-		return &corev1.Secret{},
+		return &corev1.Secret{}, false,
 			fmt.Errorf("cannot fetch or initialize Policy Server CA secret: %w", err)
 	}
+	return &CASecret, false, nil
+}
+
+// FetchKubewardenCARootSecret function used to get the secret storing the root
+// CA secret used by Kubewarden controllers.
+func (r *Reconciler) FetchKubewardenCARootSecret(ctx context.Context) (*corev1.Secret, error) {
```
flavio, line 164: It seems pretty similar to what `FetchOrInitializeRootCASecret`, do we need both methods?

controllers/ca_controller.go
jvanz: Actually it can try to create the root CA. For example, if the secret is deleted. This controller we recreated. Thus, the error could happen during the secret creation. Let me change the error message to be more generic and leave the original error to give more context about the actual errors.

controllers/ca_controller.go
jvanz: This method is regenerating all the policy server secrets. However, it's not necessary to rollout the policy servers. Acordially to the Kubernetes [docs](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) the volume will be update with the latest content of the secret:

> When a volume contains data from a Secret, and that Secret is updated, Kubernetes tracks this and updates the data in the volume, using an eventually-consistent approach

I'll double check the webhooks

jvanz: Ah, but the policy server just read the secret content in the startup. Yeah, a rollout will be necessary unless we find a way to signal policy server to reread the secret. Which is out of scope of this PR for now. I'll ensure that server will be restarted.

controllers/policyserver_controller.go
jvanz: Yes, the CA controller also updates policy server by setting the conditions to the policy server about secret reconciliation

jvanz: Let me share some toughs that I had while I was coding this controller and see what you think. I believe this is pertinent for this discussion.  

In this PR version the policy server controller still takes care of the secret of the policy server. It will recreate it if it's missing. And the CA controller will touch policy server secret only when it creates the root CA. Which leads to too controller updating policy server resource. The more I think about this more I'm in favor to remove completely the policy server CA reconciliation from the policy server controller and move it to the CA controller. 

I've decided to keep policy server CA reconciliation in the policy server controller as well because it sounds nice to have "all policy server stuff" in a single controller. This will not fix the conflicts on updates. But the code may be simpler to maintain if we move all the CA secret reconciliation to the dedicated controller and change the policy server controller to just wait for the secret to show up before continuing the reconciliation. 

What do you think @kubewarden/kubewarden-developers

controllers/policyserver_controller.go
jvanz: The CA controller ensures that the root CA will be there and all the policy server secrets will update when the root CA changed.

This question is also related to my comment [here](https://github.com/kubewarden/kubewarden-controller/pull/502/files#r1290656323)

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -53,39 +57,66 @@ func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context,
...
+func (r *Reconciler) buildPolicyServerCASecret(policyServer *policiesv1.PolicyServer, caSecret *corev1.Secret, generateCert generateCertFunc) (*corev1.Secret, error) {
 	admissionregCA, err := extractCaFromSecret(caSecret)
 	if err != nil {
 		return nil, err
 	}
 	servingCert, servingKey, err := generateCert(
 		admissionregCA.CaCert,
-		fmt.Sprintf("%s.%s.svc", policyServerName, r.DeploymentsNamespace),
-		[]string{fmt.Sprintf("%s.%s.svc", policyServerName, r.DeploymentsNamespace)},
+		fmt.Sprintf("%s.%s.svc", policyServer.NameWithPrefix(), r.DeploymentsNamespace),
```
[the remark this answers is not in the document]
jvanz: Actually the service has the prefix. I'm just checking in the e2e tests cluster and the service for the policy server called `default` is `policy-server-default`. 
```console
k8s get policyserver -n kubewarden
NAME      REPLICAS   IMAGE
default   1          ghcr.io/kubewarden/policy-server:v1.7.0-rc2

k8s get service -n kubewarden
NAME                                    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
kubewarden-controller-metrics-service   ClusterIP   10.43.228.168   <none>        8443/TCP   111s
kubewarden-controller-webhook-service   ClusterIP   10.43.29.61     <none>        443/TCP    111s
policy-server-default                   ClusterIP   10.43.171.244   <none>        8443/TCP   95s

```

Furthermore, old code was passing the name with prefix:

https://github.com/kubewarden/kubewarden-controller/blob/df95a402457b16d58a985b249c739ec6ed15484c/internal/pkg/admission/reconciler.go#L181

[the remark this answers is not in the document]
jvanz: I've did this change because I want to access more info from the policy server and I would like to avoid adding more parameters

controllers/ca_controller.go
flavio: > I'll double check the webhooks

Yesterday I was in a hurry while doing the review. You have to change the contents of [caBundle](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#webhookclientconfig-v1-admissionregistration-k8s-io) to be the new CA

jvanz:
After a talk about the [RFC](https://github.com/kubewarden/rfc/pull/24) I've decided to open another [PR](https://github.com/kubewarden/kubewarden-controller/pull/510) because we change kind a lot on how to root CA will be generated and maintained.  Therefore, I'm closing this one.

PR #504, opened 2023-08-10
feat: support policy-optimizer

flavio:
Ensure the PolicyServer Deployment has a init container running the policy-optimizer job.

This is part of the work required to implement [this RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0017-policy-server-reuse-wasm-optimized-modules.md).

More work has to be done, that's why this isn't a PR made against the `main` branch.

Note: right now the PolicyServer doesn't make any use of persistent storage, this is something that has to be done


flavio:
Closing, we are not going to go after this implementation

PR #510, opened 2023-08-28
feat: create root ca.

jvanz:
## Description

Updates the controller to create the root ca used in all the certificates used by the Kubewarden stack.

Related to https://github.com/kubewarden/kubewarden-controller/issues/493

### DIFF
--- a/controllers/policyserver_controller.go
+++ b/controllers/policyserver_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -80,6 +81,12 @@ func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	reconcileResult, reconcileErr := r.reconcile(ctx, &policyServer, policies)
 
 	if err := r.Client.Status().Update(ctx, &policyServer); err != nil {
+		if apierrors.IsConflict(err) {
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second * 5,
+			}, nil
+		}
 		return ctrl.Result{}, fmt.Errorf("update policy server status error: %w", err)
 	}
 
@@ -190,6 +197,27 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					},
 				},
 			}
+		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			// watches for secret to detect when a policy server
+			// secret change. Therefore, the certificate can be
+			// recreated and the webhooks updated
+			secret, ok := object.(*corev1.Secret)
+			if !ok {
+				r.Log.Info("object is not type of secret: %+v", "secret", secret)
+				return []ctrl.Request{}
+			}
+
+			if policyServerName, isPolicyServerSecret := secret.Labels[constants.PolicyServerLabelKey]; isPolicyServerSecret {
+				return []ctrl.Request{
+					{
+						NamespacedName: client.ObjectKey{
+							Name: policyServerName,
+						},
+					},
+				}
+			}
+			return []ctrl.Request{}
 		})).
 		Complete(r)
 
--- a/internal/pkg/admission/policy-server-ca-secret.go
+++ b/internal/pkg/admission/policy-server-ca-secret.go
@@ -17,7 +17,7 @@ import (
 
 type generateCAFunc = func() (*admissionregistration.CA, error)
 type pemEncodeCertificateFunc = func(certificate []byte) ([]byte, error)
-type generateCertFunc = func(ca []byte, commonName string, extraSANs []string, CAPrivateKey *rsa.PrivateKey) ([]byte, []byte, error)
+type generateCertFunc = func(ca []byte, extraSANs []string, CAPrivateKey *rsa.PrivateKey) ([]byte, []byte, error)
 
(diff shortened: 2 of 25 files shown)


main.go
```diff
@@ -350,3 +389,137 @@ func webhooks() []webhookwrapper.WebhookRegistrator {
...
+// available by the controller. This is necessary to properly generate the
+// certificate
+//
+// deploymentsNamespace is the namespace where the controller is deployed
+//
+// validatingWebhooks are the ValidatingWebhookConfiguration names which should
+// be updated when the root CA is initialized
+//
+// mutatingWebhooks are the MutatingWebhookConfiguration names which should
+// be updated when the root CA is initialized
```
flavio, line 420: Aren't these names hard coded by us?

main.go
```diff
@@ -350,3 +389,137 @@ func webhooks() []webhookwrapper.WebhookRegistrator {
...
+			return []byte{}, []byte{}, false, fmt.Errorf("failed to reconcile root CA certificate secret: %s", err.Error())
+		}
+	}
+
+	certificate, privateKey, err := admission.ExtractCertificateData(caSecret)
+	if err != nil {
+		return []byte{}, []byte{}, false, fmt.Errorf("failed to initialize root CA certificate")
+	}
+
+	controllerSecret, initialized, err := admission.FetchOrInitializeCertificate(ctx, clusterClient, controllerWebhookService, deploymentsNamespace, constants.ControllerCertificateSecretName, caSecret, admissionregistration.GenerateCert)
```
flavio, line 440: I think the code below this line should be moved to another function. The name of the this function is `setupCA`, hence it should only ensure the root CA is properly created.

main.go
```diff
@@ -350,3 +389,137 @@ func webhooks() []webhookwrapper.WebhookRegistrator {
...
+		}
+	}
+	return certificate, privateKey, (caInitialized || initialized), nil
+}
+
+func setupCAWebhooksInDevelopmentMode(ctx context.Context, clusterClient client.Client, certificate []byte, validatingWebhooks []string, mutatingWebhooks []string) error {
+	// If the controller is NOT running in development mode, the
+	// webhooks are created by Helm chart and missing the
+	// `caBundle` field with the root CA certificate. Therefore,
+	// the controller needs to patch these resources
```
flavio, line 462: That means that when the controller is deployed via helm our helm charts take care of creating the webhook configurations? Can't we have the controller always create them? I mean, we already have the code somewhere to make the development environment work... This would make everything more consistent

flavio: Moreover, the name of this function is confusing, it should be `setupCAWebhooksInProductionMode`, but again, I hope we can get rid of this code and have the webhook configurations managed by the controller in any kind of deployment environment

main_test.go
flavio, line 1: Having a `main_test` is not good, please move all the functions being tested to a different file and then rename this file accordingly.

flavio (changes requested):
This is quite a big change, which makes the review harder (not your fault, this must touch several places of the codebase).

I think we should not get this change into the 1.7.0 release, this is a too big change.

controllers/policyserver_controller.go
```diff
@@ -191,6 +198,27 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			}
 		})).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			// watches for secret to detect when a policy server
+			// secret change. Therefore, the certificate can be
+			// recreated and the webhooks updated
```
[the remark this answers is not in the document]
jvanz: Yes, this is what is done in the controller. The comment is wrong. I'm still walkingthrouhg your comments. But this is the goal. The [webhooks](https://github.com/kubewarden/kubewarden-controller/blob/da552cf41337c040e09d331659f4c3386cb49e1c/controllers/policy_utils.go#L152-L165) always use the root CA  to validate the server certificate and the policy server deployment will rollout a new version when the [secret version change](https://github.com/kubewarden/kubewarden-controller/blob/da552cf41337c040e09d331659f4c3386cb49e1c/internal/pkg/admission/policy-server-deployment.go#L407) (indicating that policy server certificate change)

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -28,64 +28,48 @@ func (r *Reconciler) reconcileCASecret(ctx context.Context, secret *corev1.Secre
...
-		ctx,
-		client.ObjectKey{
-			Namespace: r.DeploymentsNamespace,
-			Name:      policyServerName},
-		&policyServerSecret)
-	if err != nil && apierrors.IsNotFound(err) {
-		secret, err := r.buildPolicyServerCASecret(policyServerName, caSecret, generateCert)
-		if err != nil {
-			return secret, fmt.Errorf("cannot fetch or initialize Policy Server CA secret: %w", err)
+func ReconcileSecret(ctx context.Context, clusterClient client.Client, secret *corev1.Secret) error {
```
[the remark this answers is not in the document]
jvanz: Actually, it's used [here](https://github.com/kubewarden/kubewarden-controller/blob/da552cf41337c040e09d331659f4c3386cb49e1c/internal/pkg/admission/reconciler.go#L182). But maybe we can merge both functions in to one. I'll double check that out

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -28,64 +28,48 @@ func (r *Reconciler) reconcileCASecret(ctx context.Context, secret *corev1.Secre
...
-func (r *Reconciler) buildPolicyServerCASecret(policyServerName string, caSecret *corev1.Secret, generateCert generateCertFunc) (*corev1.Secret, error) {
-	admissionregCA, err := extractCaFromSecret(caSecret)
-	if err != nil {
-		return nil, err
+func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context, policyServerName string, policyServerServiceName string, caSecret *corev1.Secret, generateCert generateCertFunc) (*corev1.Secret, error) {
+	secret, initialized, err := FetchOrInitializeCertificate(ctx, r.Client, policyServerServiceName, r.DeploymentsNamespace, policyServerServiceName, caSecret, generateCert)
+	if initialized {
+		// label used to detect when the policy server certificate
+		// change and triggering the webhook's caBundle updates
+		secret.Labels[constants.PolicyServerLabelKey] = policyServerName
```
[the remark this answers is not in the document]
jvanz: This label is not used to rollout webhooks. It's used to rollout a policy server deployment.

PR #522, opened 2023-09-13
Disable schedule tests on forks

kravciak:
This should limit problems with parallel tests on nightly job.


flavio (approved):
LGTM, it could also be dangerous to run e2e tests from forks

fabriziosestito (approved):

jvanz (approved):

Issue #443, opened 2023-05-04
Deleting policyServer.imagePullSecret puts kubewarden resource versions out of sync


kravciak:
### Current Behavior

Problem was found in lines 21 & 22 https://github.com/kubewarden/kubewarden-end-to-end-tests/blob/f89beb63db59574aba670148c688474edf887998/tests/private-registry-tests.bats#L20

When I
 - set up defaults with secret `helm update kubewarden-defaults --set policyServer.imagePullSecret=secret-registry-docker`
- deploy policy
- delete secret
- delete policy

Then resource versions go out of sync
```bash
~ kubectl get deploy -n kubewarden policy-server-default -o json
"kubewarden/config-version": "2275"

~ kubectl get cm -n kubewarden policy-server-default -o json | jq '.metadata.resourceVersion'
2459
```

All new policies stay in pending at this point because of ConfigurationVersionMismatch defined here - https://github.com/kubewarden/kubewarden-controller/blob/1c8afcce20f11f1de5056b8adcda53b51e455b3d/controllers/policystatus_utils.go#L49

It does not seem right it breaks this way, but maybe it's a feature - pls confirm...

### To reproduce:
- comment out `teardown_file` function from private-registry-tests.bats, then
```bash
make clean cluster install private-registry-tests.bats
kubectl --namespace kubewarden delete secret secret-registry-docker
kubectl delete clusteradmissionpolicies private-pod-privileged

# check
kubectl get deploy -n kubewarden policy-server-default -o json
kubectl get cm -n kubewarden policy-server-default -o json | jq '.metadata.resourceVersion'
```

fabriziosestito:
Controller logs after deleting the secret:
```console
2023-09-19T09:44:03+02:00       ERROR   Reconciler error        {"controller": "policyserver", "controllerGroup": "policies.kubewarden.io", "controllerKind": "PolicyServer", "PolicyServer": {"name":"default"}, "namespace": "", "name": "default", "reconcileID": "2ee43ac8-5aa1-4421-875b-c91fbb052df5", "error": "reconciliation error: cannot get spec.ImagePullSecret: secrets \"secret-registry-docker\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /home/lain/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329
[...]

PR #530, opened 2023-09-21
feat: validate PolicyServer and imagePullSecret

fabriziosestito:
## Description

This PR adds validation to the PolicyServer CRD admission webhook.
It validates that the PolicyServer name is valid (< 63 characters), and
if an ImagePullSecret was specified it validates the following:
- the resource exists
- the resource is in fact a Secret
- the Secret is of type `dockerconfig/json`

To achieve this, a custom webhook validator was created. 
It wasn't possible to implement the `admission.Validator` interface in the CRD directly,
since we need to pass a client to perform the validation against the Secret.

The `ImagePullSecret` validation was refactored in a new internal package,
so it can be consumed by the controller reconciler loop.
Prior this PR, we were returning error in the reconciler loop causing the reconciliation to go in an infinite loop.
With these changes, the controller does not return error and logs instead, fixing #443.

## Test

Added validation and webhook tests were needed.

## Additional Information

Since we are using https://github.com/ereslibre/kube-webhook-wrapper/, the webhook setup needed to create a function that returns the setup function. See: 
This will be removed in the future if we move away from `kube-webhook-wrapper` with the cert-manager removal (see: ) or if we decide to refactior/merge `kube-werbhook-wrapper` in the current codebase.


pkg/apis/policies/v1/policyserver_webhook.go
```diff
@@ -29,24 +34,73 @@ import (
...
+func (v *policyServerValidator) validate(ctx context.Context, obj runtime.Object) error {
+	policyServer, ok := obj.(*PolicyServer)
+	if !ok {
+		return fmt.Errorf("expected a PolicyServer object, got %T", obj)
+	}
+
+	// The PolicyServer name must be maximum 63 like all Kubernetes objects to fit in a DNS subdomain name
+	if len(policyServer.GetName()) > validationutils.DNS1035LabelMaxLength {
+		return fmt.Errorf("spec.ImagePullSecret cannot be longer than %d characters", validationutils.DNS1035LabelMaxLength)
+	}
```
jvanz, line 84: :clap:

pkg/apis/policies/v1/webhook_suite_test.go
```diff
@@ -581,11 +582,21 @@ var _ = Describe("validate PolicyServer webhook with ", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		By("checking default values")
+		fmt.Fprint(GinkgoWriter, pol.ObjectMeta.Finalizers)
```
jvanz, line 585: I think this is a debug statement that should be removed. Isn't it? :)

jvanz (changes requested):

pkg/apis/policies/v1/webhook_suite_test.go
fabriziosestito: woops! thanks

config/webhook/manifests.yaml
flavio, line 1: Reminder: we have to propagate this change into the kubewarden-controller helm chart

internal/pkg/policyserver/validation.go
```diff
@@ -0,0 +1,43 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ValidateImagePullSecret validates that the specified PolicyServer imagePullSecret exists and is of type kubernetes.io/dockerconfigjson
+func ValidateImagePullSecret(ctx context.Context, k8sClient client.Client, imagePullSecret string, deploymentsNamespace string) error {
+	// By using Unstructured data we force the client to fetch fresh, uncached
+	// data from the API server
```
flavio, line 17: I think that in this case it's fine to go through the go client cache

pkg/apis/policies/v1/policyserver_webhook.go
```diff
@@ -29,17 +34,25 @@ import (
...
+			return fmt.Errorf("failed enrolling webhook with manager: %w", err)
+		}
+
+		return nil
 	}
-	return nil
 }
 
-//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-policyserver,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=policyservers,verbs=create;update,versions=v1,name=mpolicyserver.kb.io,admissionReviewVersions={v1,v1beta1}
+// +kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-policyserver,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=policyservers,verbs=create;update,versions=v1,name=mpolicyserver.kb.io,admissionReviewVersions={v1,v1beta1}
```
flavio, line 55: This was changed by the code formatted, I hope that doesn't break the code generator.
I think this didn't happen, I guess the change to the `manifest.yaml` file inside of this PR has been done by the code generator. Am I right?

pkg/apis/policies/v1/policyserver_webhook.go
```diff
@@ -50,3 +63,44 @@ func (ps *PolicyServer) Default() {
...
+
+func (v *policyServerValidator) validate(ctx context.Context, obj runtime.Object) error {
+	policyServer, ok := obj.(*PolicyServer)
+	if !ok {
+		return fmt.Errorf("expected a PolicyServer object, got %T", obj)
+	}
+
+	// The PolicyServer name must be maximum 63 like all Kubernetes objects to fit in a DNS subdomain name
+	if len(policyServer.GetName()) > validationutils.DNS1035LabelMaxLength {
+		return fmt.Errorf("spec.ImagePullSecret cannot be longer than %d characters", validationutils.DNS1035LabelMaxLength)
```
flavio, line 83: I think the error message is wrong, it's the policy-server name that is longer than 64 chars, not the ImagePullSecret

flavio (commented):
Overall good, I left some minor comments

pkg/apis/policies/v1/policyserver_webhook.go
fabriziosestito: it works either way. We are not very consistent either we use the version with the space in other part of the code alongside the one with the one with no space.

pkg/apis/policies/v1/policyserver_webhook.go
fabriziosestito: thanks!

pkg/apis/policies/v1/policyserver_webhook.go
flavio: Fine with me, I don't have a strong opinion about this formatting. I just want the code generator to work :)

flavio (approved):
LGTM, please create a PR against the helm charts repo to ensure the new webhook is created

config/webhook/manifests.yaml
fabriziosestito: done in: https://github.com/kubewarden/helm-charts/pull/301

jvanz (approved):

PR #541, opened 2023-09-28
chore: remove kube-webhook-wrapper dependency.

jvanz:
## Description

Removes the kube-webhook-wrapper dependency to remove similar code present in the controller and the lib dependency. As well as simplify the next changes where the root CA certificate will be handle by the controller.


fabriziosestito (approved):
LGTM
As agreed,  we can continue with the crypto package duplication that requires further refactoring in the next PR. 
As for now, we decided to include webhook wrapper in this codebase and possibly consider other solutions for the dev environment, see #548.

viccuad (approved):
LGTM.

PR #550, opened 2023-10-06
feat: root ca generation.

### LINKED ISSUE #544: Generate CA and certificate
- Generate Certificate Authority
- Generate certificate for `kubewarden-controller` and sign it with our CA
- Start the Kubewarden controller HTTPS server using this certificate
- Register the webhook against the Kubernetes API, using the certificate of our CA inside of
  `clientConfig.caBundle`
  

Related to: https://github.com/kubewarden/rfc/blob/main/rfc/0018-certificate-handling.md#kubewarden-fresh-install

Maybe this could be helpful: https://github.com/open-policy-agent/cert-controller

### Anything else?

Remove cert manager dead code.

jvanz:
## Description

Adds code to create a root CA, create another certificates signed by the root CA create previously and use this certificate in the webhook server. As well as update the webhooks to use the root CA certificate to validate the webhook server certificates.

Fix #544 

**IMPORTANT NOTES BEFORE REVIEW**

There is an issue in the manager start up due a lack of sync between the certificate creation and the webhook server trying to read it. I'm working on this and will update the PR soon. controller-runtime seems to have a module which can help on that and I'm validating it. But I want to give you the opportunity to read the first version of the code to speed up the review process. 

The package or their names can be rearranged in the future. For example, the `admissionregistration` package could be rewritten soon. 

## Test

You will not be able to run this in host machine. The code that setup the certificate locally in the host has not been updated yet. You need to build the container, load it in a cluster and install Kubewarden using helm charts.


jvanz:
@kubewarden/kubewarden-developers the controller is changed to create the root CA, the certificates for the webhook server and update the webook to use the root CA. 

To prevent some concurrency issues like trying to read the certificate file before its available a custom webhook server has been created to wait for  the job that create/check the certificates. Furthermore,  to prevent the control plane to contact the server before the certificates the ready checker return only after the certificates are ready. 

Keep in mind that routine can change when implementing the certificate rotation. And file structure can be updated as well. 

To properly test this change you can build the container image and test with this [branch](https://github.com/kubewarden/helm-charts/tree/ca-changes) of the helm chart. This is a command that you can use:

```
make docker-build  IMG=ghcr.io/jvanz/kubewarden-controller:root-ca-change && 
pushd . &&
cd <path to the kubewarden e2e tests repository> && 
make clean cluster && 
k3d image import ghcr.io/jvanz/kubewarden-controller:root-ca-change --cluster kubewarden-testing && 
make install KUBEWARDEN_CHARTS_LOCATION=<path to helm chart repository dir with the ca-changes branch code>
popd
```

internal/pkg/crypto/ca.go
```diff
@@ -18,28 +18,28 @@ type CA struct {
 	CaPrivateKey *rsa.PrivateKey
 }
 
-func (ca *CA) PEMEncodeCertificate() (string, error) {
+func (ca *CA) PEMEncodeCertificate() (*bytes.Buffer, error) {
```
viccuad, line 21: I think we should keep returning encoded PEM as strings, given that in it we get base64 payload, ascii compatible, so utf compatible. Hence why they have been encoded.

internal/pkg/constants/constants.go
```diff
@@ -8,6 +8,8 @@ var (
 )
 
 const (
+	RootCASecretName          = "kubewarden-root-ca"
+	ControllerCertsSecretname = "kubewarden-controller-certificate"
```
viccuad, line 12: Silly nitpick with CamelCase:

```suggestion
	ControllerCertsSecretName = "kubewarden-controller-certificate"
```

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+			if err != nil {
+				return err
+			}
+			logger.Info("Start to update webhook with latest CA")
+			err = updateWebhooksCaBundle(ctx, k8s, certificatePEMEncoded.Bytes())
+			if err != nil {
+				return err
+			}
+			logger.Info("Certificates and webhooks updated")
+			os.Exit(constants.ControllerReturnCodeCAInitialized)
```
viccuad, line 67: I would add a comment explaining this;
"We have updated the certs and the webhooks, trigger a new controller container instance that will load the new and correct ones, by closing the current one with os.Exit"
or something similar

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+
+	if err != nil {
+		return fmt.Errorf("unable to create controller certificates")
+	}
+	controllerCertSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.ControllerCertsSecretname,
+			Namespace: deploymentsNamespace,
+		},
+		Immutable:  new(bool),
```
viccuad, line 91: do we need to allocate memory for this bool `false`?

viccuad (commented):
It looks good! :heart: . I have some questions and simple nitpicks before approving.

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	logger = ctrl.Log.WithName("certificates")
+)
+
+// This function will create and store in secrets, when necessary, the root CA and the certificate used in the
+// controller. If the certificate is create, the process exit to force kubernetes to redeploy the pod with the
```
flavio, line 23:
```suggestion
// controller. If the certificate is created, the process exit to force kubernetes to redeploy
```

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+// latest version of the certificates.
+func RenewRootCAIfNedded(ctx context.Context, mgr ctrl.Manager, deploymentsNamespace string, controllerWebhookServiceName string) error {
+	logger.Info("Start to verify certificates")
```
flavio, line 26: This should be a debug statement, plus the message is a bit misleading. We are verifying the root CA, not the certificates

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+	k8s := mgr.GetClient()
+	rootCASecret := &corev1.Secret{}
+	err := k8s.Get(ctx, client.ObjectKey{Namespace: deploymentsNamespace, Name: constants.RootCASecretName}, rootCASecret)
+	if apierrors.IsNotFound(err) {
+		logger.Info("unable to find root CA secret.")
```
flavio, line 31: I would move all this inner block to a dedicated function, like `createRootCA` that would generate the root CA and write that into Kubernetes.

Out of memory, I think we used to have this function

internal/pkg/certificates/certificates.go
flavio: I would add a log statement that explains why the controller is exiting

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+			logger.Info("Certificates and webhooks updated")
+			os.Exit(constants.ControllerReturnCodeCAInitialized)
+			return nil
+		} else {
+			logger.Error(err, "unable to create root CA secret.")
+			return err
+		}
+	}
+	if err == nil {
+		logger.Info("All certificates are valid")
```
flavio, line 75: I think this is misleading. From what I understand this is executed only if the secret that contains the root CA has been found. No verification has been done. I think we can omit this check

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+var (
+	logger = ctrl.Log.WithName("certificates")
+)
+
+// This function will create and store in secrets, when necessary, the root CA and the certificate used in the
+// controller. If the certificate is create, the process exit to force kubernetes to redeploy the pod with the
+// latest version of the certificates.
+func RenewRootCAIfNedded(ctx context.Context, mgr ctrl.Manager, deploymentsNamespace string, controllerWebhookServiceName string) error {
```
flavio, line 25: Having reviewed the code, this function makes sure the root CA is defined inside of a specified secret. If the root CA is not found, then the controller creates it, stores it into Kubernetes and then creates a new certificate to be used by the webhook endpoint provided by the controller.

I think the name of the function is misleading

internal/pkg/certificates/certificates.go
flavio: I **think** this is done because `Immutable` is a `*bool`, hence we either declare a variable before and then reference it, or we use `new`.

The end result is the same, isn't it?

internal/pkg/crypto/ca.go
```diff
@@ -18,28 +18,28 @@ type CA struct {
...
 	if err != nil {
-		return "", errors.Join(errors.New("cannot PEM encode certificate"), err)
+		return nil, errors.Join(errors.New("cannot PEM encode certificate"), err)
 	}
-	return caPEM.String(), nil
+	return caPEM, nil
 }
 
-func (ca *CA) PEMEncodePrivateKey() (string, error) {
+func (ca *CA) PEMEncodePrivateKey() (*bytes.Buffer, error) {
```
flavio, line 33: Same here, is there a reason to not return a `string`?

internal/pkg/crypto/integration_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"crypto/x509"
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+)
+
+const port = "8181"
+
+func TestCAAndCertificateCreationInAHttpsServer(t *testing.T) {
```
flavio, line 15: If we want to make sure the certificate being issued is valid, we could use the `Certificate.Verify` function, like done inside of the [example](https://pkg.go.dev/crypto/x509#Certificate.Verify) provided by the std library.
Using a http server is overly complicated IMHO.

flavio (changes requested):

.gitignore
```diff
@@ -26,3 +26,4 @@ testbin/*
 
 # Manually generated crds. Directory is needed though
 generated-crds/*
+vendor/
```
fabriziosestito, line 29: why is this needed?

internal/pkg/certificates/certificates.go
viccuad: I think you are right, in that case, ignore this conversation.

.gitignore
jvanz: Just to avoid adding the vendor directory in a commit.

internal/pkg/certificates/certificates.go
jvanz: I think you are talking about the function inside the `internal/pkg/admission` and `internal/pkg/admissionregistration`. I'm aware of that. Actually, the `internal/pkg/certificates` package is a kind of copy of some of those functions. However, I have not merge all of the code yet. Just to avoid having a huge PR as before. I'll merge the code  when I start to change the reconciliation loop for policy server and policies. For now, I'm reorganizing the necessary  functions on demand.

internal/pkg/crypto/ca.go
jvanz: To void another conversion. This function is returning `*bytes.Buffer` because I do not need string value. All the usage of the data returned by this function needs to be bytes. Not strings. But, I keep using `*bytes.Buffer` instead of `[]byte` because if in the next changes is necessary a string I could access it using the available function [`String`](https://pkg.go.dev/bytes@go1.21.4#Buffer.String) or [`Bytes`](https://pkg.go.dev/bytes@go1.21.4#Buffer.Bytes).

internal/pkg/crypto/ca.go
jvanz: To void another conversion. This function is returning `*bytes.Buffer` because I do not need string value. All the usage of the data returned by this function needs to be bytes. Not strings. But, I keep using `*bytes.Buffer` instead of `[]byte` because if in the next changes is necessary a string I could access it using the available function [`String`](https://pkg.go.dev/bytes@go1.21.4#Buffer.String) or [`Bytes`](https://pkg.go.dev/bytes@go1.21.4#Buffer.Bytes).

internal/pkg/crypto/integration_test.go
jvanz: I see. Indeed, that makes sense. As I said in a previous comment, as this package is a children from other packages in the `internal/pkg` directory. Which includes the test cases as well. Let me adapt the tests

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -1,88 +1,60 @@
...
-		secretContents   map[string][]byte
-		generateCACalled bool
+		name           string
+		r              Reconciler
+		secretContents map[string][]byte
 	}{
-		{"Existing CA", createReconcilerWithExistingCA(), nil, mockSecretContents, false},
-		{"CA does not exist", createReconcilerWithEmptyClient(), nil, caSecretContents, true},
+		{"Existing CA", createReconcilerWithExistingCA(caSecretContents), caSecretContents},
+		//FIXME {"CA does not exist", createReconcilerWithEmptyClient(), caSecretContents},
```
flavio, line 35: Is there something that prevents us from having this test case enabled as part of this PR?

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -1,88 +1,60 @@
...
 	}
 
 	for _, test := range tests {
 		ttest := test // ensure tt is correctly scoped when used in function literal
 		t.Run(ttest.name, func(t *testing.T) {
-			secret, err := ttest.r.fetchOrInitializePolicyServerCARootSecret(context.Background(), generateCAFunc, pemEncodeCertificateFunc)
+			secret, _ := ttest.r.fetchOrInitializePolicyServerCARootSecret(context.Background())
```
flavio, line 41: I think we should add an assertion that ensures `err` is `nil`

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -91,47 +63,34 @@ func TestFetchOrInitializePolicyServerSecret(t *testing.T) {
...
-		secretContents     map[string]string
-		generateCertCalled bool
+		name           string
+		r              Reconciler
+		secretContents map[string]string
 	}{
-		{"Existing cert", createReconcilerWithExistingCert(), nil, mockSecretCert, false},
-		{"cert does not exist", createReconcilerWithEmptyClient(), nil, caSecretContents, true},
+		{"Existing cert", createReconcilerWithExistingCert(caSecretContents), caSecretContents},
+		//FIXME {"cert does not exist", createReconcilerWithEmptyClient(), caSecretContents},
```
flavio, line 71: same here

internal/pkg/admission/policy-server-ca-secret_test.go
```diff
@@ -91,47 +63,34 @@ func TestFetchOrInitializePolicyServerSecret(t *testing.T) {
...
 	}
 
 	for _, test := range tests {
 		ttest := test // ensure tt is correctly scoped when used in function literal
 		t.Run(ttest.name, func(t *testing.T) {
-			secret, err := ttest.r.fetchOrInitializePolicyServerCASecret(context.Background(), "policyServer", caSecret, generateCertFunc)
+			secret, _ := ttest.r.fetchOrInitializePolicyServerCASecret(context.Background(), "policyServer", caSecret)
```
flavio, line 77: same here

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+	k8s := mgr.GetClient()
+	rootCASecret := &corev1.Secret{}
+	err := k8s.Get(ctx, client.ObjectKey{Namespace: deploymentsNamespace, Name: constants.RootCASecretName}, rootCASecret)
+	if apierrors.IsNotFound(err) {
+		rootCA, err := createRootCa(ctx, k8s, rootCASecret, deploymentsNamespace)
+		if err != nil {
+			return errors.Join(errors.New("unable to create root CA secret"), err)
+		}
+		logger.Info("Root CA secret created")
+		err = renewControllerCerts(ctx, k8s, rootCA, []string{fmt.Sprintf("%s.%s.svc.cluster.local", controllerWebhookServiceName, deploymentsNamespace), fmt.Sprintf("%s.%s.svc", controllerWebhookServiceName, deploymentsNamespace)}, deploymentsNamespace)
```
flavio, line 38: nitpicking, maybe we should create some variables to hold the results of the `fmt.Sprintf`. The names of these variables would make clear what they are about to someone reading the code

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+			return errors.Join(errors.New("unable to PEM encode CA certificate"), err)
+		}
+		err = updateWebhooksCaBundle(ctx, k8s, certificatePEMEncoded.Bytes())
+		if err != nil {
+			return err
+		}
+		logger.Info("Certificates and webhooks updated")
+		logger.Info("Restarting controller due a root certificate authority renewal")
+		os.Exit(constants.ControllerReturnCodeCAInitialized)
+		return nil
```
flavio, line 54: We can leave that out (and update the function documentation accordingly):

* We are going to create the certificate before starting the webhook server
* The webhook server provided by controller runtime can detect changes done to the certificate & its key. We just need to update the Secret, kubelet will sync the contents on the files mounted inside of the Pod, controller-runtime will reload them and start answering with the new cert

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		logger.Info("Start to update webhook with latest CA")
+		certificatePEMEncoded, err := rootCA.PEMEncodeCertificate()
+		if err != nil {
+			return errors.Join(errors.New("unable to PEM encode CA certificate"), err)
+		}
+		err = updateWebhooksCaBundle(ctx, k8s, certificatePEMEncoded.Bytes())
+		if err != nil {
+			return err
+		}
```
flavio, line 50: Can you please switch to the more idiomatic:

```golang
if err = foo(); err != nil {
}
```

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+	rootCASecret.Data = map[string][]byte{
+		"tls.crt": certificatePEMEncoded.Bytes(),
+		"tls.key": privateKeyPEMEncoded.Bytes(),
+	}
+	rootCASecret.Name = constants.RootCASecretName
+	rootCASecret.Namespace = deploymentsNamespace
+	err = k8s.Create(ctx, rootCASecret)
+	if err != nil {
+		err = errors.Join(errors.New("cannot create root ca secret"), err)
+	}
```
flavio, line 84: Can you please switch to the more idiomatic:

```golang
if err = foo(); err != nil {
}
```

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return errors.Join(errors.New("unable to create controller certificate secret"), err)
+	}
+	logger.Info("controller certificate secret created")
+	return nil
+}
+
+func updateWebhooksCaBundle(ctx context.Context, k8s client.Client, caBundle []byte) error {
+	webhookConfigList := admissionregistrationv1.ValidatingWebhookConfigurationList{}
+	err := k8s.List(ctx, &webhookConfigList, client.MatchingLabels{"app.kubernetes.io/part-of": "kubewarden"})
```
flavio, line 117: The controller's webhook certs have a hard coded name, we should just do a `Get`

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,148 @@
...
+			patch.Webhooks[i].ClientConfig.CABundle = caBundle
+		}
+		err = k8s.Patch(ctx, patch, client.MergeFrom(&webhookConfigList.Items[i]))
+		if err != nil {
+			return errors.Join(fmt.Errorf("cannot patch the validating webhook: %q ", webhookConfig.Name), err)
+		}
+	}
+
+	mutatingWebhookConfigList := admissionregistrationv1.MutatingWebhookConfigurationList{}
+	err = k8s.List(ctx, &mutatingWebhookConfigList, client.MatchingLabels{"app.kubernetes.io/part-of": "kubewarden"})
```
flavio, line 133: The controller's webhook certs have a hard coded name, we should just do a `Get`

internal/pkg/constants/constants.go
```diff
@@ -54,4 +56,8 @@ const (
 	WebhookConfigurationPolicyScopeLabelKey          = "kubewardenPolicyScope"
 	WebhookConfigurationPolicyNameAnnotationKey      = "kubewardenPolicyName"
 	WebhookConfigurationPolicyNamespaceAnnotationKey = "kubewardenPolicyNamespace"
+
+	ControllerReturnCodeSuccess       = 0
+	ControllerReturnCodeError         = 1
+	ControllerReturnCodeCAInitialized = 2
```
flavio, line 62: I think we can remove this one, see my previous comment about exiting once the CA is updated/initialized

internal/pkg/crypto/cert.go
```diff
@@ -50,7 +49,7 @@ func GenerateCert(ca []byte,
 	newCertificate := x509.Certificate{
 		SerialNumber: serialNumber,
 		Subject: pkix.Name{
-			CommonName:    commonName,
+			// CommonName:    "", // CommonName is deprecated. Use SANS instead.
```
flavio, line 52: Let's do it right now. It looks like we are not setting the SAN

internal/pkg/crypto/integration_test.go
flavio: Reminder to update the test case

internal/pkg/webhookserver/server.go
flavio, line 1: Let's either move the the synchronous init as part of this PR or let's do that with another PR. In both cases, let's remove this file (if I got this right, this is required to have the async init)

internal/pkg/constants/constants.go
flavio: On top of that, if we remove the exit status from above I think we can remove all the other constants and just revert to the initial code.

main.go
```diff
@@ -93,6 +94,10 @@ func main() {
 		"deployments-namespace",
 		"",
 		"The namespace where the kubewarden resources will be created.")
+	flag.StringVar(&controllerWebhookServiceName,
+		"controller-webhook-service-name",
+		"",
+		"The controller webhook service name")
```
flavio, line 100: Do we really need it? I think this was needed in the past before we used Tilt for local development.

main.go
```diff
@@ -64,12 +64,13 @@ func init() {
 }
 
 func main() {
-	retcode := 0
+	retcode := constants.ControllerReturnCodeSuccess
```
flavio, line 67: I think we can revert this change, see my previous comment about the exit codes

main.go
```diff
@@ -105,7 +110,7 @@ func main() {
 		shutdown, err := metrics.New(openTelemetryEndpoint)
 		if err != nil {
 			setupLog.Error(err, "unable to initialize metrics provider")
-			retcode = 1
+			retcode = constants.ControllerReturnCodeError
```
flavio, line 113: I think we can revert this change, see my previous comment about the exit codes

main.go
```diff
@@ -118,7 +123,7 @@ func main() {
 
 			if err := shutdown(ctx); err != nil {
 				setupLog.Error(err, "Unable to shutdown telemetry")
-				retcode = 1
+				retcode = constants.ControllerReturnCodeError
```
flavio, line 126: I think we can revert this change, see my previous comment about the exit codes

flavio (changes requested):
Good improvement over the previous iteration. Thanks for having removed the code duplication!

internal/pkg/webhookserver/server.go
fabriziosestito: I think that this is not needed anymore

jvanz:
@kubewarden/kubewarden-developers this is not ready for review. I've moved it back to in progress. It still missing the sync cert creation.

jvanz:
@kubewarden/kubewarden-developers can you do another round of reviews? I still want to do some tests. But consider the time and timezone differences, I'll not finish Today and I would like to work in your possible feedback Tomorrow. ;-)

internal/pkg/crypto/ca.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("failed to decode root ca private key")
+	}
+
+	caPrivateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("CA private key could not be extracted from secret %s", caSecret.Kind)
+	}
+
+	return &CA{CaCert: *caCert, CaCertBytes: caCertificateBytes.Bytes, CaPrivateKey: caPrivateKey}, nil
+}
```
jvanz, line 133: I'm aware that this function is very similar to another in the `internal/pkg/admission/policy-server-ca-secret.go`. I have not remove that function because that require further changes in the reconciliation loop to keep  the tests passing and I would like to keep this PR as small as possible. This reconciliation loop changes will be the next step.

main.go
```diff
@@ -184,6 +192,37 @@ func main() {
...
+		retcode = 1
+		return
+	}
+
+	if err := setupWebhooks(mgr, deploymentsNamespace); err != nil {
+		retcode = 1
+		return
+	}
+
+	//+kubebuilder:scaffold:builder
```
viccuad, line 205: question: this is for injecting the kubebuilder scaffolded code for the controllerManager that does the `setupWithManager()`, correct?
I couldn't find any docs, only this [comment](https://togithub.com/kubernetes-sigs/kubebuilder/issues/1487#issuecomment-1485346413). The relevant code seems to be [pkg/machinery](https://github.com/kubernetes-sigs/kubebuilder/blob/a897dae204d9b32483d14c60a78573fca6827378/pkg/machinery).

main.go
```diff
@@ -9,7 +10,7 @@ You may obtain a copy of the License at
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+r WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
viccuad, line 13: whoops, a word was eaten.

internal/pkg/crypto/ca.go
viccuad: I still think that String would be a better fit than bytes.Buffer or []Bytes. Once we have created the PEM encoded certificate, we don't need to change it (String is immutable, the other not), we don't need to sequentially add chars (so we don't need a Buffer that is cost-effective for that), and the result is supposed to be human-readable (String).

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,322 @@
...
+		logger.Info("root CA secret created")
+		certificatePEMEncoded, err := rootCA.PEMEncodeCertificate()
+		if err != nil {
+			return nil, false, errors.Join(errors.New("unable to PEM encode CA certificate"), err)
+		}
+		if err := updateWebhooksCaBundle(ctx, k8s, deploymentsNamespace, certificatePEMEncoded.Bytes()); err != nil {
+			return nil, false, err
+		}
+	} else {
+		rootCA, err = controllerCrypto.ExtractCaFromSecret(rootCASecret)
```
viccuad, line 77: I like this, just to make sure that we have renewed correctly.

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -80,24 +76,29 @@ func (r *Reconciler) buildPolicyServerCASecret(policyServerName string, caSecret
...
 	}
-	caPrivateKeyBytes, ok := caSecret.Data[constants.PolicyServerCARootPrivateKeyCertName]
-	if !ok {
+	caCert, err := x509.ParseCertificate(caCertBytes)
+	if err != nil {
+		return nil, fmt.Errorf("CA certificate could not be extracted from secret %s", caSecret.Kind)
+	}
+
+	caPrivateKeyBytes, hasCARootCert := caSecret.Data[constants.PolicyServerCARootPrivateKeyCertName]
+	if !hasCARootCert {
```
viccuad, line 90: Nit for clarity:

```suggestion
	if !hasCARootCertKey {
```

viccuad (commented):
Looks good here! I would like the PEM encoded certs to be stored as String though, but that's it.

jvanz:
> Looks good here! I would like the PEM encoded certs to be stored as String though, but that's it.

Do you mean the data in the secret?

viccuad:
> @viccuad , do you mean the data in the secret?

No, I was refering to https://github.com/kubewarden/kubewarden-controller/pull/550#discussion_r1427862729

internal/pkg/certificates/certificates.go
```diff
@@ -0,0 +1,322 @@
...
+//  3. If root CA is created or webhook server certificate is missing, create a
+//     certificate to be used by the webhook server and store
+//
+// it in a secret
+//  4. Wait until the webhooks server certificates are mount in the local
+//     filesystem.
+//
+// It returns an error if something wrong happen in any of the step mentioned
+// above
+func SetupCA(ctx context.Context, k8sClient client.Client, deploymentsNamespace string, controllerWebhookServiceName string) error {
```
fabriziosestito, line 225: nit(readability): we could move the public function on top of the file so it's easier to understand the public interfaces.

internal/pkg/admission/reconciler.go
```diff
@@ -155,7 +154,7 @@ func (r *Reconciler) Reconcile(
 	policyServer *policiesv1.PolicyServer,
 	policies []policiesv1.Policy,
 ) error {
-	policyServerCARootSecret, err := r.fetchOrInitializePolicyServerCARootSecret(ctx, admissionregistration.GenerateCA, admissionregistration.PemEncodeCertificate)
+	policyServerCARootSecret, err := r.fetchOrInitializePolicyServerCARootSecret(ctx)
```
fabriziosestito, line 157: nice one, I like we are getting read of the inversion of control here

Tiltfile
```diff
@@ -10,8 +10,8 @@ if str(local("command -v " + kubectl_cmd + " || true", quiet = True)) == "":
     fail("Required command '" + kubectl_cmd + "' not found in PATH")
 
 # Install cert manager
-load('ext://cert_manager', 'deploy_cert_manager')
-deploy_cert_manager()
+#load('ext://cert_manager', 'deploy_cert_manager')
```
fabriziosestito, line 13: we can remove this

internal/pkg/crypto/ca.go
```diff
@@ -0,0 +1,133 @@
+package crypto
```
fabriziosestito, line 1: I think that the  package  name is a bit confusing. Usually crypto is used for packages that deal with cryptography algorithms, in this case I think this package can be merged with the certificates package

fabriziosestito (approved):
Overall LGTM!
Just some minor comments added.

I see that the policy-server controller now uses the certificate generation from the crypto package. Ideally we should use the same certificate generated in the main setup function. Is this planned for the next iteration?

internal/pkg/crypto/ca.go
jvanz: That's right. I'll push the changes fixing that soon

viccuad (approved):
LGTM! Quite a job!

The only thing I don't like is https://github.com/kubewarden/kubewarden-controller/pull/550#discussion_r1360424422, but I don't want to block the PR for it.

PR #553, opened 2023-10-09
fix!: Don't reconcile KUBEWARDEN_LOG_FMT envvar

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/issues/310

Reconciling this env var means users cannot set it as needed, for example to configure exporters to output JSON.

## Test

Not needed.

### DIFF
--- a/internal/pkg/admission/policy-server-deployment_test.go
+++ b/internal/pkg/admission/policy-server-deployment_test.go
@@ -11,14 +11,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const policyServerName = "testing"
-const policyServerContainerName = "policy-server-testing"
-const invalidPolicyServerName = "invalid"
-const dropCapabilityAll = "all"
+const (
+	policyServerName          = "testing"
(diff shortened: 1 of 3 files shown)


internal/pkg/admission/policy-server-deployment_test.go
```diff
@@ -435,7 +438,7 @@ func TestDefaultContainerSecurityContext(t *testing.T) {
 }
 
 func TestMetricAndLogFmtEnvVarsDetection(t *testing.T) {
-	for _, envVarName := range []string{constants.PolicyServerEnableMetricsEnvVar, constants.PolicyServerLogFmtEnvVar} {
+	for _, envVarName := range []string{constants.PolicyServerEnableMetricsEnvVar} {
```
fabriziosestito, line 441: we could change the test name

fabriziosestito (approved):
LGTM, just left a minor comment

internal/pkg/admission/policy-server-deployment_test.go
viccuad: True! done

PR #557, opened 2023-10-11
docs: Reformat CRD docstrings for automation into main docs site

jhkrug:
Changes to docstrings to be better formatted by the time they reach a docs page. So, added HTML for a few `<li>` and `<pre>` elements. It's messy with plenty of `<br/>` and `&nbsp;`, so needs some thought.

Minor changes to the Makefile to build markdown rather than asciidoc.


pkg/apis/policies/v1alpha2/policy_types.go
```diff
@@ -76,17 +76,20 @@ type PolicySpec struct {
 
 	// matchPolicy defines how the "rules" list is used to match incoming requests.
 	// Allowed values are "Exact" or "Equivalent".
-	//
-	// - Exact: match a request only if it exactly matches a specified rule.
+	//<ul>
```
viccuad, line 79:
```suggestion
	// <ul>
```

The golangci-lint run in the CI complains of the following:
https://github.com/kubewarden/kubewarden-controller/actions/runs/6482896028/job/17603316697?pr=557#step:4:13844
> commentFormatting: put a space between `//` and comment text (gocritic)


Here is the only place it happens, from what I see.

viccuad (approved):
LGTM. I'm happy merging these changes to the go doc strings. We are the only consumers of this for now.
I expect this to change once or twice per year or so, and given that we need to keep backwards-compat, it should be only additions.

viccuad:
renamed title, added labels so release-drafter job adds its to the release changelog.

flavio (approved):
It's a bit weird to have htlm tags inside of the go code, I wonder how IDE/godoc will react to that.

However, I'm fine with this change. I doubt many people look at the inline doc from an IDE/need to use the godoc for this specific part of our code

Issue #224, opened 2022-05-17
 Research if controller is able to detect and fix changes between the webhook object and its related `ClusterAdmissionPolicy`/`AdmissionPolicy` configuration.


jvanz:
Verify if the Kubewarden controller is able to detect divergences between the webhook objects and the `ClusterAdmissionPolicy`/`AdmissionPolicy` configuration. If so, try to make the controller able to restore the webhook to the expected state.  This issue trys to help mitigate the [threat #4](https://github.com/kubewarden/rfc/blob/main/rfc/0006-threat-model.md#threat-4---attacker-has-rights-to-delete-or-modify-the-k8s-webhook--object)

**NOTE:** This is an issue created from RFC discussing the admission control threat model. It's created to allow the Kubewarden team discuss the proposed mitigation further and start to work on each actionable item when possible

flavio:
We want the `(Validating|Mutating)WebhookConfiguration` object associated to a `(Cluster)|AdmissionPolicy` to be like our controller originally created it.

If a user changes these low level configuration objects, the controller should revert these changes.

viccuad:
Reopening, the feature was reverted from kubewarden-controller @main with https://github.com/kubewarden/kubewarden-controller/pull/569.

The current approach was to set OwnerReferences on the webhooks and watch them, for the controller to retrigger the reconcile of policies, eventually reconciling the webhooks too.

One cannot set an OwnerReference on a dependent across namespaces, or in our case, an OwnerReference on a cluster-wide dependant resource where the owner is a namespace. See [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications). If the GC would allow this, it would need for info to leak between namespaces, to allow for resources in one namespace to block deletion of resources in another one, would need to check for racing conditions, etc.

PR #566, opened 2023-10-30
feat: Always enforce `{Validating,Mutating}WebhookConfiguration` state

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/224

- feat: Reconcile `{Validating,Mutating}WebhookConfigurations` when they are changed by any other means than our controller. This is done by triggering the policy reconciler, which ends up reconciling the WebhookConfigurations.
- Now, the `{Validating,Mutating}WebhookConfigurations` have an `OwnerReference` that points to their policy.
- The `ownerReferences.blockOwnerDeletion` field is automatically set to true, which means that WebhookConfigurations block the deletion of policy CRDs (see [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications)). This doesn't create any problem, and it's beneficial for us.

## Test

- Tested manually, by editing the WebhookConfigurations'  `Webhooks.ClientConfig.Service.Path` for example, to simulate a MITM. After the edit getting successfully applied (bumping the `Generation`), our controller successfully updates the Webhook and restores its configuration.
- CI

### Tradeoff

Given that now the WebhookConfigurations have an `OwnerReference`, we could defer to the K8s garbage collector instead of deleting the webhooks on our own via the policy reconciler. Still, I prefer to let our [reconciler do it](https://github.com/kubewarden/kubewarden-controller/blob/main/controllers/policy_utils.go#L186).

### Potential improvement

Note that the reconciliation of older WebhookConfigurations that don't have the `OwnerReference` yet doesn't happen until they or the policies get an update. Once that happens, they will have the `OwnerReference` and be watched.
This is the case for already-running Kubewarden deployments. We could try to treat this case (by adding controller code, refreshing all the policies resources, etc).

### DIFF
--- a/internal/pkg/admission/mutating-webhook.go
+++ b/internal/pkg/admission/mutating-webhook.go
@@ -7,12 +7,14 @@ import (
 	"reflect"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 )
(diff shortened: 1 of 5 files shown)


viccuad:
rebased on top of main.

internal/pkg/admission/mutating-webhook.go
```diff
@@ -103,4 +111,15 @@ func (r *Reconciler) mutatingWebhookConfiguration(
 			},
 		},
 	}
+
+	// set policy as the owner of the webhook
+	scheme := runtime.NewScheme()
+	if err := policiesv1.AddToScheme(scheme); err != nil {
+		return nil, fmt.Errorf("cannot add policies.kubewarden.io/v1 to scheme: %w", err)
+	}
+	if err := controllerutil.SetControllerReference(policy, &webhook, scheme); err != nil {
```
jvanz, line 120: Should not we use `SetOwnerReference`? 
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/controller/controllerutil#SetOwnerReference

jvanz (changes requested):
I have a question. I think the code is using the wrong functon to set the owner reference.

internal/pkg/admission/mutating-webhook.go
viccuad: I had the same mistake when I first tried it :D. See the commit message of https://github.com/kubewarden/kubewarden-controller/pull/566/commits/db21d3cee95740bc79594fd1bf357085352dd823. Note that `SetOwnerReference` sets the OwnerReference field, but doesn't get the object reconciled, which is what we want here.

jvanz: Indeed. My fault. I skin read the docs and missed that. I was intrigued by method name. hehe

jvanz (approved):

Issue #548, opened 2023-10-02
spike: consider using Tilt and Telepresence for development  


fabriziosestito:
We could consider using [Tilt](https://github.com/tilt-dev/tilt) and [Telepresence](https://github.com/telepresenceio/telepresence) for development. 
This means we could get rid of `kube-webhook-wrapper` related code, that is now needed in development mode.

Related to: https://github.com/kubewarden/kubewarden-controller/pull/541

flavio:
The reason we're looking into these solutions is to allow a better developer experience when working on the controller. Most important, we want to remove the hackish way we expose the controller HTTPs server during the development mode.

This https server is used to serve the webhook endpoints that validate our CRDs.

To decide which solution to pick, I propose to implement the following test:

1. Create a simple web application

The application will expose one single route `/` which returns a hello world. I think the application should be written using Go using the standard libraries, like with do with the controller. Basically what is shown [here](https://pkg.go.dev/net/http#hdr-Servers).

The application will start a HTTPS server that uses a certificate and key loaded from the local filesystem.

2. Define a Kubernetes Service

The application should be exposed inside of the Kubernetes cluster using a `ClusterIP` service.

3. Access the application from the cluster

From within the cluster, start a container and try to access the application by going through the internal service.

A container with alpine + curl should just work out of the box.

The CA used to generate the application certificate can be copy-pasted into the container

4. Do changes to the local app

Change the `hello world` response to something else. Ensure the change is seen by the alpine container running inside of the cluster.

jvanz:
It looks like to me that telepresence is the closest tool to archive what we do now manually. Which is running a local process that is accessible from the cluster workloads. And it works fine. This is a example:

Let's create a simple web server in go:

```go
package main

import (
	"fmt"
	"net/http"
)

func handler(w http.ResponseWriter, r *http.Request) {
	fmt.Println("Handler here!")
	responseString := "Hello Kubewarden!"
	fmt.Fprintln(w, responseString)
}

func main() {
	http.HandleFunc("/", handler)
	fmt.Println("Servidor em execução na porta 8080...")
	http.ListenAndServe(":8080", nil)
}
```

[...]

PR #567, opened 2023-10-30
feat(docs): using telepresence in development mode.

jvanz:
## Description

Updates the CONTRIBUTING.md adding instructions on how to use telepresence to run the Kubewarden controller locally.

Fix https://github.com/kubewarden/kubewarden-controller/issues/548

## Test

Follow the instructions added in the PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-This document contains instructions on how to build and run locally the controller
+This document contains instructions on how to build and run the controller locally
 allowing developers to test their changes.
 
 ## Building
@@ -54,7 +54,7 @@ to access workloads in the cluster and receive requests to the controller runnin
 in the developer machine.  
 
 > Install the telepresence binary from the Github releases page. The official 
-> once from Ambassador Labs may require login. 
+> ones from Ambassador Labs may require login. 
 
 Before running the controller, install the custom resource definitions:
 
@@ -71,10 +71,10 @@ kubectl create ns kubewarden
 Now, as we are using `telepresence` to intercept the intra cluster
 communication, we need to create the controller deployment and service that
 `telepresence` will intercept. For this, install the Kuberwarden stack
-following the steps described quickstart guide. In other words, install
-Kubewarden controller helm chart. However, a change is required. The controller
-deployment must allow root users. This is disable by default and it is required
-because `telepresence` will add a init container in the deployment which need
+following the steps described in the quickstart guide. In other words, install
+the Kubewarden controller helm chart. However, a change is required. The controller
+deployment must allow root users. This is disabled by default and it is required
+because `telepresence` will add a init container in the deployment which needs
 root access.
 
 ```console
@@ -137,9 +137,9 @@ spec:
 EOF
 ```
 
-> Note: if you want to use some kind o NetworkPolicy to block the original controller traffic
-> you may need to have a multi node cluster. Because in a single node cluster the traffic 
-> policies will not be applied. Because the traffic will not leave the node and never the
+> Note: if you want to use some kind of NetworkPolicy to block the original controller traffic
+> you may need to have a multi node cluster. In a single node cluster the traffic 
+> policies will not be applied. Because the traffic will not leave the node and will never be
 > checked against the network policies
 
 To intercept requests sent to the Kubewarden controller run the following command:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,8 @@
 # Contributing
 
+This document contains instructions on how to build and run locally the controller
+allowing developers to test their changes.
+
 ## Building
 
 To build kubewarden-controller some packages are required. If you are using
@@ -20,9 +23,8 @@ make
 You can run the controller by executing `make run`. This Makefile
 target executes the controller locally as a regular process.
 
-In order to execute this Makefile target, you need to have created a
-Kubernetes cluster with `k3d`, `kind` or `minikube` that is reachable
-through your `~/.kube/config` kubeconfig file.
+In order to execute this Makefile target, you need to have created a Kubernetes
+cluster with that is reachable through your `~/.kube/config` kubeconfig file.
 
 These are the relevant environment variables:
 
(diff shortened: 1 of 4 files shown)


flavio:
I followed the instructions of the PR locally. There are some extra things that have to be done.

## Avoid fights between the in-cluster controller and the one on the dev machine

Telepresence will forward the webhook calls to the local process, but the Pod running the controller will keep working inside of the cluster.

Because of that, we will end up having two controllers attempting to reconcile the same set of resources. This leads to "interesting" situations... like the default PolicyServer being continuously redeployed because the two controllers keep changing the Deployment definition :facepalm: 

To prevent that from happening, delete the controller that was created by the helm chart:

```console
kubectl delete -n kubewarden deployments.apps kubewarden-controller
```

Create a new one using the dummy file I've attached. We must have a Deployment defined for kubewarden-controller, otherwise telepresence won't be able to intercept it.

```console
kubectl apply --namespace kubewarden -f hack/kubewarden-controller-dummy-deployment.yaml
```

## Use the proper certificate

The local webhook server is reached by the kube-api-server, but it complains because the certificate being used uses the wrong SAN. It's not the one of the cluster-ip, but it's something about localhost.

I've fixed that by changing the `run` target of the Makefile to:
* create an empty `cert` directory
* download the certificate and its private key from the kubernetes secret

I've then patched the main program to take a new cli param that is used to customize the directory where these files are expected to be. By default the code from kube-builder searches the certificate under `/tmp/something`, which is the path where k8s injects the secret, this flag overrides this behavior.

The controller has to be started with the following command, this should be patched inside of the instructions:

```console
WEBHOOK_HOST_LISTEN=127.0.0.1make run
```
Now everything should be working as expected

## Don't forget to quit!

The instructions should also mention that you should do a `telepresence quit` once you're done with it (or if you delete the testing cluster and create it again). Otherwise weird behaviors will pester you.

## Summary

The right order to follow is:

* install cert-manager
* install kubewarden crds
* create kubewarden namespace
* install the kubewarden-controller helm chart, I'm pretty sure we can use the default options (no need to change the run as root because we're going to delete the deployment)
* delete the deployment of the kubewarden-controller
* create the dummy deployment of the kubewarden controller
* telepresence helm install
* telepresence connect -n kubewarden
* telepresence intercept kubewarden-controller --port 9443:443
* `WEBHOOK_HOST_LISTEN=127.0.0.1make run`

Everything should be working now, to verify that:

```console
helm install -n kubewarden kubewarden-defaults --set recommendedPolicies.enabled=True kubewarden/kubewarden-defaults
```

You should see all the output produced by the controller running on your workstation

## More notes

We are not running the controller in "local development" mode anymore. Because of that the webhooks exposed by the controller are NOT registered by the Go code. We rely on the `(Validating|Mutating)WebhookConfiguration` objects created by the helm chart.

This is fine, unless you add a new endpoint that is not yet known by the helm chart. I think this is fine, we have to document that.

flavio:
@jvanz I've done a forced push against your branch because I needed to rebase against `main`

jvanz:
> > Telepresence will forward the webhook calls to the local process, but the Pod running the controller will keep working inside of the cluster.
> 
> I'm not facing this issue. After the pod with the `telepresence` agent is running the old one is removed:
> 
> ```
> $ k8s get pods -n kubewarden
> NAME                                     READY   STATUS    RESTARTS   AGE
> kubewarden-controller-5577f5dbb6-m7w9w   2/2     Running   0          3m41s
> ```
> 
> Furthermore, if we applied the deployment manually, we will not using the production helm charts which, as far as I remember, will be one of our goal. To be as the closest as we can from the production. I agree that there are cases where we need something more customized that may require another deployment file and so on. But I'm not sure if this will be required for most of the cases.

CONTRIBUTING.md
```diff
@@ -1,5 +1,8 @@
 # Contributing
 
+This document contains instructions on how to build and run locally the controller
```
jhkrug, line 3:
```suggestion
This document contains instructions on how to build and run the controller locally
```

CONTRIBUTING.md
```diff
@@ -41,65 +43,171 @@ These are the relevant environment variables:
...
+
+To run the controller in the developer workstation it's possible to use
+[telepresence](https://github.com/telepresenceio/telepresence). It allows
+developers to "add" the local workstation in the cluster. Thus, it's possible
+to access workloads in the cluster and receive requests to the controller running
+in the developer machine.  
 
-### Install Custom Resource Definitions
+> Install the telepresence binary from the Github releases page. The official 
+> once from Ambassador Labs may require login.
```
jhkrug, line 57:
```suggestion
> Install the telepresence binary from the Github releases page. The official 
> ones from Ambassador Labs may require login. 
```

CONTRIBUTING.md
```diff
@@ -41,65 +43,171 @@ These are the relevant environment variables:
...
 
 ```console
 kubectl create ns kubewarden
 ```
 
-### Running
+Now, as we are using `telepresence` to intercept the intra cluster
+communication, we need to create the controller deployment and service that
+`telepresence` will intercept. For this, install the Kuberwarden stack
+following the steps described quickstart guide. In other words, install
```
jhkrug, line 74:
```suggestion
following the steps described in the quickstart guide. In other words, install
```

CONTRIBUTING.md
```diff
@@ -41,65 +43,171 @@ These are the relevant environment variables:
...
+Kubewarden controller helm chart. However, a change is required. The controller
```
jhkrug, line 75:
```suggestion
the Kubewarden controller helm chart. However, a change is required. The controller
```

CONTRIBUTING.md
```diff
@@ -41,65 +43,171 @@ These are the relevant environment variables:
...
+deployment must allow root users. This is disable by default and it is required
+because `telepresence` will add a init container in the deployment which need
+root access.
```
jhkrug, line 78:
```suggestion
deployment must allow root users. This is disabled by default and it is required
because `telepresence` will add a init container in the deployment which needs
root access.
```

CONTRIBUTING.md
```diff
@@ -41,65 +43,171 @@ These are the relevant environment variables:
...
+      serviceAccount: kubewarden-controller
+      serviceAccountName: kubewarden-controller
+  replicas: 1
+EOF
+```
+
+> Note: if you want to use some kind o NetworkPolicy to block the original controller traffic
+> you may need to have a multi node cluster. Because in a single node cluster the traffic 
+> policies will not be applied. Because the traffic will not leave the node and never the
+> checked against the network policies
```
jhkrug, line 143:
```suggestion
> Note: if you want to use some kind of NetworkPolicy to block the original controller traffic
> you may need to have a multi node cluster. In a single node cluster the traffic 
> policies will not be applied. Because the traffic will not leave the node and will never be
> checked against the network policies
```

jhkrug (changes requested):
A few minor suggestions and a typo or two.

jvanz:
After a discussion in other communication channels, we device to use tilt. Closing this PR.

PR #569, opened 2023-11-01
Revert "Merge pull request #566 from viccuad/feat-webhooks-reconciled"

jvanz:
## Description

I've opened this PR to revert the change adding owner reference the webhooks.  This is necessary because the version `v1.9.0-rc2` containing this feature is not able to set the admission policy as active. This is the error from the controller logs:

```
2023-11-01T14:05:06Z    ERROR   Reconciler error        {"controller": "admissionpolicy", "controllerGroup": "policies.kubewarden.io", "controllerKind": "AdmissionPolicy", "AdmissionPolicy": {"name":"pod-privileged","namespace":"kubewarden"}, "namespace": "kubewarden", "name": "pod-privileged", "reconcileID": "cf067c30-e596-4eaa-ae89-91dc683a659e", "error": "error reconciling validating webhook\ncannot set OwnerReference on WebhookConfiguration: cluster-scoped resource must not have a namespace-scoped owner, owner's namespace kubewarden"}
```

As far as I can [see](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/), the issue is that we are using a namespace scoped resource (AdmissionPolicy) as the owner of a cluster scoped resource (ValidatiningWebHook). This is not allowed:

```
Cross-namespace owner references are disallowed by design. Namespaced dependents can specify cluster-scoped or namespaced owners. A namespaced owner must exist in the same namespace as the dependent. If it does not, the owner reference is treated as absent, and the dependent is subject to deletion once all owners are verified absent.

Cluster-scoped dependents can only specify cluster-scoped owners. In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner, it is treated as having an unresolvable owner reference, and is not able to be garbage collected.

In v1.20+, if the garbage collector detects an invalid cross-namespace ownerReference, or a cluster-scoped dependent with an ownerReference referencing a namespaced kind, a warning Event with a reason of OwnerRefInvalidNamespace and an involvedObject of the invalid dependent is reported. You can check for that kind of Event by running kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace.
```

### DIFF
--- a/controllers/admissionpolicy_controller.go
+++ b/controllers/admissionpolicy_controller.go
@@ -26,7 +26,6 @@ import (
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/naming"
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


flavio:
I'm fine with the revert, we can figure out the proper solution without blocking the 1.9 release

flavio:
@kravciak it would be nice to expand the e2e tests to catch this type of regression we've introduced

PR #571, opened 2023-11-04
chore(dx): use tilt for development

fabriziosestito:
## Description

This PR is a POC on using Tilt for local development.

It adds a `Tiltfile` on the root of the project, an example settings file, and a Tilt-specific Dockerfile that is used by Tilt to inject the controller in the cluster and auto-reload it when the code changes.

The Tiltfile executes the following steps:

1. Read the `tilt-settings.yaml` file
2. Deploy cert-manager
3. Applys the CRD from the `config/crd` path of the projects. I didn't use the `kubewarden-crds` helm chart here since a developer could be interested in the latest CRD changes while developing.
4. Read the kubewarden controller rendered templates from the helm chart, overriding the image value with one in a private registry (as for now only `ghcr.io` is supported, see the usage section below).
5. Patch the deployment template adding the possibility to run the process as root. This is needed since Tilt requires root access to live-update the running process.
6. Enable hot reload by watching the source files.
7. Build the controller container, which is wrapped with tilt custom code to allow the auto-reload.
8. Finally, push the controller container and inject it automatically into the deployment.

The last step takes few seconds to complete.
The ghcr.io personal registry will host a new container image that has a `tilt-*` label. Then Tilt will inject the image (or auto reload the existing one) specified by the helm chart values in step 4.

## Usage

To use this setup, a developer needs to run a fresh k3d  cluster and needs login to `ghcr.io` from the docker daemon (`docker login ghcr.io -u user --password-stdin`).
Since k3d uses the same docker daemon, the cluster can access the private ghcr.io registry. This is easier than setting up a local registry or using the `k3d import` command (which apparently is slower).

Then, we need to create `tilt-settings.yaml` in the root of the project.
An example `tilt-settings.yaml.example` is provided.
The settings files look like this:

```
registry: ghcr.io
image: <your github handle>/kubewarden-controller
helm_charts_path: ../helm-charts/
```

where `image` is the ghcr.io location of the tilt managed controller image (it does not need to exist) and the `helm_charts_path` is the path to the local kubewarden helm charts repository.

Now we can start tilt with `tilt up --stream=true` and after the setup, we will see the controller logs running.

Any modification to the controller code will be pushed in the cluster and the controller will be auto-reloaded.
You can test it by adding some info logs in the `main.go` function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ testbin/*
 
 # Manually generated crds. Directory is needed though
 generated-crds/*
+
+# Tilt settings
+tilt-settings.yaml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,76 +15,56 @@ Then, can run the following command to build the package:
 make
 ```
 
-## Running
+## Development
 
-You can run the controller by executing `make run`. This Makefile
-target executes the controller locally as a regular process.
+To run the controller for development purposes, you can use [Tilt](https://tilt.dev/).
 
-In order to execute this Makefile target, you need to have created a
-Kubernetes cluster with `k3d`, `kind` or `minikube` that is reachable
-through your `~/.kube/config` kubeconfig file.
+### Pre-requisites
 
-These are the relevant environment variables:
-
-- `KUBEWARDEN_DEVELOPMENT_MODE`: if its value is `1` or `true`, the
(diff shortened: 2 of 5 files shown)


flavio (approved):
Looks really neat to me. Thanks for having included the change required to use a private registry different from ghcr.io

jvanz (approved):

PR #576, opened 2023-11-14
feat: reconcile policy webhook configurations

fabriziosestito:
## Description

This PR fixes: #224, adding reconciliation of ValidatingWebhookConfiguration and MutatingWebhookConfiguration if those are changed (by the user).

## Test

This PR introduces a k3s `testcontainer` to test the controllers against a real cluster, by using `envtest`. 
The `envtest` setup was already there, but not using a real cluster we weren't able to test Kubernetes builtin resources "side-effects", e.g. creating a deployment was not causing a ReplicaSet and Pods to be created. This is because by default `envtest` spins up a fake control plane with no Kubernetes built-in controllers.
Also, we were lacking tests for the admission policies controllers' happy paths. This PR adds them.

NOTE: Webhooks integration tests are still not running against a real cluster, but using `envtest` in this case might be overkill.
Consider refactoring these tests to unit tests (see: #579)


controllers/suite_test.go
```diff
@@ -65,9 +68,23 @@ var _ = BeforeSuite(func() {
 	ctx, cancel = context.WithCancel(context.TODO())
 
 	By("bootstrapping test environment")
+	k3sContainer, err := k3s.RunContainer(ctx,
+		testcontainers.WithImage("docker.io/rancher/k3s:v1.27.1-k3s1"),
```
fabriziosestito, line 72: maybe from env variable?

go.mod
```diff
@@ -19,16 +19,28 @@ require (
...
+	github.com/Microsoft/hcsshim v0.11.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
+	github.com/containerd/containerd v1.7.7 // indirect
+	github.com/containerd/log v0.1.0 // indirect
+	github.com/cpuguy83/dockercfg v0.3.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/docker/distribution v2.8.2+incompatible // indirect
+	github.com/docker/docker v24.0.6+incompatible // indirect
```
viccuad, line 34: this is quite a list of indirect deps. I'm afraid they will appear in kubewarden-controller image scans.

fabriziosestito: dependency imported in `_test` files are not included in the release binary

controllers/admissionpolicy_controller.go
```diff
@@ -148,3 +156,41 @@ func (r *AdmissionPolicyReconciler) findAdmissionPoliciesForPolicyServer(ctx con
 	}
 	return r.findAdmissionPoliciesForConfigMap(&configMap)
 }
+
+func (r *AdmissionPolicyReconciler) findAdmissionPolicyForWebhookConfiguration(_ context.Context, webhookConfiguration client.Object) []reconcile.Request {
+	if _, found := webhookConfiguration.GetLabels()["kubewarden"]; !found {
+		return []reconcile.Request{}
+	}
+
+	policyScope, found := webhookConfiguration.GetLabels()["kubewardenPolicyScope"]
```
jvanz, line 165: I would say to add this string as a constant in the [constants.go](https://github.com/kubewarden/kubewarden-controller/blob/6361dc9a26885d4c14923665b57eba0b3a444121/internal/pkg/constants/constants.go#L39-L41). As we do for other labels as well. This applies for all the labels and annotations used in your changes

controllers/clusteradmissionpolicy_controller.go
```diff
@@ -145,3 +153,34 @@ func (r *ClusterAdmissionPolicyReconciler) findClusterAdmissionPoliciesForPolicy
...
+		return []reconcile.Request{}
+	}
+
+	policyScope, found := webhookConfiguration.GetLabels()["kubewardenPolicyScope"]
+	if !found {
+		r.Log.Error(nil, "Found a webhook configuration without a scope label", "name", webhookConfiguration.GetName())
+		return []reconcile.Request{}
+	}
+
+	// Filter out ClusterAdmissionPolicies
```
jvanz, line 168:
```suggestion
	// Filter out AdmissionPolicies
```

controllers/suite_test.go
jvanz: Yeah, sounds a good idea. Maybe keep this version as default value but allow a configuration using env var or something similar

controllers/suite_test.go
```diff
@@ -65,9 +69,23 @@ var _ = BeforeSuite(func() {
...
+
+	restcfg, err := clientcmd.RESTConfigFromKubeConfig(kubeConfigYaml)
+	Expect(err).NotTo(HaveOccurred())
+
+	trueValue := true
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: true,
+		Config:                restcfg,
+		UseExistingCluster:    &trueValue,
```
jvanz, line 88: Is this means that now we are running the tests in a cluster that actually reconcile resources (e.g. pod)? Am I understanding correctly?

controllers/utils_test.go
```diff
@@ -18,121 +18,153 @@ package controllers
...
-			Image:    "some-registry/some-policy-server:latest",
+			Image:    "ghcr.io/kubewarden/policy-server:latest",
 			Replicas: 1,
 		},
 	}
 	templateClusterAdmissionPolicy = policiesv1.ClusterAdmissionPolicy{
 		Spec: policiesv1.ClusterAdmissionPolicySpec{
 			PolicySpec: policiesv1.PolicySpec{
-				Module: "registry://some-registry/some/module:latest",
+				Module: "registry://ghcr.io/kubewarden/policies/echo",
```
jvanz, line 50: Can we use policies under the `test` path, as we use in the e2e tests? Just to avoid polluting the statistics by mistake.

jvanz (changes requested):

controllers/utils_test.go
jvanz: This is an [example](https://github.com/kubewarden/kubewarden-end-to-end-tests/blob/ac9b9958a00a9b284d8667708892c766c082a439/resources/policy-pod-privileged.yaml#L7) of the path used. We may need to push the echo policy there. But I don't see this as a big problem.

controllers/suite_test.go
fabriziosestito: Yes!

controllers/utils_test.go
fabriziosestito: Good point, I've changed it to use the pod-privilged policy, since it doesn't really matter which policy we use in this test.

controllers/admissionpolicy_controller.go
fabriziosestito: done in: https://github.com/kubewarden/kubewarden-controller/pull/576/commits/7381c5d59ded0ec721cff4a18efe905db474774b

Makefile
```diff
@@ -111,7 +113,8 @@ unit-tests: manifests generate fmt vet setup-envtest ## Run unit tests.
 
 .PHONY: setup-envtest integration-tests
 integration-tests: manifests generate fmt vet setup-envtest ## Run integration tests.
-	ACK_GINKGO_DEPRECATIONS=2.12.0 KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" go test ./pkg/... ./controllers/... -ginkgo.v -ginkgo.progress -test.v -coverprofile cover.out
+	ACK_GINKGO_DEPRECATIONS=2.12.0 KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" go test ./pkg/... -ginkgo.v -ginkgo.progress -test.v -coverprofile cover.out
```
fabriziosestito, line 116: calling `go test` twice is needed since we have different envtest configuration, otherwise the real cluster tests will not show log in the console for some reason. However, we are planning to rewrite the webhooks tests as unit tests so this should go away soonish.

controllers/utils_test.go
```diff
@@ -18,121 +18,153 @@ package controllers
...
 var (
-	someNamespace = corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "some-namespace",
-		},
-	}
 	templatePolicyServer = policiesv1.PolicyServer{
 		Spec: policiesv1.PolicyServerSpec{
-			Image:    "some-registry/some-policy-server:latest",
+			Image:    "ghcr.io/kubewarden/policy-server:latest",
```
viccuad, line 43: Given that now it gets instantiated, I wonder if we should instead use a known working tag, just in case that `:latest` doesn't end up ready or doesn't behave correctly.

fabriziosestito: @viccuad you are right, what about we use an env variable as we did for the k3s testcontainer?

viccuad: that sounds like a good idea :).

fabriziosestito: done in: https://github.com/kubewarden/kubewarden-controller/pull/576/commits/4efdcc9538326cd596a4bde2f5a7732a49325bc0

flavio (approved):
LGTM, this is some great work :clap:

Makefile
fabriziosestito: @flavio what about adding this to #579 instead?

flavio: even better!

Issue #574, opened 2023-11-13
Remove webhook wrapper library


jvanz:
Remove the [kube-webhook-wrapper](https://github.com/kubewarden/kube-webhook-wrapper)  library in use by the controller. This library is use only for development and it makes our controller behaviour differ from the production environment. Therefore, we need to remove it and keep the controller running in development mode as close as possible to the production environment.

PR #577, opened 2023-11-14
chore: remove kube-webhook-wrapper library.

jvanz:
## Description

To simplify the development process and keep the controller behavior as close as possible to production removes the kube-webhook-wrapper library. This library is used to create all the certificates necessary to run the controller locally. It "simulates" what cert-manager should do in the cluster. However, as we are using tilt to run the development workflow locally, this is not necessary anymore and the lib could be removed.

Fix https://github.com/kubewarden/kubewarden-controller/issues/574


main.go
```diff
@@ -267,120 +253,15 @@ func main() {
...
-						APIGroups:   []string{policiesv1.GroupVersion.Group},
-						APIVersions: []string{policiesv1.GroupVersion.Version},
-						Resources:   []string{"admissionpolicies"},
-					},
-				},
-			},
-			WebhookPath: "/validate-policies-kubewarden-io-v1-admissionpolicy",
-			Mutating:    false,
-		},
+func setupWebhooks(mgr ctrl.Manager, deploymentsNamespace string) error {
```
flavio, line 256: Do we still need this function?

Looking into the [`policiesv1.SetupPolicyServerWebhookWithManager`](https://github.com/kubewarden/kubewarden-controller/blob/6361dc9a26885d4c14923665b57eba0b3a444121/pkg/apis/policies/v1/policyserver_webhook.go#L38-L54) I see there are comments about that being a function required by rafa's library

flavio (changes requested):

main.go
jvanz: I'll check the comment in that function. However, we need this function to make linter happy. It was complaining about the complexity of the man function when I added the calls to setup the webhook in the manager.

fabriziosestito: @flavio the comment was referring to the removal of the workaround to return a function, since the former library was taking a setup function as first class citizen parameter when setting up the webhook. The Setup function as refactored in this PR.

fabriziosestito (approved):
LGTM!

main.go
jvanz: > @flavio the comment was referring to the removal of the workaround of returning a function, since the former library was taking a setup function as first class citizen parameter when setting up the webhook. The PolicyServer webhook setup function was refactored in this PR.

Yes, I've noticed that when I've read the code and already fix it.

PR #588, opened 2023-12-01
Escaping {} in <pre></pre> for docusaurus V3.

jhkrug:
Otherwise, all the nested formatting (yuck) falls to pieces.


viccuad (approved):
Many thanks!

PR #599, opened 2023-12-20
fix: Update SecurityContext of the pod inside of policy-server Deployments

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/pull/361.

Before this, setting the `policyServer.Spec.SecurityContexts.Pod` value would only work on first creation of the Deployment from the PolicyServer CRD (so, on first install of kubewarden-defaults chart for example). With this fix, one can reconfigure the PolicyServers.

## Test

Added a unit test.


flavio (approved):
LGTM

PR #600, opened 2023-12-20
chore: Add example resources and policies to config/samples/

viccuad:
## Description

Just some samples of k8s resources and policies to do exploration testing or demos.

## Test

I use these from time to time, so I pinky swear they work.

### DIFF
--- a/config/samples/example-secrets-pod.yml
+++ b/config/samples/example-secrets-pod.yml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-secrets
+spec:
+  containers:
+    - name: nginx
+      image: nginx
+      imagePullPolicy: IfNotPresent
+      env:
+        - name: rsa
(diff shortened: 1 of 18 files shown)


config/samples/example-secrets-pod.yml
flavio, line 1: I fear this would cause some false positives from security scanners

flavio (commented):
I have mixed feelings. Having these files under `config/samples` is misleading. Maybe having everything under a `demo` directory, nicely grouped together would be beneficial.

What I mean with grouped together is: one folder with the ingress demo that contains the file with the `ClusterAdmissionPolicy` and the demo resources; having a README explaining what's in there would also help.

I wonder what the other @kubewarden/kubewarden-developers think about this PR

config/samples/example-secrets-pod.yml
viccuad: that's a good point, that even a `demo/` folder wouldn't solve. I will shelve this PR for now.

Issue #623, opened 2024-01-31
kubewarden-controller: enable codecov


flavio:
Our unit test should emit information about the code coverage of our codebase. The coverage will then be used by CodeCov.

## Acceptance criteria:

* Enable CodeCov bot on the repository

viccuad:
Performed evaluation of Codecov.
Docs:
- https://docs.codecov.com/docs/github-tutorial
- https://docs.codecov.com/docs/common-recipe-list

Codecov works by running the language code coverage tool, and uploading the results to the Codecov SaSS by using their own `codecov` cli utility. It's free for Open Source repos.
When uploading, one can "flag" different coverage reports to bundle or treat them with different rules. Example: a mature backend that should be well tested and a younger frontend which is allowed to have less tests.

For us, that means separating the unit-tests and integration-tests.

Codecov needs a GH Application enabled for the GH organization, so it can read the repos and write commit statuses and PRs.

Opened https://github.com/kubewarden/kubewarden-controller/pull/631.

PR #631, opened 2024-02-02
ci: Save coverage files, add codecov workflow

viccuad:
## Description

Relates to https://github.com/kubewarden/kubewarden-controller/issues/623.
 
Change makefile targets to save coverage into specific `coverage/{unit-tests,integration-tests}/*` folders.

Add new `codecov.yml` GHA workflow that runs `make test` and uploads the coverage results to codecov via their `codecov` cli.

Tagged unit-tests and integration-tests results separately by setting the codecov flags on the workflow.

Needs a secret `CODECOV_ORG_TOKEN` present, which should be provided by the Codecov GH Application when installed.

## Test

To see this working, we need the [Codecov GH application installed in the org,](https://github.com/apps/codecov) and then we will see it working on the next PR opened against the repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,15 @@ jobs:
         with:
           go-version: "1.21"
       - run: make unit-tests
+      - name: Upload unit-tests coverage to Codecov
+        uses: codecov/codecov-action@e0b68c6749509c5f83f984dd99a76a1c1a231044 # v4.0.1
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_ORG_TOKEN }}
+        with:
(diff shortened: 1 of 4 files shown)


viccuad:
Will change the PR so that the upload of the test coverage results happen as part of the normal workflows included in the CI workflow, to not duplicate runs.

viccuad:
PR changed, the upload seems to happen, see here:
https://github.com/kubewarden/kubewarden-controller/actions/runs/7782853533/job/21220120905#step:5:93

Still, codecov server returns a 500. I would like to merge this and see how it behaves for follow-up PRs that don't come from forks.

flavio (approved):
LGTM

PR #646, opened 2024-02-22
fix(test): rewrite webhooks tests as unit tests.

### LINKED ISSUE #579: Rewrite Webhook tests as unit tests
Currently the webhook tests are written as integration tests, but we think they could be rewritten as unit tests. Integration tests are more expensive than unit tests.

## Acceptance criteria

* [x] Investigate if the webhook tests can be rewritten as unit tests
* [ ] Rewrite the webhook tests as unit tests

jvanz:
## Description

The webhook_suite_test.go originally test webhook validations with integrations test on Ginkgo. Which is slow because it setup a whole test environment to simulate Kubernetes cluster. This commit changes that rewriting the test as unit tests calling the validation function directly.

Fix #579


fabriziosestito (changes requested):
The table testing looks great and IMHO it is much better than relying on etcd to test validations. 
However, I see other improvements that could be made.
Here, we are not testing the webhooks but the pure validation functions, which are common to every struct that implements the `Policy` interface.
It's a bit weird that the admission policy uses those functions from the cluster admission policy type file so that we could have two simple changes to clean up a little bit:
1. move the common validations functions to a separate file (i.e. `policy_validation.go`) and rename the `clusteradmission_webhook_test.go` file accordingly.
2. create `clusteradmissionpolicy_webhook_test` and `admissionpolicy_webhook_test` where we test that the wire-up is ok by testing `Default()`, `ValidateCreate`, and `ValidateUpdate` with simple valid data (don't need to re-test that the validation works, just that the methods are doing what expected). This overlaps a bit with the validation testing but it gives us the confidence that the methods are working correctly (for instance, they are updating the policy struct correctly or returning the expected default type).

pkg/apis/policies/v1/policy_test_utils.go
```diff
@@ -0,0 +1,75 @@
+/*
```
fabriziosestito, line 1: we should rename this to something with the `_test` suffix or these functions will be included in the final binary. (e.g. `utils_test.go`)

fabriziosestito (approved):
LGTM
:shipit: once the utils file is renamed

flavio:
@jvanz I think you can merge this PR after you have rebased it against `main`

PR #654, opened 2024-02-28
fix: policy server resource deleted before policies

### LINKED ISSUE #650: Webhook inconsistencies when deleting a policy

### Current Behavior

In my case if I delete a specific policy and/or its policy server I get the error 

```
Error from server (InternalError): Internal error occurred: failed calling webhook "clusterwide-run-as-non-root.kubewarden.admission": failed to call webhook: Post "https://policy-server-default.kubewarden.svc:8443/validate/clusterwide-run-as-non-root?timeout=10s": service "policy-server-default" not found
```

If I add the policy server again (it is always the `default` server) I get 

```
Error from server (InternalError): Internal error occurred: failed calling webhook "clusterwide-run-as-non-root.kubewarden.admission": failed to call webhook: the server could not find the requested resource
```

I am only able to stop the errors by adding the policy again. Any other policy works ok

### Expected Behavior

I should be able to delete a policy or policy server and not affect the apply of resources after that.

### Steps To Reproduce

delete policy or delete policy server

### Environment

```markdown
- OS: Linux
- Architecture: amd64
```

### Anything else?

_No response_

jvanz:
## Description

There is an issue in the policy server controller which is causing inconsistencies when the policy server is deleted. It's possible, due the concurrent nature of events and controllers loops, where the policies ended in a inconsistent state. Where the policy and its webhooks are register in the cluster. But the policy server which should run it is gone. Causing an issue when the control plane tries to contact the policy server.

This is caused because when the policy server controller deletes all the policy server policies, it filters out policies which have the deletion timestamp field. Once no policies are found, in other words, when no policies without deletion timestamp are found, the policy server resources are deleted. However, the deletion timestamp does not ensure that the resource is not register in the cluster. It tells that the resource is marked to be deleted by the control plane. Which means that we can reach a state where the policy server is deleted before the policies. Causing issue when the control plane tries to send request to the webhooks.This commit fixes that by removing the filter used to get the policies in the deletion reconciliation loop.

Fix #650 

Thanks @flavio for the directions!

### DIFF
--- a/controllers/policyserver_controller_test.go
+++ b/controllers/policyserver_controller_test.go
@@ -21,6 +21,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2" //nolint:revive
 	. "github.com/onsi/gomega"    //nolint:revive
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
@@ -33,23 +34,45 @@ var _ = Describe("PolicyServer controller", func() {
 		Expect(
 			k8sClient.Create(ctx, policyServerFactory(policyServerName)),
 		).To(haveSucceededOrAlreadyExisted())
+		// Wait for the Service associated with the PolicyServer to be created
+		Eventually(func(g Gomega) error {
+			_, err := getTestPolicyServerService(policyServerName)
+			return err
+		}, timeout, pollInterval).Should(Succeed())
 	})
 
-	Context("it has no assigned policies", func() {
+	Context("with no assigned policies", func() {
 		It("should get its finalizer removed", func() {
 			By("deleting the policy server")
 			Expect(
 				k8sClient.Delete(ctx, policyServerFactory(policyServerName)),
 			).To(Succeed())
 
-			policyServer, err := getTestPolicyServer(policyServerName)
-
-			Expect(err).ToNot(HaveOccurred())
-			Expect(policyServer.Finalizers).ToNot(ContainElement(constants.KubewardenFinalizer))
+			Eventually(func(g Gomega) (*policiesv1.PolicyServer, error) {
+				return getTestPolicyServer(policyServerName)
+			}, timeout, pollInterval).ShouldNot(
+				HaveField("Finalizers", ContainElement(constants.KubewardenFinalizer)),
+			)
 		})
+
+		AfterEach(func() {
+			// It's necessary remove the test finalizer to make the
+			// BeforeEach work as extected. Otherwise, the policy service
+			// creation will not work as expected
+			policyServer, err := getTestPolicyServer(policyServerName)
+			Expect(err).Should(Succeed())
+			controllerutil.RemoveFinalizer(policyServer, IntegrationTestsFinalizer)
+			err = reconciler.Client.Update(ctx, policyServer)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func(g Gomega) error {
+				_, err := getTestPolicyServer(policyServerName)
+				return err
+			}, timeout, pollInterval).ShouldNot(Succeed())
+		})
+
 	})
 
-	Context("it has assigned policies", func() {
+	Context("it has assigned policies", Serial, func() {
 		policyName := newName("policy")
 
 		It("should delete assigned policies", func() {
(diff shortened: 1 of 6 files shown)


jvanz:
@flavio I've update the code removing functions not used, adjusting the logs and changing a little bit the tests to allow them run together (I found an issue in the suite where you added the test for this issue). Let me know if I'm missing something.

controllers/policyserver_controller_test.go
```diff
@@ -70,7 +93,44 @@ var _ = Describe("PolicyServer controller", func() {
...
+					HaveField("Finalizers", Not(ContainElement(constants.KubewardenFinalizer))),
+					HaveField("Finalizers", ContainElement(IntegrationTestsFinalizer)),
+				),
+			)
+
+			Expect(
+				getTestPolicyServerService(policyServerName),
+			).To(
+				HaveField("DeletionTimestamp", BeNil()),
+			)
```
flavio, line 112: Looking back at that, we do not add finalizers to the Service object associated with the Policy Server (we don't need to). Because of that, we either will find the Service or not find it.

Can you rewrite this assertion to be:

```golang
Eventually(func(g Gomega) error {
	_, err := getTestPolicyServerService(policyServerName)
	return err
}, timeout, pollInterval).Should(Succeed())
```

controllers/policyserver_controller_test.go
```diff
@@ -70,7 +93,44 @@ var _ = Describe("PolicyServer controller", func() {
...
 		It(fmt.Sprintf("should get its %q finalizer removed", constants.KubewardenFinalizer), func() {
+			By("not having policies assigned")
+			policy, err := getTestClusterAdmissionPolicy(policyName)
+			Expect(err).ToNot(HaveOccurred())
+
+			controllerutil.RemoveFinalizer(policy, IntegrationTestsFinalizer)
+			err = reconciler.Client.Update(ctx, policy)
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func(g Gomega) error {
```
flavio, line 124:
```suggestion
			// wait for the reconciliation loop of the ClusterAdmissionPolicy to remove the resource
			Eventually(func(g Gomega) error {
```

controllers/utils_test.go
```diff
@@ -134,6 +149,18 @@ func getTestPolicyServer(name string) (*policiesv1.PolicyServer, error) {
 	return &policyServer, nil
 }
 
+func getTestPolicyServerService(policyServerName string) (*corev1.Service, error) {
+	policyServer := policiesv1.PolicyServer{}
+	policyServer.Name = policyServerName
```
flavio, line 154: Can't we do that with a single line:

```golang
policyServer := policiesv1.PolicyServer{
	Name: policyServerName,
}
```

flavio (changes requested):
LGTM, thanks for the cleanup! I left some minor comments

controllers/policyserver_controller_test.go
```diff
@@ -53,15 +53,17 @@ var _ = Describe("PolicyServer controller", func() {
...
+		AfterEach(func() {
 			// It's necessary remove the test finalizer to make the
 			// BeforeEach work as extected. Otherwise, the policy service
 			// creation will not work as expected
 			policyServer, err := getTestPolicyServer(policyServerName)
 			Expect(err).Should(Succeed())
 			controllerutil.RemoveFinalizer(policyServer, IntegrationTestsFinalizer)
 			err = reconciler.Client.Update(ctx, policyServer)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ShouldNot(HaveOccurred())
```
flavio, line 66: This is a nitpick, in this case using `ToNot` leads to a correct English sentence: "expect err to not have occurred". While `ShouldNot` leads to something that isn't

flavio: Also, given this is now an after each block, we don't know what is actually done inside of the test -> we can't assume the deletion of the PolicyServer was requested.

The code should be something like:

```
err := deletePolicyServer()
expect(err).ToNot(HaveOccurred())

removeFinalizer(policyServer, IntegrationTestFinalizer)
err = update(policyServer)
expect(err).ToNot(HaveOccurred())

eventually(wait for PolicyServer to not be found)
```

flavio (commented):
LGTM, I left one last round of comments

controllers/policyserver_controller_test.go
jvanz: I don't think we need to do all of that. Notice that I've added the `AfterEach` on the `with no assigned policies` context. Not in the most external. If I've added it in the same level of the current `BeforeEach` other tests will failed. Because the policy server will be recreated on every `It` call (test/spec). But we do not want that. Because the tests under the `it has assigned policies` context need the state of the previous tests (`It` calls). 

However, I think that a eventually call is nice thing to add there.

flavio: I see, I take my comment back

flavio (approved):
LGTM, please clean up a little the history (I'm referring to the last 3 commit, they should be squashed together)

PR #683, opened 2024-03-18
Adding a pointer to sources.yaml reference

jhkrug:
to InsecureSources and SourceAuthorities.

The current docstrings were not explicit enough.


docs/crds/CRD-docs-for-docs-repo.md
```diff
@@ -7,6 +7,12 @@
 
 ## policies.kubewarden.io/v1
 
+This linter directive is added to avoid issues with the
```
viccuad, line 10: I'm not sure these phrases should be part of this file, if there's a linter that gives a false positive, we should disable it for the specific file with https://github.com/kubewarden/kubewarden-controller/blob/main/.golangci.yml#L11.

viccuad (approved):

Issue #692, opened 2024-03-25
Feature Request: allow `PolicyServer` to have Pod disruption budget


flavio:
### Is your feature request related to a problem?

This is a feature part of [this epic](https://github.com/kubewarden/kubewarden-controller/issues/564)

### Solution you'd like

Users must be allowed to specify [Disruption Budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for the Policy Server workloads.

The `PolicyServer` CRD must be extended to have the following new numerical attributes:
* `minAvailable`: which is a description of the number of pods from that set that must still be available after the eviction, even in the absence of the evicted pod. minAvailable can be either an absolute number or a percentage.
* `maxUnavailable`: which is a description of the number of pods from that set that can be unavailable after the eviction. It can be either an absolute number or a percentage.

Note, both fields are optional. According to [PodDisruptionBudgetSpec](https://github.com/kubernetes/kubernetes/blob/20d0ab7ae808aaddb1556c3c38ca0607663c50ac/staging/src/k8s.io/api/policy/v1/types.go#L28) they are both `omitempty`.

When set, the kubewarden-controller will create a [`v1/PodDisruptionBudget`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#poddisruptionbudget-v1-policy) object that has the attributed provided by the user. The `selector` attribute will be created by the controller to match the pods of the PolicyServer instance the CRD refers to.

Note: the PDB spec has a beta feature called `unhealthyPodEvictionPolicy`. This is currently in beta phase, we will not handle it during this first iteration.

### Admission Criteria

* [x] The PolicyServer CRD has two new attributes: `minAvailable` and `maxUnavailable`. Both are optional and have type `*intstr.IntOrString`. See the [`PodDisruptionBudgetSpec` definition](https://github.com/kubernetes/kubernetes/blob/20d0ab7ae808aaddb1556c3c38ca0607663c50ac/staging/src/k8s.io/api/policy/v1/types.go#L28) for details
* [x] When one of these new attributes is set, a PodDistruptionBudget object is created by the controller
* [x] When these fields are unset, the associated PodDisruptionBudget object is deleted by the controller
* [x] When none of these fields are set, there's no PodDisruptionBudget object associated with the PolicyServer
* [x] The PodDisruptionBudget targets the Pods of the specific PolicyServer instance that is being reconciled
* [x] Each PolicyServer instance can have a different PDB strategy
[...]

PR #698, opened 2024-03-27
feat: add PDB fields in the policy server spec.

jvanz:
## Description

Adds two new fields in the PolicyServerSpec: minAvailable and maxUnavailable. These fields are used to create a PodDisruptionBudget for the policy server pods. Both fields cannot be set together because Kubernetes does not allow setting both together in the PDB spec.

Fix https://github.com/kubewarden/kubewarden-controller/issues/692

## Test

```shell
make test
```
---

**NOTE**: this is a draft PR because I would like to make it "ready for review" together with the Helm chart changes. However, you can review it already.


controllers/policyserver_controller_test.go
```diff
@@ -18,130 +18,261 @@ package controllers
...
-				k8sClient.Create(ctx, clusterAdmissionPolicyFactory(policyName, policyServerName, false)),
+				k8sClient.Create(ctx, policyServer),
 			).To(haveSucceededOrAlreadyExisted())
+			// Wait for the Service associated with the PolicyServer to be created
+			Eventually(func(g Gomega) error {
+				_, err := getTestPolicyServer(policyServerName)
+				return err
+			}, timeout, pollInterval).Should(Succeed())
+			Eventually(func(g Gomega) error {
+				return policyServerPodDisruptionBudgetExists(policyServerName)
```
fabriziosestito, line 193: We could make the tests stricter by using a `getTestPDB` utility function and testing that the values are set correctly see https://github.com/kubewarden/kubewarden-controller/blob/37073dbfcce2130626775614a1fe4a27e3f39562/controllers/admissionpolicy_controller_test.go#L91 instead of asserting the existence of the PDB.

pkg/apis/policies/v1/policyserver_webhook.go
```diff
@@ -84,6 +84,11 @@ func (v *policyServerValidator) validate(ctx context.Context, obj runtime.Object
 		}
 	}
 
+	// Kubernetes does not allow to set both MinAvailable and MaxUnavailable at the same time
+	if policyServer.Spec.MinAvailable != nil && policyServer.Spec.MaxUnavailable != nil {
+		return fmt.Errorf("spec.MinAvailable and spec.MaxUnavailable cannot be set at the same time")
```
fabriziosestito, line 89: super nitpick: we could change this to "minAvailable and maxUnavailable cannot be both set" which is the same message k8s returns to the user in this scenario.

fabriziosestito (changes requested):

controllers/utils_test.go
```diff
@@ -220,3 +221,12 @@ func randStringRunes(n int) string {
 func newName(prefix string) string {
 	return fmt.Sprintf("%s-%s", prefix, randStringRunes(8))
 }
+
+func policyServerPodDisruptionBudgetExists(policyServerName string) error {
+	podDisruptionBudgetName := fmt.Sprintf("policy-server-%s-%s", policyServerName, "pdb")
```
flavio, line 226: nitpicking: I think we can omit `pdb` from the final PDB name, it's a bit redundant

flavio: Also, we have a function that computes that name: `buildPodDisruptionBudgetName`. It should be used

internal/pkg/admission/policy-server-pod-disruption-budget.go
```diff
@@ -0,0 +1,72 @@
...
+	err := client.IgnoreNotFound(k8s.Delete(ctx, pdb))
+	if err != nil {
+		err = errors.Join(fmt.Errorf("failed to delete PodDisruptionBudget"), err)
+	}
+	return err
+}
+
+func buildPodDisruptionBudgetName(policyServer *policiesv1.PolicyServer) string {
+	return fmt.Sprintf("%s-%s", policyServer.NameWithPrefix(), constants.PolicyServerPodDisruptionBudgetSuffix)
+}
```
flavio, line 40: I think we can just use the name of the policy-server to identify the PDB. The Policy Server name is unique, hence there's no risk of having clashes

internal/pkg/admission/policy-server-pod-disruption-budget.go
```diff
@@ -0,0 +1,72 @@
...
+				constants.PolicyServerLabelKey: policyServer.GetName(),
+			},
+		}
+		if policyServer.Spec.MinAvailable != nil {
+			pdb.Spec.MinAvailable = policyServer.Spec.MinAvailable
+		} else {
+			pdb.Spec.MaxUnavailable = policyServer.Spec.MaxUnavailable
+		}
+		return nil
+	})
```
flavio, line 67: The part where we set the ownerRef can be handled in a nicer way using [`controllerutil.SetOwnerReference`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/controller/controllerutil#SetOwnerReference).

For example:
```golang
secret := &corev1.Secret{
			TypeMeta: metav1.TypeMeta{
				Kind:       "Secret",
				APIVersion: "v1",
			},
				Namespace: app.ObjectMeta.Namespace,
				Name:      generatedRuntimeConfigSecretName,
				Labels: map[string]string{
					spinapp.NameLabelKey: app.ObjectMeta.Name,
				},
			},
			Data: map[string][]byte{
				"runtime-config.toml": tomlValue,
			},
		}
		err = controllerutil.SetOwnerReference(app, secret, r.Scheme)
		if err != nil {
			return fmt.Errorf("failed to set runtimeconfig owner reference: %w", err)
		}

		err = r.Client.Create(ctx, secret)
```

this is taken from [here](https://github.com/spinkube/spin-operator/blob/9ffa340b1e057a47f9fb8c2eaf8edf675f377db0/internal/controller/spinapp_controller.go#L252C3-L273C37)

internal/pkg/constants/constants.go
```diff
@@ -34,6 +34,9 @@ const (
 	PolicyServerVerificationConfigEntry         = "verification-config"
 	PolicyServerVerificationConfigContainerPath = "/verification"
 
+	// PolicyServer PodDisruptionBudget
+	PolicyServerPodDisruptionBudgetSuffix = "pdb"
```
flavio, line 38: Probably we can get rid of that if we name the PDB object using the Policy Server's name

controllers/policyserver_controller.go
```diff
@@ -57,6 +57,8 @@ type PolicyServerReconciler struct {
 //+kubebuilder:rbac:namespace=kubewarden,groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:namespace=kubewarden,groups=apps,resources=replicasets,verbs=get;list;watch
 //+kubebuilder:rbac:namespace=kubewarden,groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:namespace=kubewarden,groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
```
flavio, line 61: We don't need the second line, only the first one. The first one grants privileges against PDB objects inside of a namespace. The 2nd one grants provileges against PDB resources inside of the whol cluster. This is too wide.

flavio (changes requested):

flavio (changes requested):
I think the RBAC comment must be addressed. It would also be nice to address the other ones

controllers/policyserver_controller.go
jvanz: I'm not sure about this. I had to add this line to make the  controller able to access the PDB. But I'll double check that.

controllers/policyserver_controller_test.go
jvanz: I've pushed a gomega matcher to address this request. It's a little bit different from what you request. During the tests I found out that if we follow the example that you sent, the tests will fail after the first failed tried in the `Eventually` call. As far as I can see, the example test is not failing only because the order of the tests. In other words, it's passing because the order of the execution and due how the controller set policy status. The webhook always exists when the test start. To fix the test, we can follow the approach that I've used in this PR. Or we should update the `Expect` calls to be `g.Expect` (using the argument of the function.

jvanz: I'll open an issue to address this problem.

controllers/policyserver_controller.go
jvanz: Testing this I found that some roles in the helm charts does not correspond to the inline annotations heres. Let me try to fix them.

jvanz: I've fixed this. I forgot to add the PDB in the client cache configuration. That's why I need the clusterscope permission.

flavio (approved):
LGTM now. Thanks for the changes

fabriziosestito (approved):
LGTM. we can merge once the linter is happy

PR #704, opened 2024-04-03
feat: Use rbac from controller repo in the development environment

jvanz:
## Description

Updates the Tiltfile to change the Roles and ClusterRoles defined in the Helm charts to use the rules defined in the RBAC defined in the local directory. Therefore, when permissions are added,changed or removed, there is no need to copy the content to the Helm chart directory.

During this process I've found out that some kubebuilder directives does not match what we use in the Helm charts. Thus, I've updated them. Furthermore, for some reason, kubebuilder generates a duplicate rule in the manager-role role. Witch is causing the resultant role to not have create/update access to the deployments resource. A Kustomize patch to remove that rule has been added.  This problematic rule is not present in the Helm chart roles as well. 

This changes are written during my tests on #698 

## Test

```shell
tilt  up --stream
```

## Additional Information

I found weird that a duplicate rule of a RBAC overwrite a previous rule defined in the same role. As far as I can remember, this should not happen. But it was happening in my local cluster. Furthermore, the same problematic rule is remove in the Helm charts. Maybe this is can be a problem indeed. I've also tried to change the Kubebuilder directives to avoid that, but I could not find a way to workaround that there. If you know why that happen or another workaround, let me know.


viccuad (changes requested):
LGTM, nice changes!

On the CRDs, I think we should only include `policies.kubewarden.io_policyservers.yaml` and not the other formatting changes, as they may break the crds-to-docs.kubewarden.io workflow.

For the duplicated rule in manager-role, I can't reproduce from this PR with and without the new changes. Could it be that you have `./bin/controller-gen` != 0.8.0?

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
flavio, line 1: I think this should not be part of the PR

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
flavio, line 1: I think this should not be part of the PR

flavio (commented):
I'm conflicted. I understand why you would like to have tilt consume the RBAC rules from this repository, instead of the helm chart.

However there's also the fact that the real (as in the final one for our users) rules are the one from the helm chart. Hence it would make sense to not change the current behavior.

That's something that we can probably solve during one of our meetings (rather then endless conversations on GH)

viccuad (approved):

flavio (approved):
LGTM. I'm glad that you found what was creating the wrong RBAC rules. It's great to not have that kustomize thing around :partying_face:

viccuad:
rebased via UI as it was needed.

Issue #705, opened 2024-04-03
Wrong calls for the `Expect` function inside `Eventually` calls in the integration tests


jvanz:
In our integration tests we use the Ginkgo/Gomega `Eventually` function to wait for some resources to show up in the cluster. For example, in the following code, we wait for the webhooks and checks its values:

https://github.com/kubewarden/kubewarden-controller/blob/3fa9744d29d9a195a8ccff8e8802ce75f5254e94/controllers/clusteradmissionpolicy_controller_test.go#L63-L77

Even if we can use the `Eventually` for that, we are doing in the wrong way. All the `Expect` calls in that code should be `g.Expect`. Notice that we need to use `g Gomega` argument to call the function. This is necessary because if some of the execution of the polling function failed for some reason, like the resource is not there yet, `Eventually` will not retry and the test will fail. Therefore, we need to check out integration tests and see where we are doing this and fix it. 

As far as I can see, our tests are not failing all the time because the way the controller reconcile the resources, most of the times, the webhooks will be there when the test run. Furthemore, reviewing again [other](https://github.com/kubewarden/kubewarden-controller/issues/653) issues about [flaky tests](https://github.com/kubewarden/kubewarden-controller/issues/628), I believe this is the root cause of those issues. 

### Acceptance criteria

- Check all the integration tests and check all the `Evantually` where `Expect` is called not refering the `g` argument
- Fix all the wrong `Expect` call in the tests

fabriziosestito:
For reference: https://pkg.go.dev/github.com/onsi/gomega#Eventually

PR #708, opened 2024-04-05
feat: add limits and requests to PolicyServer

fabriziosestito:
## Description

#710 

Adds limits and requests field to the `PolicyServer` resource.
The fields are propagated to the PolicyServer deployment down to the running PolicyServer pods.

The PolicyServer webhook validates the fields and implements the default logic (if limits are specified and requests are left empty, the requests default to limits).

## Testing
Integration tests and unit tests added


pkg/apis/policies/v1/policyserver_webhook.go
```diff
@@ -103,3 +121,31 @@ func (v *policyServerValidator) ValidateUpdate(ctx context.Context, _, obj runti
...
+	for limitName, limitQuantity := range limits {
+		if limitQuantity.Cmp(resource.Quantity{}) < 0 {
+			return fmt.Errorf("%s limit must be greater than or equal to 0", limitName)
+		}
+	}
+
+	for requestName, requestQuantity := range requests {
+		if requestQuantity.Cmp(resource.Quantity{}) < 0 {
+			return fmt.Errorf("%s request must be greater than or equal to 0", requestName)
+		}
```
jvanz, line 139: Does make sense to have zero quantities? Furthermore, I think this can cause issues. For instance, I'm not sure if  Kubernetes allow defining memory with `0` quantity. It looks weird. Therefore, I suggest to not allow zero quantities neither.

controllers/policyserver_controller_test.go
```diff
@@ -283,4 +282,78 @@ var _ = Describe("PolicyServer controller", func() {
...
+			By("creating a deployment with limits and requests set")
+			Eventually(func(g Gomega) error {
+				deployment, err := getTestPolicyServerDeployment(policyServerName)
+				if err != nil {
+					return err
+				}
+
+				Expect(deployment.Spec.Template.Spec.Containers[0].Resources.Limits).To(Equal(policyServer.Spec.Limits))
+				By("setting the requests to the same value as the limits")
+				Expect(deployment.Spec.Template.Spec.Containers[0].Resources.Requests).To(Equal(policyServer.Spec.Limits))
```
jvanz, line 308:
```suggestion
				g.Expect(deployment.Spec.Template.Spec.Containers[0].Resources.Limits).To(Equal(policyServer.Spec.Limits))
				By("setting the requests to the same value as the limits")
				g.Expect(deployment.Spec.Template.Spec.Containers[0].Resources.Requests).To(Equal(policyServer.Spec.Limits))
```

Just to avoid [this](https://github.com/kubewarden/kubewarden-controller/issues/705)

controllers/policyserver_controller_test.go
```diff
@@ -283,4 +282,78 @@ var _ = Describe("PolicyServer controller", func() {
...
+
+			By("creating a pod with limit and request set")
+			Eventually(func(g Gomega) error {
+				pod, err := getTestPolicyServerPod(policyServerName)
+				if err != nil {
+					return err
+				}
+
+				Expect(pod.Spec.Containers[0].Resources.Limits).To(Equal(policyServer.Spec.Limits))
+				Expect(pod.Spec.Containers[0].Resources.Requests).To(Equal(policyServer.Spec.Limits))
```
jvanz, line 321:
```suggestion
				g.Expect(pod.Spec.Containers[0].Resources.Limits).To(Equal(policyServer.Spec.Limits))
				g.Expect(pod.Spec.Containers[0].Resources.Requests).To(Equal(policyServer.Spec.Limits))
```

controllers/utils_test.go
```diff
@@ -168,6 +169,36 @@ func getTestPolicyServerService(policyServerName string) (*corev1.Service, error
 	return &service, nil
 }
 
+func getTestPolicyServerDeployment(policyServerName string) (*appsv1.Deployment, error) {
+	policyServer := policiesv1.PolicyServer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: policyServerName,
+		},
+	}
+	deploymentName := policyServer.NameWithPrefix()
```
jvanz, line 178: I think we have bunched of places doing the same thing. Which signals me that this piece of code is a good candidate for a dedicated function. However, this is not a comment that should block merging this PR

internal/pkg/admission/policy-server-deployment.go
```diff
@@ -125,6 +125,7 @@ func shouldUpdatePolicyServerDeployment(policyServer *policiesv1.PolicyServer, o
 		originalDeployment.Spec.Template.Spec.SecurityContext != newDeployment.Spec.Template.Spec.SecurityContext ||
 		originalDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] != newDeployment.Annotations[constants.PolicyServerDeploymentConfigVersionAnnotation] ||
 		!reflect.DeepEqual(originalDeployment.Spec.Template.Spec.Containers[0].Env, newDeployment.Spec.Template.Spec.Containers[0].Env) ||
+		!reflect.DeepEqual(originalDeployment.Spec.Template.Spec.Containers[0].Resources, newDeployment.Spec.Template.Spec.Containers[0].Resources) ||
```
jvanz, line 128: This made me think... can we have 100% sure that the first container will be the policy server one? Cannot be the OPTEL collector sometimes?

If we are not sure about that, I think we should open an issue to investigate or maybe mitigate this.

jvanz (changes requested):

pkg/apis/policies/v1/policyserver_webhook.go
fabriziosestito: Weirdly enough,`0` is a valid value for Kubernetes in this context, it just means that no limits are set for the given resource

internal/pkg/admission/policy-server-deployment.go
fabriziosestito: Good call, I would defer the changes here as we are probably going to refactor this by using controller-runtime CreateOrUpdate utilities in 1.13

pkg/apis/policies/v1/policyserver_webhook_test.go
```diff
@@ -16,14 +16,41 @@ package v1
...
 )
 
+func TestDefault(t *testing.T) {
+	policyServer := &PolicyServer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "policy-server",
+		},
+		Spec: PolicyServerSpec{
+			Limits: corev1.ResourceList{
+				"cpu":    resource.MustParse("1Gi"),
```
viccuad, line 39: I suppose this is ok, but 1Gi CPU units sounds weird.

viccuad (approved):
LGTM

pkg/apis/policies/v1/policyserver_webhook_test.go
fabriziosestito: fixed in https://github.com/kubewarden/kubewarden-controller/pull/708/commits/836340188d0af4b387a419455b6d618e1886d979

flavio (approved):
LGTM

jvanz (approved):
As we discuss during the daily... LGTM

PR #713, opened 2024-04-09
fix: linter errors and remove requests defaults.

fabriziosestito:
## Description

There were some linter errors we did not catch since the Golangci action is acting funny.

For this reason, the PolicyServer controller integration tests were passing even if the expectation was wrong.
I've removed setting requests defaults to limits when the requests are empty to mimic the Kubernetes behavior, which only does this at the Pod level.


jvanz (approved):
LGTM. Just double check the failing tests.

fabriziosestito:
@jvanz I realized couple of things so I am moving this to in progress/draft

flavio (approved):
LGTM. It will be up to kubernets to set the `request` values later, when the Pod is being created. :+1:

PR #719, opened 2024-04-15
fix: flaky webhook nil labels/annotations test

fabriziosestito:
## Description

Fixes flaky test by merging the `It` blocks that were running in parallel,

See:  https://onsi.github.io/ginkgo/#mental-model-ginkgo-assumes-specs-are-independent


viccuad (approved):
Good find!

Issue #660, opened 2024-02-29
Increase the usage of `metadata.ownerReferences`


flavio:
Each `PolicyServer` resource owns a series of Kubernetes resources:

* Deployment
* Secret: to hold the certificates
* ConfigMap
* Service

Currently none of these resources are set to be owned by the PolicyServer instance that created them. Because of that, we have [this custom code](https://github.com/kubewarden/kubewarden-controller/blob/41946ee71bbe5de39c49984af073a636e28b21fd/internal/pkg/admission/reconciler.go#L41) that deals with the cleanup of these resources.

We can rely on Kubernetes' garbage collector to cleanup all these resources, removing code from our code base.

## Acceptance criteria

* [ ] Ownership relationships are set accordingly between a Policy Server and its sub-resources
* [ ] The Policy Server reconciliation loop is no longer in charge of cleaning up all the sub-resources. The controller only takes care of triggering the deletion of all the policies scheduled on the Policy Server instance and then, once no policy is scheduled, removes the `finalizer` from the Policy Server instance

jvanz:
@kubewarden/kubewarden-developers this is block due an issue in the upgrade process of the controller. As we discussed during one of our daily, I'll work on code to use the `createOrUpdate` helper function from the controller-runtime package to simplify our code and  workaround this.

jvanz:
I've rebased the PR for this issue on top of #730 . Moving this to block until we merge that PR.

PR #720, opened 2024-04-15
feat: use owner reference with policy server resources.

jvanz:
## Description

Updates the controller to set owner references for all the resources created to make the policy server works. Which includes: service, configmap, deployment and secret. Therefore, we can remove the reconciliation loop code to remove this resources when the user deletes the policy server and rely on Kubernetes garbage collector to do the job.

Fix https://github.com/kubewarden/kubewarden-controller/issues/660

## Test

```shell
make test
# manual testing with tilt
tilt up --stream
# install and uninstall policy server
```


controllers/policyserver_controller_test.go
```diff
@@ -137,16 +226,47 @@ var _ = Describe("PolicyServer controller", func() {
...
+				policyServer, err := getTestPolicyServer(policyServerName)
+				Expect(err).Should(Succeed())
+				controllerutil.RemoveFinalizer(policyServer, IntegrationTestsFinalizer)
+				err = reconciler.Client.Update(ctx, policyServer)
+				Expect(err).ToNot(HaveOccurred())
+
+				Eventually(func() error {
+					_, err := getTestPolicyServer(policyServerName)
+					return err
+				}, timeout, pollInterval).ShouldNot(Succeed())
```
flavio, line 249: Can you refactor that to make sure it fails with the error "not found"?

jvanz:
This PR is based on top of #730. Waiting for that PR to get merged before moving this to pending review.

jvanz:
@kubewarden/kubewarden-developers the integration tests are failing but I do not think this is related to the changes here. It looks like something different. I'm moving this PR to be reviewed by the team while I investigate the issue. It may take some time because I cannot reproduce this in my local branch. :crossed_fingers:

internal/pkg/admission/policy-server_integration_test.go
```diff
@@ -1,95 +0,0 @@
...
-	"testing"
-	"time"
-
-	"github.com/kubewarden/kubewarden-controller/internal/pkg/admissionregistration"
-	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
-)
-
-const port = "8181"
-
-func TestCAAndCertificateCreationInAHttpsServer(t *testing.T) {
```
viccuad, line 18: Should we drop this test?

viccuad (commented):
LGTM, just one question about the cert test refactor.

internal/pkg/admission/policy-server_integration_test.go
jvanz: Fair enough, I've re added this test in the internal/pkg/admission/policy-server-ca-secret_test.go file. ;)

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
```diff
@@ -163,11 +163,13 @@ spec:
                           items:
                             type: string
                           type: array
+                          x-kubernetes-list-type: atomic
```
flavio, line 166: Where is this coming from?

flavio (approved):
LGTM, I left one minor comment

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
jvanz: This came from a recent [update](https://github.com/kubernetes/apimachinery/commit/bfbbdecc1d6d161ce84da90c4a58e21c6eb8261f) in the apimachinery package. As we use label selector objects in the policy spec, the CRDs will be updated with the latest changes there as well. I've added a commit here to fix that in our main branch and avoid generating noise every time that we run the controller locally.  

If you think is better, I can open another PR just for that update in the CRDs.

PR #741, opened 2024-05-23
fix: cache configMap resources.

### LINKED ISSUE #645: Kubewarden controller potentially doing too many requests to apiserver

### Current Behavior

I noticed that compared to other operators, kubewarden controller seems to send a large amount of requests to the api server, probably to support reconciliation of the policy server. Based on information from my ElasticSearch instance here are some stats:
* 10 million logs in 24 hours coming from the kubewarden controller
* From those we have 2,5M that are `create` requests for `services, deployments and secrets` called `policy-server-default` that always fail with 409 because those resources already exist. They are all spread evenly so it seems kw controller is always trying to create these 3 resources constantly. Each of these (services, deployments and secrets) have 5 log entries per second, all with 409 status
* The big chunk are patch  operations on deployments and configmaps also called policy-server-default , which total 1.7M logs. Here we get 2-3 logs lines for each deployment and configmaps per second
* Next is update  on services called policy-server-default  that total 859K, these are more or less 1 per second
* Finally 5M of them are get configmaps, 4-5 log lines per second

I think there should be some room from improvement by slowing down some of these cycles because, in my opinion, it should be enough to send these requests every 2-3 seconds, which would decrease dramatically the load on the api-server. Again I don't have a deep knowledge about the reason to these cadence of requests, but maybe some of these request categories can be streamlined to remove some of the load on the apiserver.

Thanks

### Expected Behavior

I would exptect a lower load of requests on the apiserver

### Steps To Reproduce

Analyse logs from ElasticSearch coming from the apiserver - usually called audit logs

### Environment

```markdown
- OS: Linux
- Architecture: amd64
```

### Anything else?

Not really. Thanks

jvanz:
## Description

Updates the controller configuration to cache ConfigMap resources to avoid too many request to the API server during the reconciliation loops.

Fix #645

### DIFF
--- a/controllers/policyserver_controller_test.go
+++ b/controllers/policyserver_controller_test.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"      //nolint:revive
@@ -557,6 +558,72 @@ var _ = Describe("PolicyServer controller", func() {
 				))
 				return nil
 			}).Should(Succeed())
+		})
+
+		It("should set the configMap version as a deployment annotation", func() {
+			policyServer := policyServerFactory(policyServerName)
+			createPolicyServerAndWaitForItsService(policyServer)
+			configmap, err := getTestPolicyServerConfigMap(policyServerName)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() error {
+				deployment, err := getTestPolicyServerDeployment(policyServerName)
+				if err != nil {
+					return err
+				}
+				if deployment.GetAnnotations()[constants.PolicyServerDeploymentConfigVersionAnnotation] != configmap.GetResourceVersion() {
+					return errors.New("deployment configmap version did not change")
+				}
+				if deployment.Spec.Template.GetLabels()[constants.PolicyServerDeploymentConfigVersionAnnotation] != configmap.GetResourceVersion() {
+					return errors.New("pod configmap version did not change")
+				}
+				return nil
+			}, timeout, pollInterval).Should(Succeed())
+		})
+
+		It("should update the configMap version after adding a policy", func() {
+			policyServer := policyServerFactory(policyServerName)
+			createPolicyServerAndWaitForItsService(policyServer)
+			initalConfigMap, err := getTestPolicyServerConfigMap(policyServerName)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() error {
+				deployment, err := getTestPolicyServerDeployment(policyServerName)
+				if err != nil {
+					return err
+				}
+				if deployment.GetAnnotations()[constants.PolicyServerDeploymentConfigVersionAnnotation] != initalConfigMap.GetResourceVersion() {
+					return errors.New("deployment configmap version did not change")
+				}
+				if deployment.Spec.Template.GetLabels()[constants.PolicyServerDeploymentConfigVersionAnnotation] != initalConfigMap.GetResourceVersion() {
+					return errors.New("pod configmap version did not change")
+				}
+				return nil
+			}, timeout, pollInterval).Should(Succeed())
+
+			policyName := newName("validating-policy")
+			policy := clusterAdmissionPolicyFactory(policyName, policyServerName, false)
+			Expect(k8sClient.Create(ctx, policy)).To(Succeed())
+
+			Eventually(func() error {
+				configmap, err := getTestPolicyServerConfigMap(policyServerName)
+				if err != nil {
+					return err
+				}
+				if configmap.GetResourceVersion() == initalConfigMap.GetResourceVersion() {
+					return errors.New("configmap version did not change")
+				}
+				deployment, err := getTestPolicyServerDeployment(policyServerName)
+				if err != nil {
+					return err
+				}
+				if deployment.GetAnnotations()[constants.PolicyServerDeploymentConfigVersionAnnotation] != configmap.GetResourceVersion() {
+					return errors.New("deployment configmap version did not change")
+				}
+				if deployment.Spec.Template.GetLabels()[constants.PolicyServerDeploymentConfigVersionAnnotation] != configmap.GetResourceVersion() {
+					return errors.New("pod configmap version did not change")
+				}
+				return nil
+			}, timeout, pollInterval).Should(Succeed())
+
 		})
 
 	})
--- a/main.go
+++ b/main.go
@@ -165,16 +165,8 @@ func main() {
 				&corev1.Pod{}:                        namespaceSelector,
 				&corev1.Service{}:                    namespaceSelector,
 				&k8spoliciesv1.PodDisruptionBudget{}: namespaceSelector,
-			},
-		},
-		// These types of resources should never be cached because we need fresh
-		// data coming from the cliet. This is required to perform the rollout
-		// of the PolicyServer Deployment whenever a policy is added/changed/removed.
-		// Because of that, there's not need to scope these resources inside
-		// of the cache, like we did for Pods, Services,... right above.
-		Client: client.Options{
-			Cache: &client.CacheOptions{
-				DisableFor: []client.Object{&corev1.ConfigMap{}, &appsv1.Deployment{}},
+				&corev1.ConfigMap{}:                  namespaceSelector,
+				&appsv1.Deployment{}:                 namespaceSelector,
 			},
 		},
 	})


jvanz:
Ok, let me share some number in my test validating this change. I've tried to simulate some load as described in the issue. Therefore, I've created some clusters with the following audit policy:

```yaml
apiVersion: audit.k8s.io/v1 # This is required.
kind: Policy
# Don't generate audit events for all requests in RequestReceived stage.
# omitStages:
#   - "RequestReceived"
rules:
  # Log pod changes at RequestResponse level
  - level: Request
    resources:
    - group: ""
      resources: ["configmaps", "services", "secrets"]

  - level: Request
    resources:
    - group: "apps"
      resources: ["deployments"]
```

Than I've  wrote some script to exercise the Kuberwarden controller and generate the audit logs. They are not that fancy. They just create and delete policies multiple time while the audit log are collected:

```shell
#!/bin/bash

PROFILE=too-many-requests-$1

minikube delete --profile $PROFILE
minikube start \
	--profile $PROFILE \
	--extra-config=apiserver.audit-policy-file=/etc/ssl/certs/audit-policy.yaml \
	--extra-config=apiserver.audit-log-path=-

if [ -e "$1.log" ]; then
    rm "$1.log"
    echo "Removed $1.log"
fi

kubectl --context=$PROFILE logs --follow kube-apiserver-$PROFILE -n kube-system | grep audit.k8s.io/v1 &> $1.log &
pid=$!

helm --kube-context=$PROFILE upgrade --install --wait --namespace cert-manager --create-namespace --set installCRDs=true cert-manager jetstack/cert-manager
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --create-namespace kubewarden-crds kubewarden/kubewarden-crds
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --set image.tag="$1" --set image.repository="$2" kubewarden-controller kubewarden/kubewarden-controller
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --set "recommendedPolicies.enabled=True" kubewarden-defaults kubewarden/kubewarden-defaults

for i in $(seq 1 10); do
	echo "Iteration $i"
	kubectl --context=$PROFILE apply -f policy.yaml
	kubectl --context=$PROFILE apply -f policy2.yaml
	kubectl --context=$PROFILE apply -f policy3.yaml
	kubectl --context=$PROFILE apply -f policy4.yaml
	kubectl --context=$PROFILE apply -f policy5.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all

	kubectl --context=$PROFILE apply -f policy6.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all

	kubectl --context=$PROFILE delete --wait -f policy6.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all

	kubectl --context=$PROFILE delete --wait -f policy.yaml
	kubectl --context=$PROFILE delete --wait -f policy2.yaml
	kubectl --context=$PROFILE delete --wait -f policy3.yaml
	kubectl --context=$PROFILE delete --wait -f policy4.yaml
	kubectl --context=$PROFILE delete --wait -f policy5.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all
done

echo "Killing $pid"
kill $pid
wait
```

And the second one:

```shell
#!/bin/bash

PROFILE=too-many-requests-2-$1

minikube delete --profile $PROFILE
minikube start \
	--profile $PROFILE \
	--extra-config=apiserver.audit-policy-file=/etc/ssl/certs/audit-policy.yaml \
	--extra-config=apiserver.audit-log-path=-

if [ -e "$1-2.log" ]; then
    rm "$1-2.log"
    echo "Removed $1-2.log"
fi

kubectl --context=$PROFILE logs --follow kube-apiserver-$PROFILE -n kube-system | grep audit.k8s.io/v1 &> $1-2.log &
pid=$!

helm --kube-context=$PROFILE upgrade --install --wait --namespace cert-manager --create-namespace --set installCRDs=true cert-manager jetstack/cert-manager
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --create-namespace kubewarden-crds kubewarden/kubewarden-crds
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --set image.tag="$1" --set image.repository="$2" kubewarden-controller kubewarden/kubewarden-controller
helm --kube-context=$PROFILE upgrade --install --wait -n kubewarden --set "recommendedPolicies.enabled=True" kubewarden-defaults kubewarden/kubewarden-defaults

for i in $(seq 1 100); do
	echo "Iteration $i"
	kubectl --context=$PROFILE apply -f policy.yaml
	kubectl --context=$PROFILE apply -f policy2.yaml
	kubectl --context=$PROFILE apply -f policy3.yaml
	kubectl --context=$PROFILE apply -f policy4.yaml
	kubectl --context=$PROFILE apply -f policy5.yaml
	kubectl --context=$PROFILE apply -f policy6.yaml
	kubectl --context=$PROFILE apply -f policy7.yaml
	kubectl --context=$PROFILE apply -f policy8.yaml
	kubectl --context=$PROFILE apply -f policy9.yaml
	kubectl --context=$PROFILE apply -f policy10.yaml
	kubectl --context=$PROFILE apply -f policy11.yaml
	kubectl --context=$PROFILE apply -f policy12.yaml
	kubectl --context=$PROFILE apply -f policy13.yaml
	kubectl --context=$PROFILE apply -f policy14.yaml
	kubectl --context=$PROFILE apply -f policy15.yaml
	kubectl --context=$PROFILE apply -f policy16.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all

	kubectl --context=$PROFILE delete -f policy15.yaml
	kubectl --context=$PROFILE delete -f policy16.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all

	kubectl --context=$PROFILE delete -f policy.yaml
	kubectl --context=$PROFILE delete -f policy2.yaml
	kubectl --context=$PROFILE delete -f policy3.yaml
	kubectl --context=$PROFILE delete -f policy4.yaml
	kubectl --context=$PROFILE delete -f policy5.yaml
	kubectl --context=$PROFILE delete -f policy6.yaml
	kubectl --context=$PROFILE delete -f policy7.yaml
	kubectl --context=$PROFILE delete -f policy8.yaml
	kubectl --context=$PROFILE delete -f policy9.yaml
	kubectl --context=$PROFILE delete -f policy10.yaml
	kubectl --context=$PROFILE delete -f policy11.yaml
	kubectl --context=$PROFILE delete -f policy12.yaml
	kubectl --context=$PROFILE delete -f policy13.yaml
	kubectl --context=$PROFILE delete -f policy14.yaml
	kubectl --context=$PROFILE wait --timeout=5m --for=condition="PolicyUniquelyReachable" clusteradmissionpolicies --all
done

echo "Killing $pid"
kill $pid
wait
```

I've run both scripts with two versions. Kubewarden v1.12.0 and the code from this PR.  When I've counted the number of request on `configmap` resource the results are:

| Test     | v1.12.0 | PR     |
| -------- | ------- | ------ |
| test.sh  | 105611  | 52602  |
| test2.sh | 354951  | 249924 |


**NOTE**: I've counted the number of request using this command: `cat <log file generated>  | jq "select(.objectRef.namespace == \"kubewarden\" and .objectRef.resource == \"configmaps\") | .auditID" | wc -l`

We can see the an improvement. To improve even more, we need to change how we structure our reconciliation loops. Because of that, I've opened another issue to work on that in a future release.

controllers/policyserver_controller_test.go
```diff
@@ -559,6 +560,79 @@ var _ = Describe("PolicyServer controller", func() {
...
 		})
 
+		It("the deployment must have the configMap version as an annotation", func() {
+			policyServer := policyServerFactory(policyServerName)
+			createPolicyServerAndWaitForItsService(policyServer)
+			Eventually(func() error {
+				configmap, err := getTestPolicyServerConfigMap(policyServerName)
+				if err != nil {
+					return err
+				}
```
flavio, line 570: I think this could be moved right outside of the `eventually` block. The confimap is not supposed to change while this test runs, isn'it it?

controllers/policyserver_controller_test.go
```diff
@@ -559,6 +560,79 @@ var _ = Describe("PolicyServer controller", func() {
...
+		})
+
+		It("the deployment must have the configMap version update after adding a policy to the policy server", func() {
+			policyServer := policyServerFactory(policyServerName)
+			createPolicyServerAndWaitForItsService(policyServer)
+			Eventually(func() error {
+				configmap, err := getTestPolicyServerConfigMap(policyServerName)
+				if err != nil {
+					return err
+				}
```
flavio, line 592: move outside, and call it `initialConfigMap` or something like that. You can use it later on instead of `oldConfigMap`

flavio (commented):
I think the way you reproduced the issue is slightly different, because users complained about the amount of requests done at rest. Meaning, when absolutely no change is done to the policies or to other parts of the KW stack.
However, I think the numbers you got are impressive and I'm confident the changes you made will improve also the "idle scenario".

Now that the `ConfigMap` are cached, we have to rely on the speed of reaction of the reflectors used by the k8s' go-client. With that I mean the amount of time it passes between: a change is made to the ConfigMap, the controller receives the event notification from the API server.

I wonder if now the speed of reconciliation could be slightly worse. Let me be clear, I think there's no other way around that if we want to reduce the amount of requests done against the API server.
I just wonder if, given the current state of things, it wouldn't make sense to start caching also the Deployment objects.

controllers/policyserver_controller_test.go
```diff
@@ -559,6 +560,79 @@ var _ = Describe("PolicyServer controller", func() {
 			}).Should(Succeed())
 		})
 
+		It("the deployment must have the configMap version as an annotation", func() {
```
fabriziosestito, line 563: nit: `It("should set the configMap version as a deployment annotation")`

PR #747, opened 2024-05-28
Fix the local development with tilt

flavio:
Ensure the `PolicyReport` and `ClusterPolicyReport` CRDs are installed into the cluster.


fabriziosestito (approved):
Thanks!

PR #753, opened 2024-05-31
refactor: use only `CreateOrPatch`

flavio:
Consolidate the code to use only `CreateOrPatch` instead of a mix of that and `CreateOrUpdate`

Somehow related with https://github.com/kubewarden/helm-charts/issues/457


jvanz (changes requested):
LGTM. But I think you can update the kube-builder directives at kubewarden-controller/controllers/policyserver_controller.go to reflect this change as well.

Issue #775, opened 2024-06-18
Refactor `admissionregistration` package in `cert` 


fabriziosestito:
Rename `admissionregistration` package in `cert` and remove usage of `generateCertFunc` [inversion of control](https://github.com/kubewarden/kubewarden-controller/blob/030edce6f7c57d54b88453054e39f9baa20377cc/internal/pkg/admission/policy-server-ca-secret.go#L21) 

We can use this as a starting point: https://github.com/kubewarden/kubewarden-controller/tree/feat-ca-changes/internal/pkg/certificates

Issue #664, opened 2024-03-01
Refactor the `internal/pkg/admissionregistration` package


jvanz:
The  `/internal/pkg/admissionregistration` package is used by the controller to generate the CA and certificates. But it  has a misleading name and the tests for it is minimal. I think we should refactor this package renaming it and maybe bring the changes from the [feat-ca-changes](https://github.com/kubewarden/kubewarden-controller/tree/2d8d73e405ff18937e9457d86159e1fb5f258ffe/internal/pkg/certificates) branch into main. 

As the code in the branch already have tests, this will help us in our goal to reach 70% code coverage. Plus, its  name make more sense. Furthermore, we advance a little bit in the path of removing cert-manager. However,  this is **not** the main reason for this change here.

fabriziosestito:
@jvanz could you elaborate a bit more on which changes you are proposing to bring from the ca changes feature branch?

jvanz:
> @jvanz could you elaborate a bit more on which changes you are proposing to bring from the ca changes feature branch?

I was thinking of coping the [certificates](https://github.com/kubewarden/kubewarden-controller/tree/2d8d73e405ff18937e9457d86159e1fb5f258ffe/internal/pkg/certificates) package from the branch, which is basically a copy of the `admissionregistration` package with better names and tests, into main branch. Updating the references to the `admissionregistration` package in the `admission`  and using the new `certificates` package.  We may use the `admission` package from the feature branch to as [example](https://github.com/kubewarden/kubewarden-controller/blob/2d8d73e405ff18937e9457d86159e1fb5f258ffe/internal/pkg/admission/policy-server-ca-secret.go#L57-L61) of how to do it. But we just need to leave the code used to get and generate the root CA created by the controller out .

jvanz:
This is has been done at #777

PR #777, opened 2024-06-20
refactor: certificates generation package and secrets

fabriziosestito:
## Description
This PR refactors the generation utilities in the `admissionregistration` package,
removing duplication and simplifying the current implementation.
It renames `admissionregistration` in `certs` and reactors the existing reconciler unit test in integration tests clauses.

## Additional info

- This PR also changes the golangci configuration to add an exception for the `ok` map index in the `varnamelen` linter. However we should uniform and clean up the golangci configurations across all the KW repos, this is tracked by: #778 

- There are a few incongruences with the naming, for instance, this status  https://github.com/kubewarden/kubewarden-controller/blob/030edce6f7c57d54b88453054e39f9baa20377cc/pkg/apis/policies/v1/policyserver_types.go#L132 should be probably named `CertSecretReconciled` instead. Fixing this is out of scope for this PR and I will create an issue about it.

Fixes: #775, #664


controllers/utils_test.go
```diff
@@ -164,6 +164,14 @@ func getTestPolicyServerService(policyServerName string) (*corev1.Service, error
 	return &service, nil
 }
 
+func getTestPolicyServerCASecret() (*corev1.Secret, error) {
```
flavio, line 167: This is retrieving the Root CA created by our controller. I suggest to rename this helper function to reflect that.

Maybe something like `getInternalRootCASecret` ?

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -12,21 +12,18 @@ import (
...
+	"github.com/kubewarden/kubewarden-controller/internal/pkg/certs"
 	"github.com/kubewarden/kubewarden-controller/internal/pkg/constants"
 )
 
-type generateCAFunc = func() (*admissionregistration.CA, error)
-type pemEncodeCertificateFunc = func(certificate []byte) ([]byte, error)
-type generateCertFunc = func(ca []byte, commonName string, extraSANs []string, CAPrivateKey *rsa.PrivateKey) ([]byte, []byte, error)
-
-func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context, policyServer *policiesv1.PolicyServer, caSecret *corev1.Secret, generateCert generateCertFunc) error {
+func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context, policyServer *policiesv1.PolicyServer, caSecret *corev1.Secret) error {
```
flavio, line 19: Again, there's a bit of naming confusion here we inherited from the past.

This method creates the certificate used by a specific Policy Server instance, there's no CA associated with a Policy Server

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -73,37 +83,40 @@ func (r *Reconciler) fetchOrInitializePolicyServerCASecret(ctx context.Context,
...
-		return nil, fmt.Errorf("CA private key could not be extracted from secret %s", caSecret.Kind)
+		return nil, nil, fmt.Errorf("CA private key could not be extracted from secret %s", caSecret.Kind)
 	}
-	return &admissionregistration.CA{CaCert: caCert, CaPrivateKey: caPrivateKey}, nil
+
+	return caCert, privateKey, nil
 }
 
-func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Context, policyServer *policiesv1.PolicyServer, generateCA generateCAFunc, pemEncodeCertificate pemEncodeCertificateFunc) (*corev1.Secret, error) {
+func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Context, policyServer *policiesv1.PolicyServer) (*corev1.Secret, error) {
```
flavio, line 105: Same here, the naming is confusing since no CA is stored inside of the `Secret` that is returned

internal/pkg/admission/policy-server-ca-secret.go
```diff
@@ -115,29 +128,34 @@ func (r *Reconciler) fetchOrInitializePolicyServerCARootSecret(ctx context.Conte
...
 		&policyServer.Status.Conditions,
 		string(policiesv1.PolicyServerCARootSecretReconciled),
 	)
+
 	return policyServerSecret, nil
 }
 
-func updateSecretCA(policyServerSecret *corev1.Secret, generateCA generateCAFunc, pemEncodeCertificate pemEncodeCertificateFunc) error {
-	caRoot, err := generateCA()
+func updateCASecret(policyServerSecret *corev1.Secret) error {
```
flavio, line 140: This function is always generating the CA, it doesn't look into the contents of the secret that is passed as parameter.

Why don't we change the signature of the function to be something like that?

```golang
func createInternalCASecret() (policyServerSecret *corev1.Secret, error)
```

internal/pkg/admission/policy-server-ca-secret.go
flavio, line 1: Maybe something for later, but I would also rename this file since it can mislead into thinking each Policy Server has a dedicated CA

internal/pkg/certs/ca.go
```diff
@@ -10,21 +10,18 @@ import (
...
 )
 
-type CA struct {
-	CaCert       []byte
-	CaPrivateKey *rsa.PrivateKey
-}
-
-func GenerateCA() (*CA, error) {
-	privateKey, err := newPrivateKey(1024)
+func GenerateCA() ([]byte, *rsa.PrivateKey, error) {
```
flavio, line 13: Can you add some inline documentation explaining what is returned by the function?

internal/pkg/certs/cert.go
```diff
@@ -30,7 +28,7 @@ func GenerateCert(ca []byte,
 
 	// key size must be higher than 1024, otherwise the PolicyServer
 	// TLS acceptor will refuse to start
-	servingPrivateKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
```
flavio, line 31: Maybe we could create a constant defining the size of the key? This would be used to create the internal CA and the Policy Server certs.

internal/pkg/certs/cert.go
```diff
@@ -16,8 +14,8 @@ import (
 func GenerateCert(ca []byte,
```
flavio, line 14: Please add some inline docs

flavio (changes requested):
I think we should address some naming confusion about Policy Server CA, see my notes.

Overall LGTM

fabriziosestito:
@flavio I've addressed the comments in https://github.com/kubewarden/kubewarden-controller/pull/777/commits/c3cbdf4e9759e6e0fa2900d2343bcde0865cfe43. 

I tracked all the naming-related comments in: https://github.com/kubewarden/kubewarden-controller/issues/782

jvanz (approved):

fabriziosestito:
Blocked: waiting for 1.14 release

flavio (approved):
Approving, the naming changes I've suggested are going to be done inside of future PRs

PR #779, opened 2024-06-21
fix: its not necessary add the finalizer in the reconciliation loop.

jvanz:
## Description

The finalizer used by the Kubewarden controller is added in the webhook for the policy server and policies types. Therefore, it not needed to add it in the reconciliation loop. Furthermore, the deletion reconciliation is able to handle old finalizer as well. So, this change will not cause issues when deleting resources.


flavio (approved):
LGTM

viccuad (approved):
LGTM

flavio (approved):
LGTM

Issue #773, opened 2024-06-18
Split the `Reconciler` struct


fabriziosestito:
Split the Reconciler god class in internal/admission. 

This is not needed as each controller is already a reconciler. 

We can implement the reconciler logic inside the existing controllers (multiple file splitting is allowed when we want to manage the source file size). 

See https://github.com/Azure/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/ (note: the project structure in this controller is old). 

Policy common logic could be packaged as a set of utils functions or a dedicated client.

PR #785, opened 2024-06-26
refactor: policy server reconciler

fabriziosestito:
## Description
This PR refactors the `PolicyServerReconciler` to encapsulate all the logic needed to reconcile a policy server. 
This means moving the related methods and utility functions from the `admission.Reconciler` "god class" to the `PolicyServerReconciler` controller.

It also removes the unit tests, superseded by the ginkgo-based integration tests. This drops the coverage by 1.70%, but re-adding the missing paths to the integration suite will be possible later.

Part of #773


viccuad (approved):
LGTM! Another great refactor!

internal/controller/suite_test.go
```diff
@@ -137,9 +137,9 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 
 	err = (&PolicyServerReconciler{
-		Client:     k8sManager.GetClient(),
```
viccuad, line 140: (opening thread randomly here)

I understand dropping the unit tests, but I'm still sad about it. Any possibility to amend them in this PR instead?
If not, we should open an issue to track it.

PR #793, opened 2024-07-02
feat: policy server deployment tolerations.

### LINKED ISSUE #752: Extend PolicyServer CRD to have tolerations

Currently it's not possible to have the Policy Server Pods tolerate taints on the nodes. All the other pieces of the Kubewarden stack can achieve that, except for Policy Server.

### Solution you'd like

Extend the PolicyServer CRD to have a list of [`Toleration`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#toleration-v1-core) objects.

The `tolerations` attribute will then be copied into the Policy Server Deployment under  `.spec.template.spec.tolerations`

_No response_

### Anything else?

_No response_

jvanz:
## Description

Updates the PolicyServer CRDs to allow users to define the toleration to be used in the policy server deployment.

Fix #752 

## Test

```shell
make test
```

## Additional Information

I'm aware that we have a PR changing our tests (#788 ). I'm fine to wait that PR to be merged before. I'm glad to rebase this change on top of that PR after it gets merged into main.


internal/controller/policyserver_controller_deployment.go
```diff
@@ -168,6 +168,7 @@ func (r *PolicyServerReconciler) updatePolicyServerDeployment(policyServer *poli
 				SecurityContext:    podSecurityContext,
 				Containers:         []corev1.Container{admissionContainer},
 				ServiceAccountName: policyServer.Spec.ServiceAccountName,
+				Tolerations:        policyServer.Spec.Tolerations,
```
viccuad, line 171: Just in case, shouldn't it be done with something similar to https://github.com/kubewarden/kubewarden-controller/blob/main/internal/controller/policyserver_controller_deployment.go#L299-L300?

That way, we support the users changing the Deployment if there was no PolicyServer spec.tolerations set.

flavio: @viccuad  I don't understand why we wrote the code you linked. I would have handled affinity like @jvanz just did here with tolerations...

@fabriziosestito can you chime in?

jvanz: Is not related to how we used to detect changes in the resource and trigger a reconciliation? I have a vague memory about a discussion around this. However, as now we use the controller-runtime helper functions, I believe, we do not need to do that anymore.

fabriziosestito: We can remove it and use the same approach of the Tolerations field. As far as I  understood the `if` block does not change the reconciliation behaviour, it's probably some leftover since now we use the controller-runtime `CreatOr*`
> That way, we support the users changing the Deployment if there was no PolicyServer spec.tolerations set.
I don't think the user should be able to thinker with the Deployment, as the only interface we expose is the PolicyServer CRD

fabriziosestito (approved):

internal/controller/policyserver_controller_deployment.go
viccuad: If we agree then, I would like to follow the same approach of Tolerations for Affinity in this PR.

viccuad (approved):
LGTM!

fabriziosestito (approved):

PR #800, opened 2024-07-08
build: add ginkgo dep to makefile

fabriziosestito:
## Description
Adds the ginkgo cli tool dependency to the Makefile, similarly to what Kubebuilder already scaffolds for other external tools.

I've tried to cache the `bin` directory, but restoring the cache takes about the same time as installing the tools from scratch.


jvanz (approved):

viccuad (approved):

flavio (approved):
ship it :rocket:

Issue #778, opened 2024-06-21
Uniform golangci configuration across all repositories


fabriziosestito:
### Description

We need to uniform the golangci configuration across all repositories.
An idea could be using an opinionated configuration, such as: https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322, which is actvitely maintained and it is used by: 

https://github.com/spinkube/runtime-class-manager/blob/main/.golangci.yaml

This also means putting some extra effort to make the linters happy.

flavio:
I'm in favor of doing that. :+1:

PR #807, opened 2024-07-11
feat: use stricter go linter configuration.

jvanz:
## Description

Updates the Go lang linter to stricter configuration. Also, updates the controller code to make the linter happy again.

Fix https://github.com/kubewarden/kubewarden-controller/issues/778

## Test

```shell
make lint
```

## Additional Information

I've removed some `exclude-rules` not used from the suggested [configuration file](https://github.com/spinkube/runtime-class-manager/blob/main/.golangci.yaml) to be used. As well as, disabled the `testpackage` linter. I don't think we need it for now.

Keep in mind that some changed lines are `go fmt` output. ;)


flavio (approved):
Not bad! I think it looks nice

.golangci.yml
```diff
@@ -1,84 +1,304 @@
-# This file contains all available configuration options
-# with their default values.
-
-# options for analysis running
 run:
-  tests: true
-  timeout: 10m
+  # Timeout for analysis, e.g. 30s, 5m.
```
fabriziosestito, line 2: Could we base the config on the original one instead 
https://gist.githubusercontent.com/maratori/47a4d00457a92aa426dbd48a18776322/raw/309ab85402cd3e8b6e4a9a515a57c6baa8082108/.golangci.yml

and leave the header comment so we know where this comes from?

This needs updating golangci version to 1.59.1 as the config was updated for this golangci version.

.golangci.yml
```diff
@@ -1,84 +1,304 @@
...
+    - testifylint # checks usage of github.com/stretchr/testify
+    - tparallel # detects inappropriate usage of t.Parallel() method in your Go test codes
+    - unconvert # removes unnecessary type conversions
+    - unparam # reports unused function parameters
+    - usestdlibvars # detects the possibility to use variables/constants from the Go standard library
+    - wastedassign # finds wasted assignment statements
+    - whitespace # detects leading and trailing whitespace
+
+    ## you may want to enable
+    # - testpackage # makes you use a separate _test package
```
fabriziosestito, line 220: We could write a commit that we (kubewarden) disabled this, diverging from the original config.

cmd/main.go
```diff
@@ -62,6 +61,7 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+//nolint:funlen // Avoid splitting the main function in multiple functions to avoid changing the retcode logic for metrics shutdown
```
fabriziosestito, line 64: I am a bit afraid of the `funlen` linter in the long term. 
We often are forced to disable it or create many functions to make it happy, where it would be simpler to have a single function for readability.
Should we keep it around?

fabriziosestito (approved):
Great work!
A couple of comments added

cmd/main.go
flavio: let's see how often it gets in our way. We can then disable on a per-function basis. Eventually we can completely turn it off

viccuad: I'm ok with punting on it.

viccuad (approved):
LGTM modulo Fabrizio's comments. I'm ok on punting on the funlen,.

jvanz:
@fabriziosestito @flavio @viccuad can you review it again? I've noticed that the golangci.yml file that I've added in this PR was not the latest one. I've copied the version from spinkube repository. 

Now, I've copied the one from Marat Reymers gist and fix everything new that shown up.

viccuad (approved):
LGTM. Although I find weird to go with an allow-list for linters instead of deny-list.

fabriziosestito (approved):
Great work! 
Love the approach of excluding linters in the configuration file for tests, but keeping the explicit nolint commits for the controller code with an explanation,

flavio (approved):
I like the change. I'm fine having an allow list, since the configuration of golangci changes A LOT between release, causing quite some problems when new linters are added out of the blue

flavio (approved):
ship it :rocket:

PR #812, opened 2024-07-16
fix: remove default policy server name CLI flag.

### LINKED ISSUE #809: Consider removing the --default-policy-server flag

The `default-policy-server` sets the default policy server for policies that do not have the policy server field set.

### Solution you'd like

Even if we expose this flag through the [helm chart](https://github.com/kubewarden/helm-charts/blob/b88ba72948f4796c156639430393dea8b71d369e/charts/kubewarden-controller/values.yaml#L104), it is probably not used. We could remove the flag and pin the default policy server name to `default`.

Instead of removing it, provide a default value here: https://github.com/kubewarden/kubewarden-controller/blob/b22a2439068843f0ff425164f0330dc5f77787ea/cmd/main.go#L87

### Anything else?

We need to update the helm chart accordingly and document the breaking change

jvanz:
## Description

Updates the CLI flag to allow users to define the default policy server name to be used when users does not define one in policies. Now the default policy server name will be always: "default"

Fix #809

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -85,7 +85,6 @@ func main() {
 	flag.BoolVar(&enableTracing, "enable-tracing", false,
 		"Enable tracing collection for all Policy Servers")
 	flag.StringVar(&openTelemetryEndpoint, "opentelemetry-endpoint", "127.0.0.1:4317", "The OpenTelemetry connection endpoint")
-	flag.StringVar(&constants.DefaultPolicyServer, "default-policy-server", "", "The default policy server to set on policies before they are persisted")
 	flag.StringVar(&deploymentsNamespace,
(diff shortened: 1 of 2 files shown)


viccuad (approved):
LGTM, there's no other places in the controller where this gets used.

fabriziosestito (approved):

viccuad:
I would like to merge this at the same time that we change the helm-charts to not depend on this flag.

flavio (approved):
LGTM, let's not forget to update the helm chart too

jvanz:
@kubewarden/kubewarden-developers  https://github.com/kubewarden/helm-charts/pull/484

Issue #758, opened 2024-06-04
Feature Request: expose webhook `matchConditions`


flavio:
### Is your feature request related to a problem?

Starting from kubernetes 1.26, CEL support has been added into Kubernetes. On top of [`ValidatingAdmissionPolicy`](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) the `ValidatingWebhookConfiguration` and the `MutatingWebhookConfiguration` CRDs obtained a new field: `matchConditions`.

The `matchConditions` field allows to specify a list of CEL expressions that are used to determine whether the webhook endpoint should be called or not.

This is extremely powerful, since it allows to do something like "do not call this webhook endpoint if `!object.spec.containers[0].image.startsWith("registry.local.lan")`.

More information can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions).

### Solution you'd like

The `AdmissionPolicy` and the `ClusterAdmissionPolicy` CRDs should be updated to feature a `matchConditions` field.

The kubewarden-controller should then propagate them into the `(Validating|Mutating)WebhookConfiguration)` objects.

### Anything else?

The `matchConditions` field has been introduced starting with Kubernetes 1.27 (see [here](https://v1-27.docs.kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)).

Currently (Jun 2024), Kubernetes 1.26 has been [EOL](https://endoflife.date/kubernetes) for ~ 3 months. That means all the maintained Kubernetes versions have this feature available behind a feature gate:

> Note: Use of matchConditions requires the [featuregate](https://v1-27.docs.kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)AdmissionWebhookMatchConditions to be explicitly enabled on the kube-apiserver before this feature can be used.

Starting from Kubernetes 1.30 the feature gate is gone, since this feature graduated to stable.

I think our CRDs should always have the `matchConditions` field. We have to figure out what the controller should do when running on Kubernetes >= 1.27 < 1.30 and the `AdmissionWebhookMatchConditions` feature gate is disabled.

I suspect in this condition the `(Validating|Mutating)WebhookConfiguration` resources would still have this field, but it would not have any effect. I hope that's the case, because it would allow our controller to always propagate the `matchConditions` field, without having to perform any kind of check.

[...]

PR #814, opened 2024-07-17
feat: Add new `spec.matchConditions` to (Cluster)AdmissionPolicies

viccuad:
## Description

Relates to https://github.com/kubewarden/kubewarden-controller/issues/758.

Add new optional `spec.matchConditions` to (Cluster)AdmissionPolicies. This field is propagated to the underlying {Validating, Mutating}Webhooks only if the feature gate `CheckAdmissionWebhookMatchConditions` is enabled in the cluser (by default enabled in k8s >= 1.30).
`spec.matchConditions` expects a CEL expression, analogous to ValidatingAdmissionPolicies.

Regenerate CRDs and CRD docs.

## Test

Check that we set the webhook's MatchConditions{}. Default the reconcilers to assume that the feature gate is enabled. Given that we default to `K3S_TESTCONTAINER_VERSION = v1.30.0-k3s1`,
this checks for the happy path only.

Testing CheckFeatureGateAdmissionWebhookMatchConditions is not trivial, As we can't make use of the client config in the SynchronizedBeforeSuite() before connecting to the environment, to configure the reconcilers.

### Potential improvement

Enhance the `featuregates` package to query for other feature gates or CRD versions, and save the results in a struct.


jvanz (commented):
Overall, LGTM. But I have mixing feeling about how to check if the feature gate is enabled it or not. I'm wondering if having [multiples CRDs](https://book.kubebuilder.io/reference/generating-crd.html?highlight=minimal%20version#multiple-versions) version for different K8s versions is not a better approach.  However, I'm not sure about that either. Maybe  It can make maintenance harder.

internal/controller/policy_subreconciler_webhook.go
```diff
@@ -77,6 +77,9 @@ func (r *policySubReconciler) reconcileValidatingWebhookConfiguration(
 				AdmissionReviewVersions: []string{"v1"},
 			},
 		}
+		if r.featureGateAdmissionWebhookMatchConditions {
```
viccuad, line 80: I could add something like:

```golang
if !r.featureGateAdmissionWebhookMatchConditions && (policy.GetMatchConditions().Size() > 0) {
    log.Info("Skipping setting matchConditions for policy {} as the feature gate is not enabled", policy.GetName())
}
```

Does it seem worth it?

viccuad:
I would not go with multiple CRDs, seems indeed a maintenance hell as one cannot migrate to different CRDs easily, just upgrade them.

internal/controller/policy_subreconciler_webhook.go
flavio: Yes, I think it would be nice to show this information to the user. Just go with a `if found { ... } else { info }`

internal/controller/policy_subreconciler_webhook.go
```diff
@@ -162,6 +165,9 @@ func (r *policySubReconciler) reconcileMutatingWebhookConfiguration(
...
+			webhook.Webhooks[0].MatchConditions = policy.GetMatchConditions()
+		}
```
flavio, line 170: Same here, we should inform the user when this propagation cannot be done due to the missing feature

flavio:
I would not go with multiple CRDs too. Keep in mind 1.30 has this feature enabled by default, we have to keep around the code that determines whether the feature gate is enabled until 1.29 reaches EOL. This is something that will happen [28th Feb 2025](https://endoflife.date/kubernetes), not so far in the future

internal/controller/policy_subreconciler_webhook.go
viccuad: done in https://github.com/kubewarden/kubewarden-controller/pull/814/commits/2aa2120f4a30e93fcf9c8300601693b351a01bbe.

internal/controller/policy_subreconciler_webhook.go
viccuad: Done in https://github.com/kubewarden/kubewarden-controller/pull/814/commits/2aa2120f4a30e93fcf9c8300601693b351a01bbe.

internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,47 @@
...
+// CheckAdmissionWebhookMatchConditions returns true if the feature gate
+// AdmissionWebhookMatchConditions is activated. It does this by fetching the
+// OpenAPIV3 schema from the discovery client and checking for the feature
+// gate. This feature is stable since Kubernetes v1.30.
+func CheckAdmissionWebhookMatchConditions(config *rest.Config) (bool, error) {
+	// Obtain openAPIV3 client from discoveryClient
+	apiClient := discovery.NewDiscoveryClientForConfigOrDie(config).OpenAPIV3()
+	paths, err := apiClient.Paths()
+	if err != nil {
+		log.Fatalf("Failed to fetch OpenAPI spec: %v", err)
```
fabriziosestito, line 23: why `log.Fatalf` here? couldn't we just return an error?

viccuad: True, rebased and changed.

viccuad:
rebased onto main as needed.

fabriziosestito (changes requested):
Overall, it looks great, I've just added a couple of comments.

As an improvement, we might want to validate `matchConditions` in the CRDs webhooks.
If for example I try to apply a VAP with a syntax error in a matchCondition, kubernetes rejects the request with:

```
The ValidatingAdmissionPolicy "demo-policy.example.com" is invalid: spec.matchConditions[2].expression: Invalid value: "$$$$": compilation failed: ERROR: <input>:1:1: Syntax error: token recognition error at: '$'
```

We could spawn a spike to see if we can reuse/import kubenetes packages to deal with the validation.

internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,47 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// CheckAdmissionWebhookMatchConditions returns true if the feature gate
+// AdmissionWebhookMatchConditions is activated. It does this by fetching the
+// OpenAPIV3 schema from the discovery client and checking for the feature
+// gate. This feature is stable since Kubernetes v1.30.
+func CheckAdmissionWebhookMatchConditions(config *rest.Config) (bool, error) {
+	// Obtain openAPIV3 client from discoveryClient
+	apiClient := discovery.NewDiscoveryClientForConfigOrDie(config).OpenAPIV3()
```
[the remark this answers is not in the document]
viccuad: True. I wonder how that implementation would look to be able to fake the schemas.
As it is right now, we can't make use of the client config in the [SynchronizedBeforeSuite()](https://github.com/kubewarden/kubewarden-controller/blob/main/internal/controller/suite_test.go#L64) before connecting to the environment, to configure the reconcilers. We would need a bit of refactoring for just testing the OpenAPI query, which I found overkill.

internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch openapi schema for %s: %w", resourcePath, err)
+	}
+	var parsedV3Schema map[string]interface{}
+	if err = json.Unmarshal(openAPISchemaBytes, &parsedV3Schema); err != nil {
+		return false, fmt.Errorf("failed to unmarshal openapi schema for %s: %w", resourcePath, err)
+	}
+	schemas := parsedV3Schema["components"].(map[string]interface{})["schemas"]
+	validatingWebhook := schemas.(map[string]interface{})["io.k8s.api.admissionregistration.v1.ValidatingWebhook"]
+	_, exists = validatingWebhook.(map[string]interface{})["properties"].(map[string]interface{})["matchConditions"]
```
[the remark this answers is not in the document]
viccuad: I find it overkill, we will always have `io.k8s.api.admissionregistration.v1.ValidatingWebhook.properties` in the schema, and if not, we will find it anyways.
We could cast the matchConditions if they exist, but I wouldn't start doubting the k8s API. And we pass the matchConditions struct on our own CRD, so it would keep working nevertheless.

viccuad:
> As an improvement, we might want to validate matchConditions in the CRDs webhooks.

Opened https://github.com/kubewarden/kubewarden-controller/issues/815 and looking into it. I'm on the fence on merging this PR without CEL validation in place.

viccuad:
Moving this to blocked, as we feel we should provide with https://github.com/kubewarden/kubewarden-controller/issues/815 before exposing it to users.

PR #837, opened 2024-08-05
feat: policy group

### LINKED ISSUE #833: Add PolicyGroups CRD to kubewarden-controller
Updates the Kubewarden controller adding the new CRDs for the [PolicyGroups](https://github.com/kubewarden/rfc/blob/main/rfc/0020-policy-group.md). 

#### Acceptance criteria
- Define the CRDs Go types.
- Implement the reconciliation of the new CRDs, make sure we can reuse as much as possible of the sub-reconciler we already have in place.

jvanz:
## Description

Adds the CRDs definition for the `ClusterAdmissionPolicyGroup` and `AdmissionPolicyGroup`. These two new policies kinds will allows users to group policies together and make them work as a unified policy. Also adds the new controllers to reconcile the policy groups CRDs. These controllers reuse the same code from the previous policies controllers.

To allow developers to run the policy groups controller is necessary to use the latest, yet to be released, policy server. As well as update the yaml manifest used to deploy the controller to include the RBAC rules to access the new resource kinds.

Fix #833 

## Test

To test this feature, it's possible to use tilt:

```console
tilt up --stream
```

Remember this feature requires the latest policy server (not released yet):

```console
kubectl --context minikube apply -f - <<EOF
kind: PolicyServer
metadata:
  name: default
spec:
  image: ghcr.io/kubewarden/policy-server:latest
  replicas: 1
  env:
  - name: KUBEWARDEN_LOG_LEVEL
    value: debug
EOF
```

These are some policies of all the types that can be used to test the controllers:

```yaml 
kind: ClusterAdmissionPolicyGroup
metadata:
  name: cluster-policy-group-hd83ybjz
spec:
  policyServer: default
  rules:
    - apiGroups: [""]
      apiVersions: ["v1"]
      resources: ["pods"]
      operations:
        - CREATE
        - UPDATE
  backgroundAudit: true
  policies:
    - name: pod-privileged
      url: ghcr.io/kubewarden/policies/pod-privileged:v0.3.1
      settings: {}
      contextAwareResources:
        - apiVersion: "v1"
          kind: "Pod"

  expression: "pod-privileged()"
  message: "The policy group is rejected."
---
kind: AdmissionPolicyGroup 
metadata:
  name: namespace-policy-group-hd83ybjz
  namespace: default
spec:
  policyServer: "default"
  rules:
    - apiGroups: [""]
      apiVersions: ["v1"]
      resources: ["pods"]
      operations:
        - CREATE
        - UPDATE
  backgroundAudit: true
  policies:
    - name: pod-privileged
      url: ghcr.io/kubewarden/policies/pod-privileged:v0.3.1
      settings: {}
      contextAwareResources:
        - apiVersion: "v1"
          kind: "Pod"

  expression: "pod-privileged()"
  message: "The policy group is rejected."
---
kind: ClusterAdmissionPolicy
metadata:
  name: pod-privileged
  namespace: kubewarden
spec:
  module: "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.2.7"
  policyServer: default
  settings: {}
  mode: monitor
  rules:
    - apiGroups: ["apps"]
      apiVersions: ["v1"]
      resources: ["deployment"]
      operations:
        - CREATE
        - UPDATE
  mutating: false
---
kind: AdmissionPolicy
metadata:
  name: pod-privileged
  namespace: kubewarden
spec:
  module: "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.2.7"
  settings: {}
  mode: monitor
  policyServer: default
  rules:
    - apiGroups: ["apps"]
      apiVersions: ["v1"]
      resources: ["deployment"]
      operations:
        - CREATE
        - UPDATE
  mutating: false
```

The `policies.yml` data used in the default policy server can be fetcher with:

```console
kubectl get configmap -n kubewarden policy-server-default -o json  | jq -r ".data.\"policies.yml\""
```


api/policies/v1/admissionpolicygroup_types.go
```diff
@@ -0,0 +1,198 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// AdmissionPolicySpec defines the desired state of AdmissionPolicy.
+type AdmissionPolicyGroupSpec struct {
+	PolicyGroupSpec `json:""`
+}
+
+// AdmissionPolicy is the Schema for the admissionpolicies API
```
viccuad, line 30:
```suggestion
// AdmissionPolicyGroup is the Schema for the admissionpolicyGroups API
```

internal/controller/admissionpolicygroup_controller_test.go
flavio, line 1: I haven't double checked (sorry!), but I feel this test file is extremely similar to the one of the `admissionpolicy_controller_test.go`.

Would it make sense to move the core of the test into something generic and reusable?

internal/controller/clusteradmissionpolicygroup_controller.go
```diff
@@ -0,0 +1,113 @@
...
+	return findClusterAdmissionPoliciesForPod(ctx, r.Client, object)
+}
+
+func (r *ClusterAdmissionPolicyGroupReconciler) findClusterAdmissionPoliciesForPolicyServer(ctx context.Context, object client.Object) []reconcile.Request {
+	return findClusterAdmissionPoliciesForPolicyServer(ctx, r.Client, object, r.DeploymentsNamespace)
+}
+
+func (r *ClusterAdmissionPolicyGroupReconciler) findClusterAdmissionPolicyForWebhookConfiguration(_ context.Context, webhookConfiguration client.Object) []reconcile.Request {
+	return findClusterAdmissionPolicyForWebhookConfiguration(webhookConfiguration, r.Log)
+}
```
flavio, line 113: Same here

flavio (changes requested):
wow, that's quite a change. Everything looks good, I left some comments

api/policies/v1/policy.go
```diff
@@ -111,4 +111,8 @@ type Policy interface {
 	GetCategory() (string, bool)
 	GetTitle() (string, bool)
 	GetDescription() (string, bool)
+	IsPolicyGroup() bool
+	GetPolicyMembers() []PolicyGroupMember
+	GetExpression() string
+	GetMessage() string
```
fabriziosestito, line 117: This works fine for the scope of this PR
However, we might want to open an issue as I think we are reaching the limit of this interface. 
To clean up a bit we could use interface composition:
```
type AdmissionPolicy interface {
    GetSideEffects() *admissionregistrationv1.SideEffectClass
    GetRules() []admissionregistrationv1.RuleWithOperations
    GetFailurePolicy() *admissionregistrationv1.FailurePolicyType
    GetMatchPolicy() *admissionregistrationv1.MatchPolicyType
    GetMatchConditions() []admissionregistrationv1.MatchCondition
    GetTimeoutSeconds() *int32
}

type ContextAwarePolicy interface {
    IsContextAware() bool
    GetContextAwareResources() []ContextAwareResource
}

type Policy interface {
    client.Object
    PolicyModeHandler
    MetaDataHandler
    AdmissionPolicy
    ContextAwarePolicy
    // Other composed interfaces as needed
}


type PolicyGroup interface {
    Policy
    GetPolicyMembers() []PolicyGroupMember
    GetExpression() string
    GetMessage() string
}
```

And refactor the reconciler code accordingly.
Then we could cast detect if the policy that we are reconciling is a group:

```
func ProcessPolicy(p Policy) {
	// Check if the policy also implements the PolicyGroup interface
	if pg, ok := p.(PolicyGroup); ok {
		fmt.Println("This is a PolicyGroup")
		fmt.Println("Expression:", pg.GetExpression())
		fmt.Println("Message:", pg.GetMessage())
	} else {
		fmt.Println("This is a regular Policy")
	}
}
```

fabriziosestito (commented):
Overall LGTM, Great work! 

I've added a couple of comments that can be tracked as separate tech-debt issues and refined later.

api/policies/v1/clusteradmissionpolicygroup_types.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (r *ClusterAdmissionPolicyGroup) GetUniqueName() string {
+	return "clusterwide-" + r.Name
+}
+
+func (r *ClusterAdmissionPolicyGroup) GetContextAwareResources() []ContextAwareResource {
+	// FIXME: This is a temporary to make compiler happy
+	// Controller needs to be updated to handle this
+	// A policy group members can have multiple context aware resources
```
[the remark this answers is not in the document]
fabriziosestito: We can add commits to this PR after the first round of reviews, so we ship the full feature under the feature flag.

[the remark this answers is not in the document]
jvanz: Actually, this is not a problem. The policies.yaml serializer knows how to properly set the fields as necessary. Therefore, for now here we can just return a empty slice. I think we well be able to remove this "dummy" members when we refactor the interfaces suggested by Fabrizio in other comment in the PR.

internal/controller/admissionpolicygroup_controller_test.go
jvanz: Yes, they are the same basically. I agree to update the test to use the Table specs. But I would like to do it in a different PR. Just to avoid more changes in an already big PR.

api/policies/v1/policy.go
jvanz: https://github.com/kubewarden/kubewarden-controller/issues/842

internal/controller/policyserver_controller_test.go
```diff
@@ -247,32 +247,195 @@ var _ = Describe("PolicyServer controller", func() {
...
+				Expression:            clusterPolicyGroup.GetExpression(),
+				Message:               clusterPolicyGroup.GetMessage(),
+			}
+
 			policies, err := json.Marshal(policiesMap)
 			Expect(err).ToNot(HaveOccurred())
 
+			// As we have a customization in how we serialize the policies,
+			// let's check if the result json is in the expected format.
+			// Otherwise, the policy server will not start.
```
flavio, line 347: Would it be more readable to have the custom marshaller/unmarshaller tested inside of a traditional (as in, no ginko) test? Maybe introduce a `policyserver_controller_configmap_test.go`

internal/controller/clusteradmissionpolicy_controller.go
```diff
@@ -80,110 +79,35 @@ func (r *ClusterAdmissionPolicyReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+			handler.EnqueueRequestsFromMapFunc(func(_ context.Context, webhookConfiguration client.Object) []reconcile.Request {
+				return findClusterAdmissionPolicyForWebhookConfiguration(webhookConfiguration, r.Log)
+			}),
 		).
 		Watches(
 			&admissionregistrationv1.MutatingWebhookConfiguration{},
-			handler.EnqueueRequestsFromMapFunc(r.findClusterAdmissionPolicyForWebhookConfiguration),
+			handler.EnqueueRequestsFromMapFunc(func(_ context.Context, webhookConfiguration client.Object) []reconcile.Request {
+				return findClusterAdmissionPolicyForWebhookConfiguration(webhookConfiguration, r.Log)
+			}),
```
flavio, line 106: I feel sorry because I asked for this change, but I didn't expect the code to become like that. I didn't notice the functions being called need to access some instance attributes of the `r` object.

I think the old code, like the one on the left looked better. Sorry again! :pray:

internal/controller/admissionpolicygroup_controller_test.go
flavio: yes, let's do that in another PR and, before investing too much time into porting everything into that, let's have a quick look at how the end result might look like.
I don't want to end up with something that is hard to read and understand.

flavio (commented):
Everything looks good to me. I left some minor comments

internal/controller/policyserver_controller_test.go
viccuad: I'm ok with it being a Ginkgo test.

api/policies/v1/admissionpolicygroup_types.go
```diff
@@ -0,0 +1,200 @@
...
+	PolicyGroupSpec `json:""`
+}
+
+// AdmissionPolicyGroup is the Schema for the AdmissionPolicyGroups API
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Policy Server",type=string,JSONPath=`.spec.policyServer`,description="Bound to Policy Server"
+// +kubebuilder:printcolumn:name="Mutating",type=boolean,JSONPath=`.spec.mutating`,description="Whether the policy is mutating"
```
viccuad, line 36:
```suggestion
// +kubebuilder:printcolumn:name="Mutating",type=boolean,JSONPath=`.spec.mutating`,description="Whether the policyGroup is mutating"
```

api/policies/v1/admissionpolicygroup_types.go
```diff
@@ -0,0 +1,200 @@
...
+func (r *AdmissionPolicyGroup) GetPolicyServer() string {
+	return r.Spec.PolicyServer
+}
+
+func (r *AdmissionPolicyGroup) GetUniqueName() string {
+	return "namespacedpolicygroup-" + r.Namespace + "-" + r.Name
+}
+
+func (r *AdmissionPolicyGroup) GetContextAwareResources() []ContextAwareResource {
+	return []ContextAwareResource{}
```
viccuad, line 175: I would add the same comment as in ClusterAdmissionPolicyGroup's has here.

api/policies/v1/admissionpolicygroup_types.go
```diff
@@ -0,0 +1,200 @@
...
+	return r.Spec.MatchPolicy
+}
+
+func (r *AdmissionPolicyGroup) GetMatchConditions() []admissionregistrationv1.MatchCondition {
+	return r.Spec.MatchConditions
+}
+
+// GetNamespaceSelector returns the namespace of the AdmissionPolicyGroup since it is the only namespace we want the policy to be applied to.
+func (r *AdmissionPolicyGroup) GetUpdatedNamespaceSelector(string) *metav1.LabelSelector {
+	return &metav1.LabelSelector{
```
viccuad, line 149: Shouldn't be exclude the `kubewarden` namespace as we do in ClusterAdmissionPolicyGroups?
I realize that here in AdmissionPolicyGroup we have the same behaviour as in AdmissionPolicies.

Still, I'm not sure this isn't a bug both in AdmissionPolicies and AdmissionPolicyGroup.

api/policies/v1/admissionpolicygroup_webhook.go
```diff
@@ -0,0 +1,102 @@
...
+	err := ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+	if err != nil {
+		return fmt.Errorf("failed enrolling webhook with manager: %w", err)
+	}
+	return nil
+}
+
+//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-admissionpolicygroup,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=admissionpoliciesgroup,verbs=create;update,versions=v1,name=madmissionpolicygroup.kb.io,admissionReviewVersions={v1,v1beta1}
```
viccuad, line 47: it's a bit weird that we set up the mutating webhook even if the Groups don't mutate yet, but I guess it's fine.

api/policies/v1/clusteradmissionpolicygroup_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+	err := ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+	if err != nil {
+		return fmt.Errorf("failed enrolling webhook with manager: %w", err)
+	}
+	return nil
+}
+
+//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-clusteradmissionpolicy,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpoliciesgroup,verbs=create;update,versions=v1,name=mclusteradmissionpolicygroup.kb.io,admissionReviewVersions={v1,v1beta1}
```
viccuad, line 49:
```suggestion
//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-clusteradmissionpolicygroup,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpoliciesgroup,verbs=create;update,versions=v1,name=mclusteradmissionpolicygroup.kb.io,admissionReviewVersions={v1,v1beta1}
```

api/policies/v1/admissionpolicygroup_webhook_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.NoError(t, err)
+	require.Empty(t, warnings)
+}
+
+func TestAdmissionPolicyGroupValidateUpdateWithInvalidOldPolicy(t *testing.T) {
+	oldPolicy := clusterAdmissionPolicyGroupFactory()
+	newPolicy := admissionPolicyFactory()
+	warnings, err := newPolicy.ValidateUpdate(oldPolicy)
+	require.Empty(t, warnings)
+	require.ErrorContains(t, err, "object is not of type AdmissionPolicy")
```
viccuad, line 62:
```suggestion
	require.ErrorContains(t, err, "object is not of type AdmissionPolicyGroup")
```

api/policies/v1/clusteradmissionpolicygroup_webhook.go
viccuad, line 49:
```suggestion
//+kubebuilder:webhook:path=/mutate-policies-kubewarden-io-v1-clusteradmissionpolicygroup,mutating=true,failurePolicy=fail,sideEffects=None,groups=policies.kubewarden.io,resources=clusteradmissionpoliciesgroup,verbs=create;update,versions=v1,name=mclusteradmissionpolicygroup.kb.io,admissionReviewVersions={v1,v1beta1}
```

api/policies/v1/clusteradmissionpolicygroup_webhook_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.NoError(t, err)
+	require.Empty(t, warnings)
+}
+
+func TestClusterClusterAdmissionPolicyValidateUpdateWithInvalidOldPolicy(t *testing.T) {
+	oldPolicy := admissionPolicyGroupFactory()
+	newPolicy := clusterAdmissionPolicyGroupFactory()
+	warnings, err := newPolicy.ValidateUpdate(oldPolicy)
+	require.Empty(t, warnings)
+	require.ErrorContains(t, err, "object is not of type ClusterAdmissionPolicy")
```
viccuad, line 62:
```suggestion
	require.ErrorContains(t, err, "object is not of type ClusterAdmissionPolicyGroup")
```

api/policies/v1/policy_types.go
```diff
@@ -150,3 +150,163 @@ type PolicySpec struct {
...
+	// expression that evaluates to a boolean value. If the expression
+	// evaluates to true, the group policy will be considered as accepted,
+	// otherwise, it will be considered as rejected. This expression allows
+	// grouping policies calls and perform logical operations on the
+	// results of the policies. See Kubewarden documentation to learn about
+	// all the features available.
+	// +kubebuilder:validation:Required
+	Expression string `json:"expression"`
+
+	// Field  used to specify the message that will be returned when the
```
viccuad, line 301:
```suggestion
	// Message is used to specify the message that will be returned when the
```

(following godoc)

api/policies/v1/policy_types.go
```diff
@@ -150,3 +150,163 @@ type PolicySpec struct {
...
+	// TimeoutSeconds specifies the timeout for this webhook. After the timeout passes,
+	// the webhook call will be ignored or the API call will fail based on the
+	// failure policy.
+	// The timeout value must be between 1 and 30 seconds.
+	// Default to 10 seconds.
+	// +optional
+	// +kubebuilder:default:=10
+	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
+
+	// The evaluation expression to accept or reject the admission request
```
viccuad, line 287:
```suggestion
	// Expression is the evaluation expression to accept or reject the admission request
```


(following godoc)

internal/controller/policy_subreconciler.go
```diff
@@ -279,6 +280,180 @@ func isPodReady(pod corev1.Pod) bool {
...
+		return []reconcile.Request{}
+	}
+	policyMap, err := getPolicyMapFromConfigMap(configMap)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+	return policyMap.toAdmissionPolicyReconcileRequests()
+}
+
+func findClusterAdmissionPoliciesForConfigMap(object client.Object) []reconcile.Request {
```
viccuad, line 298: Quibble: maybe the `find*()` functions here in policy_subreconciler.go could be better named as follows (with matching godocs):
- `findClusterPoliciesForConfigMap()` (applies to both ClusterAdmissionPolicies and ClusterAdmissionPolicyGroups)
- `findPoliciesForConfigMap()` (applies to both AdmissionPolicies and SdmissionPolicyGroups).

Same for the rest of `find*()` functions here. This would make it easier to understand the reuse of code in the separate *controller.go, if I understood it correctly.

internal/controller/policyserver_controller.go
```diff
@@ -176,53 +176,121 @@ func (r *PolicyServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	return []ctrl.Request{
+		{
+			NamespacedName: client.ObjectKey{
+				Name: policy.Spec.PolicyServer,
+			},
+		},
+	}
+}
+
 // getPolicies returns all admission policies and cluster admission
```
viccuad, line 294: quibble: should mention AdmissionPolicies, ClusterAdmissionPolicies, AdmissionPolicyGroups, ClusterAdmissionPolicGroups

internal/controller/policyserver_controller_configmap.go
```diff
@@ -22,13 +22,68 @@ import (
...
+type policyGroupMember struct {
+	URL                   string                            `json:"url"`
+	Settings              runtime.RawExtension              `json:"settings,omitempty"`
+	ContextAwareResources []policiesv1.ContextAwareResource `json:"contextAwareResources,omitempty"`
+}
+
 type policyServerConfigEntry struct {
 	NamespacedName        types.NamespacedName              `json:"namespacedName"`
-	URL                   string                            `json:"url"`
+	URL                   string                            `json:"url,omitempty"`
```
viccuad, line 33: if we aren't using a different structure to marshall both Groups and Policies, and we want to have omitempty in URL and AllowedToMutate, I would prefer to check on Unmarshalling and:
- error if the json is a normal Policy (`len(Policies) == 0`) and doesn't provide `URL` or `AllowedToMutate`.
- error if the json is a Group (`len(Policies) != 0) and provides a `URL` or AllowedToMutate`.

viccuad (changes requested):
Looks good and is a great job!
I think we should fix the RBAC kubebuilder markers, and I would like to check the marshalling of the policy-server ConfigMap to ensure that the omitempty are being exercised only when they should.

api/policies/v1/admissionpolicygroup_types.go
jvanz: Good question, I do not remember if this is done in purpose. But I would keep as it is now and maybe open an issue to fix it later.

api/policies/v1/admissionpolicygroup_webhook.go
viccuad: true! Thanks for the explanation. Resolving.

viccuad (approved):
Thanks, looks great! :shipit:

Issue #862, opened 2024-08-30
use same certificate generation settings as helm template functions


viccuad:
As we are using `genCA` and `genSigneCert` helm template functions to bootstrap the initial certificates, we need to use the same settings as helm to re-generate the certificates during the rotation process.

Helm uses [sprig](https://github.com/Masterminds/sprig) to expose the certificate generation template functions.
We will need to use the same settings as sprig, see:  https://github.com/Masterminds/sprig/blob/master/crypto.go

viccuad:
@fabriziosestito could you expand a bit on this issue?

PR #866, opened 2024-09-02
fix: generate certificates with the same settings as helm helper functions

fabriziosestito:
## Description
Fixes: #862

### DIFF
--- a/internal/certs/certs.go
+++ b/internal/certs/certs.go
@@ -2,8 +2,9 @@ package certs
 
 import (
 	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/rsa"
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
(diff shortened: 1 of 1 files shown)


internal/certs/certs.go
```diff
@@ -14,15 +14,17 @@ import (
...
+	base         = 1
+	exp          = 128
+	caCommonName = "kubewarden-controller-ca"
 )
 
 // GenerateCA generates a self-signed CA root certificate and private key in PEM format.
 // It accepts validity bounds as parameters.
 func GenerateCA(notBefore, notAfter time.Time) ([]byte, []byte, error) {
-	serialNumber, err := rand.Int(rand.Reader, (&big.Int{}).Exp(big.NewInt(base), big.NewInt(exp), nil))
+	serialNumberUpperBound := new(big.Int).Lsh(big.NewInt(base), exp)
```
flavio, line 26: nitpicking: this is not really a `base`. the `Lsh` function does a left shift, which in this case translates to: `1 * 2^128`. Where `1` is the number multiplied by "2 power `exp`".

I would:

* remove the `base` constant and just use `1` inside of the code
* remove the `exp` constant, or rename it to `serial_upper_bound`

internal/certs/certs.go
```diff
@@ -14,15 +14,17 @@ import (
 )
 
 const (
-	bitSize = 4096
-	base    = 2
-	exp     = 159
+	bitSize      = 2048
```
flavio, line 17: 2048 bits are not enough for a RSA key nowadays. In fact, looking at the `sprig` library they use a 4096 key (see [here](https://github.com/Masterminds/sprig/blob/e708470d529a10ac1a3f02ab6fdd339b65958372/crypto.go#L158-L159))

flavio (changes requested):

internal/certs/certs.go
flavio: I was wrong, both with the link and with the statement about RSA 2048 not being enough.

Looking at the latest (July 2024) [NIST publication on the topic](https://csrc.nist.gov/pubs/sp/800/78/5/final) this is fine until 2030:

image

flavio:
After a quick discussion with @fabriziosestito, we decided to change the code to use ECDSA p256 to generate all our certificates.

That means we will also use the `genCAWithKey` and the `genPrivateKey` helpers of helm (see http://masterminds.github.io/sprig/crypto.html)

fabriziosestito:
> After a quick discussion with @fabriziosestito, we decided to change the code to use ECDSA p256 to generate all our certificates.
> 
> That means we will also use the `genCAWithKey` and the `genPrivateKey` helpers of helm (see http://masterminds.github.io/sprig/crypto.html)

moving back to "in progress"

flavio (approved):
Thanks for all the changes

Issue #854, opened 2024-08-26
tilt: webhooks should have the certificates set


jvanz:
In a recent [change](https://github.com/kubewarden/helm-charts/pull/488) the root CA and the certificate used by the controller are bootstrapped by Helm. However, the Tiltfile used in the controller repository does not use the certificates generated by the Helm chart. Because it uses the configuration generated by the kustomize in the repository. Therefore, it is always running the latest webhooks configuration defined in the controller code. However, this cause issues because the controller does not have the certificates bootstrapped. Therefore, we need to update the Tiltfile to use the root CA generated by Helm together with the configurations generated by Kustomize locally.

### Acceptance criteria

- Check if the Helm execution in the Tiltfile genetates the required certificates
- If the certificates are generated in the Tiltfile call, reuse them in the webhooks configuration together with the values generated by Kustomize locally.
- If the certificates are not generated in the Helm call, generate the certificates in the Tiltfile and updates the webhooks configuration from Kustomize.

flavio:
Thanks for pointing this out. We should get this done ASAP

flavio:
CC @fabriziosestito

PR #867, opened 2024-09-02
fix(Tiltfile): remove overriding helm webhooks with kustomize ones

fabriziosestito:
## Description

Recently we changed the Tiltfile to override the Helm chart's webhooks with the ones provided by the Kustomize folder.
However, this breaks since we now inject initial certificate bundles from the helm chart.
While having the scaffolded webhooks updated automatically can be handy, we don't expect adding webhooks frequently to justify the added complexity of merging the two configurations.

This PR fixes: #854  by removing the Kustomize webhooks logic in the tiltfile

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -39,24 +39,6 @@ for role in roles:
 if len(roles_rules_mapping["ClusterRole"]) == 0 or len(roles_rules_mapping["Role"]) == 0:
     fail("Failed to load cluster and namespace roles")
 
-# Get the webhook configuration with the latest values generated in the 
-# controller repository. 
-mutating_webhooks = []
-validating_webhooks = []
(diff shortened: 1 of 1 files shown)


flavio:
Maybe I'm wrong, but that would cause the validating/mutating webhooks defined by controller to **not** be used in development mode because they would not be registered. Am I right?

fabriziosestito:
> Maybe I'm wrong, but that would cause the validating/mutating webhooks defined by controller to **not** be used in development mode because they would not be registered. Am I right?

@flavio they will be configured by the helm chart installed by the tiltfile

flavio (approved):
LGTM

jvanz (approved):

Issue #824, opened 2024-07-23
Refactor policy validation on create and update, including tests


viccuad:
We need to refactor and cleanup webhooks validation, as we realized that we could increase readability and test coverage

### Acceptance criteria

- The file `api/policies/v1/policy_validation_matchconditions.go` was copy-pasted from k8s.io/kubernetes (see license, copyright, comment at its beginning). We tried to import as-is, but it makes more sense to own it instead. For that:
  - Enable linter in the file (check `golangci-lint.yml`).
  - Consider moving its functions into `api/policies/v1/policy_validation.go` and perform any cleanup necessary.
- Factories should be refactored with the builder pattern.
- Add integration tests for the webhooks methods: `ValidateCreate`, `ValidateDelete`, and `ValidateUpdate`.  
ATM We test each validator in isolation (see `policy_validation_test.go`), however, we do not test that the validators are correctly integrated into the webhooks. We could for instance add just a couple of test cases: one with a valid resource and one with a resource that all fields in an invalid state, so that we are sure we are passing through the right validators.

PR #870, opened 2024-09-05
feat: validate PolicyGroup expressions

### LINKED ISSUE #859: Extend kubewarden-controller admission webhook: it must validate the CEL expression defined inside of the new CRDs

fabriziosestito:
## Description
This PR adds a validation step for the `expression` field of the policy group spec.
An expression must be a valid CEL expression that evaluates to `bool`. Only logic and equality operators are allowed, and policy members are available as functions with zero arity.

It also refactors the policy validation utility functions, cleaning up the existing tests.
Additional cleanup will be needed, which is tracked by: https://github.com/kubewarden/kubewarden-controller/issues/824

Fixes #859

### DIFF
--- a/api/policies/v1/policy_validation.go
+++ b/api/policies/v1/policy_validation.go
@@ -24,70 +24,84 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+func validatePolicyCreate(policy Policy) field.ErrorList {
+	var allErrors field.ErrorList
+
+	allErrors = append(allErrors, validateRulesField(policy)...)
+	allErrors = append(allErrors, validateMatchConditionsField(policy)...)
+
+	return allErrors
+}
+
+func validatePolicyUpdate(oldPolicy, newPolicy Policy) field.ErrorList {
+	var allErrors field.ErrorList
+
+	allErrors = append(allErrors, validateRulesField(newPolicy)...)
+	allErrors = append(allErrors, validateMatchConditionsField(newPolicy)...)
+	if err := validatePolicyServerField(oldPolicy, newPolicy); err != nil {
+		allErrors = append(allErrors, err)
+	}
+	if err := validatePolicyModeField(oldPolicy, newPolicy); err != nil {
+		allErrors = append(allErrors, err)
+	}
+
+	return allErrors
+}
+
 // Validates the spec.Rules field for non-empty, webhook-valid rules.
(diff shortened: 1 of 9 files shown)


jvanz (approved):
LGTM

viccuad (approved):
LGTM! :rocket:

api/policies/v1/policy_validation.go
```diff
@@ -108,16 +122,22 @@ func validateMatchConditionsField(policy Policy) field.ErrorList {
...
 	}
+
 	return nil
 }
 
-func validatePolicyGroupMembers(policy Policy) *field.Error {
-	if policy.IsPolicyGroup() && len(policy.GetPolicyMembers()) == 0 {
-		return field.Invalid(field.NewPath("spec").Child("policies"), "", "policy groups must have at least one policy member")
+func validatePolicyModeField(oldPolicy, newPolicy Policy) *field.Error {
+	if oldPolicy.GetPolicyMode() != newPolicy.GetPolicyMode() {
```
flavio, line 137: This check is wrong, because it would actually prevent the transition from `monitor` to `protect`.

The old check was this one:

https://github.com/kubewarden/kubewarden-controller/blob/ecfb764b90588a0163e0efa8d86fef22e780c474/api/policies/v1/policy_validation.go#L152-L158

Moreover, if I'm right about this check being wrong, that means our unit tests are not covering the case of a policy changing its mode

api/policies/v1/policy_validation_test.go
```diff
@@ -263,38 +325,47 @@ func TestValidatePolicyUpdate(t *testing.T) {
...
-					Rule: admissionregistrationv1.Rule{
-						APIGroups:   []string{""},
-						APIVersions: []string{"v1"},
-						Resources:   []string{"pods"},
-					},
+			}}, nil, "default", "protect"),
+			"",
+		},
+		{
+			"policy mode changed from protect to monitor",
```
flavio, line 339: "random" comment location: we should also add a test case ensuring the transition from monitor to protect is allowed

flavio (changes requested):

api/policies/v1/policy_validation_test.go
fabriziosestito: done in https://github.com/kubewarden/kubewarden-controller/pull/870/commits/8895033badc57c4c5868a30fff9df7d98c024ce0

flavio (approved):
LGTM

Issue #872, opened 2024-09-06
PolicyGroups CRDs do not match PolicyServer config


fabriziosestito:
### Is your feature request related to a problem?

PolicyGroups CRDs do not match the PolicyServer config.
Policy group members are expressed as an array in the CRD/RFC, while PolicyServer uses a map.

E.g.

PolicyServer config:
```yaml
pod-image-signatures: # policy group
  policies:
    sigstore_pgp:
      url: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
      settings:
        signatures:
          - image: "*"
            pubKeys:
              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
    sigstore_gh_action:
      url: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
      settings:
        signatures:
          - image: "*"
            githubActions:
            owner: "kubewarden"
    reject_latest_tag:
      url: ghcr.io/kubewarden/policies/trusted-repos-policy:v0.1.12
      settings:
        tags:
          reject:
            - latest
  expression: "sigstore_pgp() || (sigstore_gh_action() && reject_latest_tag())"
  message: "The group policy is rejected."
  ```
  
CRD:
```yaml
  policies:
   - name: sigstore_pgp
     url: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
     settings:
      signatures:
        - image: "*"
          pubKeys:
            - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
            - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
    - name: sigstore_gh_action:
      url: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
      settings:
        signatures:
          - image: "*"
            githubActions:
            owner: "kubewarden"
....
  ```

### Solution you'd like

- [x] update the CRD to use a map instead of an array in the policy group members field: https://github.com/kubewarden/kubewarden-controller/pull/878
- [x] update the helm chart
- [x] update the RFC accordingly

flavio:
Everything has been done, closing

PR #878, opened 2024-09-10
fix: policy group members as map

fabriziosestito:
## Description

Fixes: #872 

Also, it renames `GetPolicyMembers` to `GetPolicyGroupMembers` and the `url` field to `module` to be consistent with the single Policy CRD.


viccuad (approved):
LGTM. Thanks for the renames!

flavio (approved):
LGTM, thanks for the cleanup

Issue #876, opened 2024-09-09
Validate PolicyGroup members name


fabriziosestito:
### Is your feature request related to a problem?

Policy group members will be bound to functions with the same name when evaluating a CEL expression. Therefore, policy member's names should follow the same validation rules as CEL names.

### Solution you'd like

Add a validation step to validate policy member names.

fabriziosestito:
@jvanz this could be helpful: https://github.com/kubewarden/cel-policy/pull/83#discussion_r1751393241

jvanz:
> [@jvanz](https://github.com/jvanz) this could be helpful: [kubewarden/cel-policy#83 (comment)](https://github.com/kubewarden/cel-policy/pull/83#discussion_r1751393241)

Thanks! I was reading the code. It's a good starting point. However, I think we cannot use that regex because it's meant to [escape the CEL expressions](https://github.com/kubernetes/apiserver/blob/14b3ab9aa060d9d0a8151a7f2bbd03e13465a1e7/pkg/cel/escaping.go#L100-L136), not to check if it's a valid identifier. For example, that regex is used to identify the dot in expressions like [a.b](https://github.com/kubernetes/apiserver/blob/14b3ab9aa060d9d0a8151a7f2bbd03e13465a1e7/pkg/cel/escaping_test.go#L36). Therefore, if we use it, we will allow policy names with dots, which is not supported. Consequently, I've written another regex based on the [policy definition](https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax). Additionally, I've decided to use a set to store the keywords reserved by the CEL policy because it matches the policy syntax.

PR #880, opened 2024-09-11
feat: validate policy group members names.

jvanz:
## Description

Updates the validation webhook for the policy groups types to validate the policies members names. This is necessary because the names are used as CEL functions names in the policy group expression. Therefore, they need to be compliant with the CEL syntax. The regex used to validate the name is based on the language definition:
https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax

Fix https://github.com/kubewarden/kubewarden-controller/issues/876

### DIFF
--- a/api/policies/v1/policygroup_validation.go
+++ b/api/policies/v1/policygroup_validation.go
@@ -2,22 +2,38 @@ package v1
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common/decls"
 	"github.com/google/cel-go/common/operators"
 	"github.com/google/cel-go/common/stdlib"
 	"github.com/google/cel-go/common/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// Regex to validate the policy members names.
+// For more information about the CEL grammar, see
+// https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax
+var idenRegex = regexp.MustCompile(`^[_a-zA-Z][_a-zA-Z0-9]*$`)
+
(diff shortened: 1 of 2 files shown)


api/policies/v1/policygroup_validation.go
```diff
@@ -44,6 +62,12 @@ func validatePolicyGroupMembers(policy Policy) *field.Error {
 	if len(policy.GetPolicyGroupMembers()) == 0 {
 		return field.Required(field.NewPath("spec").Child("policies"), "policy groups must have at least one policy member")
 	}
+	for memberName := range policy.GetPolicyGroupMembers() {
+		_, matchReservedSymbol := celReservedSymbols[memberName]
+		if len(memberName) == 0 || matchReservedSymbol || !idenRegex.MatchString(memberName) {
+			return field.Invalid(field.NewPath("spec").Child("policies"), memberName, "policy group member name is invalid")
```
fabriziosestito, line 68: this function should now return `field.ErrorList` and we should append here instead of returning

api/policies/v1/policygroup_validation_test.go
```diff
@@ -163,6 +163,120 @@ func TestValidatePolicyGroupMembers(t *testing.T) {
...
+								Module: "ghcr.io/kubewarden/tests/user-group-psp:v0.4.9",
+							},
+						},
+					},
+				},
+			},
+			`spec.policies: Invalid value: "0policy1": policy group member name is invalid`,
+		},
+		{
+			"policy member name cannot have special chars in the middle",
```
fabriziosestito, line 221: names can't have any special character at all, remove `in the middle`

fabriziosestito (changes requested):

viccuad (approved):
LGTM modulo Fabrizio's comments.

fabriziosestito (approved):

PR #881, opened 2024-09-13
Add govulncheck workflow

robertsirc:
I am adding govulncheck GH action in. This is to check for vulnerabilities on a schedule and if go.sum has been modified.


flavio (approved):
LGTM

I guess it would be nice to run it also on PRs where the `go.sum` file is changed

PR #891, opened 2024-09-23
refactor: remove watches used only by integration tests

flavio:
These watches were used a long time ago by the integration tests, they are not needed anymore.

Their removal redeuces the resource usage when running in production.


fabriziosestito (approved):
LGTM

Issue #887, opened 2024-09-20
Policy Group: `ValidatingWebhookConfiguration` is not removed after policy is deleted


flavio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Testing Kubewarden 1.17.0-rc3

I've created a `ClusterAdmissionPolicyGroup` policy, which lead to the creation of a `ValidatingWebhookConfiguration` resource.

Then I removed the `ClusterAdmissionPolicyGroup`. The configuration of the Policy Server was updated accordingly, but the associated `ValidatingWebhookConfiguration` was not removed.

This is a big deal, the webhook will still be registered, but unreachable. Hence all the request relevant for this webhook will fail. This can cause a DOS inside of the cluster.

### Expected Behavior

The ``ValidatingWebhookConfiguration` resource is removed immediately, even before the configuration of the Policy Server is updated.

### Steps To Reproduce

Deploy kubewarden 1.17.0-rc3.

Create the following policy:

```
apiVersion: policies.kubewarden.io/v1
kind: ClusterAdmissionPolicyGroup # or AdmissionPolicyGroup
metadata:
  name: demo
spec:
  rules:
    - apiGroups: [""]
      apiVersions: ["v1"]
      resources: ["pods"]
      operations:
        - CREATE
        - UPDATE
  policies:
    signed_by_alice:
      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.3.0
      settings:
        modifyImagesWithDigest: false
        signatures:
          - image: "*"
            pubKeys:
              - |
                -----BEGIN PUBLIC KEY-----
                MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEyg65hiNHt8FXTamzCn34IE3qMGcV
                yQz3gPlhoKq3yqa1GIofcgLjUZtcKlUSVAU2/S5gXqyDnsW6466Jx/ZVlg==
                -----END PUBLIC KEY-----
    signed_by_bob:
      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.3.0
      settings:
        modifyImagesWithDigest: false
        signatures:
          - image: "*"
            pubKeys:
              - |
                -----BEGIN PUBLIC KEY-----
                MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEswA3Ec4w1ErOpeLPfCdkrh8jvk3X
[...]

PR #892, opened 2024-09-23
test: add policy deletion -> webhook deletion tests

fabriziosestito:
## Description

Related to #887, adds missing deletion tests.


flavio (approved):
Thanks for having increased the coverage, we definitely missed these tests

PR #893, opened 2024-09-23
fix: add group annotation to WebhookConfiguration and filter watch requests 

fabriziosestito:
## Description

This PR adds a group annotation to ValidatingWebhookConfiguration and MutatingWebhookConfiguration for policy groups, enabling us to filter watch requests sent to the respective reconcilers.


flavio (commented):
Can you extend the unit tests please?

PR #894, opened 2024-09-24
refactor: simplify wehbook configuration controller handlers

fabriziosestito:
## Description

Before this change, we leveraged several annotations to check if a `WebhookConfiguration` was related to the type of policy managed by a controller.
For instance, the `scope` and `group` annotations were used to check if a policy was cluster or namespace scoped and if it was a group. 
When adding `PolicyGroups` we realized that it was hard to maintain multiple annotations.

This PR simplifies the `WebhookConfiguration` controller handlers, by using just two annotations (`name` and `namespace`) ~and by using the client to check if the resource of a certain type exists.
This is ok regarding performance and API server load since the request is cached.~
and returning a reconcile request.

### DIFF
--- a/internal/controller/admissionpolicy_controller.go
+++ b/internal/controller/admissionpolicy_controller.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	policiesv1 "github.com/kubewarden/kubewarden-controller/api/policies/v1"
+	"github.com/kubewarden/kubewarden-controller/internal/constants"
 )
 
 // Warning: this controller is deployed by a helm chart which has its own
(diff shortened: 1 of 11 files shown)


flavio (commented):
Thanks an interesting approach. Are there consequences caused by using a field indexer?

fabriziosestito:
> Thanks an interesting approach. Are there consequences caused by using a field indexer?

I don't think so. However I'd like to proceed with the `name`, `namespace` approach  for simplicty.

internal/controller/admissionpolicy_controller.go
```diff
@@ -101,6 +103,40 @@ func (r *AdmissionPolicyReconciler) findAdmissionPoliciesForPod(ctx context.Cont
...
+	err := r.Get(ctx, client.ObjectKey{
+		Name:      policyName,
+		Namespace: policyNamespace,
+	}, &admissionPolicy)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			r.Log.Error(err, "cannot retrieve AdmissionPolicy")
+		}
+		return []reconcile.Request{}
+	}
```
flavio, line 132: Why do we need to perform the lookup of the policy? Can't we just fire up the reconciliation event and let the policy reconciliation loop deal with the case of a missing policy?

I know this `Get` request is using the cache of k8s's client, but it seems like a duplication of code

fabriziosestito: you are totally right, I removed the get as we already do this check in the reconciler ignoring the not found error.

PR #895, opened 2024-09-26
Fix gh workflow permissions

robertsirc:
Fixing permission in govulncheck. This will fix part of the OpenSFF score card.


viccuad (approved):
Thanks! We should have an eye on followup PRs, as this PR only runs the CI code from main for security reasons.

jvanz (approved):

PR #896, opened 2024-09-26
feat: shortName for the CRDs.

jvanz:
## Description

Adds shortName for the Kubewarden CRDs. This allows users to get Kubewarden types using acronymous and making the commands smaller to type.

@kubewarden/kubewarden-developers I'm creating this PR as draft because I would like to investigate why `controller-gen` is not updating the CRDs with the `shortNames` for some CRDs. However, I would like to know if you are fine with the proposed `shortNames`. I've prefixed all of them with `k` to flag this is a `kubewarden` resource.

### DIFF
--- a/api/policies/v1alpha2/clusteradmissionpolicy_types.go
+++ b/api/policies/v1alpha2/clusteradmissionpolicy_types.go
@@ -77,12 +77,13 @@ type ClusterAdmissionPolicySpec struct {
 // ClusterAdmissionPolicy is the Schema for the clusteradmissionpolicies API
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Cluster,shortName=cap
 // +kubebuilder:printcolumn:name="Policy Server",type=string,JSONPath=`.spec.policyServer`,description="Bound to Policy Server"
 // +kubebuilder:printcolumn:name="Mutating",type=boolean,JSONPath=`.spec.mutating`,description="Whether the policy is mutating"
 // +kubebuilder:printcolumn:name="Mode",type=string,JSONPath=`.spec.mode`,description="Policy deployment mode"
 // +kubebuilder:printcolumn:name="Observed mode",type=string,JSONPath=`.status.mode`,description="Policy deployment mode observed on the assigned Policy Server"
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.policyStatus`,description="Status of the policy"
+// +kubebuilder:deprecatedversion:warning="This version is deprecated. Please, consider using v1"
 type ClusterAdmissionPolicy struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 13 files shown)


viccuad (approved):
This is great, I'm happy with the shorNames selected.

viccuad:
On second thought, I wonder if there's a possibility to have them pronounceable.
Maybe taking out the k makes them more pronounceable: ap, cap, apg (apgroup), capg (capgroup), ps.

config/crd/bases/policies.kubewarden.io_admissionpolicygroups.yaml
```diff
@@ -11,6 +11,8 @@ spec:
     kind: AdmissionPolicyGroup
     listKind: AdmissionPolicyGroupList
     plural: admissionpolicygroups
+    shortNames:
+    - kapg
```
fabriziosestito, line 15: maybe we could use `kw` as a prefix

fabriziosestito (commented):
Love this! looking forward

jvanz:
@kubewarden/kubewarden-developers I've needed to update v1alpha2 CRDs because `controller-gen` change the order of it process the files to extract the kubebuilder markers each time it runs. Therefore, each time it runs the CRDs generation is different. When the v1alpha2 CRDs types are  the last one to be processed, the `shortNames` are not define.

api/policies/v1alpha2/clusteradmissionpolicy_types.go
```diff
@@ -77,12 +77,13 @@ type ClusterAdmissionPolicySpec struct {
 // ClusterAdmissionPolicy is the Schema for the clusteradmissionpolicies API
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Cluster,shortName=cap
```
jvanz, line 80: @kubewarden/kubewarden-developers I've needed to update v1alpha2 CRDs because controller-gen change the order of it process the files to extract the kubebuilder markers each time it runs. Therefore, each time it runs the CRDs generation is different. When the v1alpha2 CRDs types are the last one to be processed, the shortNames are not define.

jvanz: I've opened an [issue](https://github.com/kubernetes-sigs/controller-tools/issues/1068) asking if this is expected or it is a bug.

flavio: Can't we drop these alpha resources? They are not used since a long time...

viccuad: +1 on dropping the alpha resources.

jvanz: I thoug about that as well. I've even use the `//+kubebuilder:skip` to ignore the `v1alpha2`. But than I've wondered, what happens if someone is still using the alpha version? The conversion will happen out of the box for resources already installed? Because of that, I've decided to mark the alpha version as deprecated first. But if we can drop it with no issue, I prefer removing it as well.

jvanz: Considering the previous comment. I suggest to keep the `v1alpha2` for now, mark it as deprecated and create an issue to attack the old CRD removal.

Issue #900, opened 2024-10-01
Kubewarden fails to run on k8s 1.23


kravciak:
Tested with:
 - K8s: k3d and rke1
 - Kubewarden: 1.17.0, 1.16.0, 1.15.0 (all have this bug)

```
> kubectl get pod -n cattle-kubewarden-system -o wide
NAME                                             READY   STATUS             RESTARTS       AGE   IP           NODE         NOMINATED NODE   READINESS GATES
audit-scanner-28795560-4j949                     0/1     Completed          0              23m   10.42.0.99   10.1.10.43   <none>           <none>
rancher-kubewarden-controller-76d7fc6467-np4tw   0/1     CrashLoopBackOff   15 (98s ago)   54m   10.42.0.98   10.1.10.43   <none>           <none>

> kubectl describe pod rancher-kubewarden-controller-76d7fc6467-np4tw -n cattle-kubewarden-system
Name:         rancher-kubewarden-controller-76d7fc6467-np4tw
Namespace:    cattle-kubewarden-system
..........................
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  47m                    default-scheduler  Successfully assigned cattle-kubewarden-system/rancher-kubewarden-controller-76d7fc6467-np4tw to 10.1.10.43
  Normal   Pulled     46m (x4 over 47m)      kubelet            Container image "ghcr.io/kubewarden/kubewarden-controller:v1.17.0" already present on machine
  Normal   Created    46m (x4 over 47m)      kubelet            Created container manager
  Normal   Started    46m (x4 over 47m)      kubelet            Started container manager
  Warning  BackOff    2m40s (x230 over 47m)  kubelet            Back-off restarting failed container

> kubectl logs rancher-kubewarden-controller-76d7fc6467-np4tw -n cattle-kubewarden-system
{"level":"error","ts":"2024-09-30T22:26:37Z","logger":"setup","msg":"unable to check for feature gate AdmissionWebhookMatchConditions","error":"failed to fetch OpenAPI spec: the server could not find the requested resource","stacktrace":"main.main\n\t/workspace/cmd/main.go:140\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:272"}
```

PR #901, opened 2024-10-01
fix: do not exit when k8s' matchConditions are not available

flavio:
Do not exit with an error when the `matchConditions` field is not available for ValidatingWebhookConfiguration and MutatingWebhookConfiguration.

The code is already structured to work when this feature is not available, hence we don't have to cause the controller to exit with an
error.

Fixes https://github.com/kubewarden/kubewarden-controller/issues/900


flavio:
CC @williamlin-suse

viccuad:
For completion, the check performed in `CheckAdmissionWebhookMatchConditions()`  may fail indeed, as we are using OpenAPIV3 to query the cluster. This Kubernetes feature is alpha since 1.23 (and may not be enabled), yet only GA since 1.27. But we shouldn't halt indeed if there's an error.

PR #952, opened 2025-01-03
fix: rename "url" field into "module".

jvanz:
## Description

The "url" field used in the policy server configuration configmap has been renamed to "module". This ensures that the name used in the configmap follow the same name used in the CRDs.

Fix https://github.com/kubewarden/policy-server/issues/906

## Test

```shell
make POLICY_SERVER_VERSION=issue906 POLICY_SERVER_REPOSITORY=ghcr.io/jvanz/policy-server test e2e-tests
```

## Additional Information

Needs to be tested together with https://github.com/kubewarden/policy-server/pull/1016

### DIFF
--- a/internal/controller/utils_test.go
+++ b/internal/controller/utils_test.go
@@ -40,12 +40,15 @@ import (
 	"github.com/kubewarden/kubewarden-controller/internal/constants"
 )
 
-const integrationTestsFinalizer = "integration-tests-safety-net-finalizer"
+const (
+	integrationTestsFinalizer   = "integration-tests-safety-net-finalizer"
+	defaultKubewardenRepository = "ghcr.io/kubewarden/policy-server"
+)
 
 var (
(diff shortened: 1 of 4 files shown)


viccuad (changes requested):
Looks good, but I think we should also include the regenerated CRD docs.

viccuad (approved):
LGTM!

internal/controller/utils_test.go
```diff
@@ -81,6 +81,14 @@ var (
 	}
 )
 
+func policyServerRepository() string {
+	repository, ok := os.LookupEnv("POLICY_SERVER_REPOSITORY")
+	if !ok {
+		return "ghcr.io/kubewarden/policy-server"
```
fabriziosestito, line 87: maybe we could use a constant

fabriziosestito (approved):

PR #955, opened 2025-01-07
feat(test): use builder to build test objects.

jvanz:
## Description

Creates builders to allow dynamic object building during tests.

Part of #824 

@kubewarden/kubewarden-developers this is one of a series of PR open to refactor the webhooks tests. This is a initial change to use builder pattern to build (hehe) test objects. This is not a 100% convertion. But it's a start to see if you are all  fine with the current changes before the next one following the same idea.

### DIFF
--- a/internal/controller/factories_test.go
+++ b/internal/controller/factories_test.go
@@ -1,0 +1,318 @@
+package controller
+
+import (
+	"os"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	policiesv1 "github.com/kubewarden/kubewarden-controller/api/policies/v1"
+	"github.com/kubewarden/kubewarden-controller/internal/constants"
+)
+
+type admissionPolicyFactory struct {
+	name         string
+	namespace    string
+	policyServer string
+	mutating     bool
+	rules        []admissionregistrationv1.RuleWithOperations
+	module       string
+}
+
+func newAdmissionPolicyFactory() *admissionPolicyFactory {
+	return &admissionPolicyFactory{
+		name:         newName("validating-policy"),
+		namespace:    "",
+		policyServer: "",
+		mutating:     false,
+		rules:        []admissionregistrationv1.RuleWithOperations{},
+		module:       "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
+	}
+}
+
+func (fac *admissionPolicyFactory) withName(name string) *admissionPolicyFactory {
+	fac.name = name
+	return fac
+}
+
+func (fac *admissionPolicyFactory) withNamespace(namespace string) *admissionPolicyFactory {
+	fac.namespace = namespace
+	return fac
+}
+
+func (fac *admissionPolicyFactory) withPolicyServer(policyServer string) *admissionPolicyFactory {
+	fac.policyServer = policyServer
+	return fac
+}
+
+func (fac *admissionPolicyFactory) withMutating(mutating bool) *admissionPolicyFactory {
+	fac.mutating = mutating
+	return fac
+}
+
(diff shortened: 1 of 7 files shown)


fabriziosestito:
@jvanz this looks great, few comments below:

- We could name the builders "factory"; this naming is commonly used in tests to describe generated test data (as opposed to fixtures). See https://github.com/beam-community/ex_machina
- [Reference link](https://github.com/kubewarden/kubewarden-controller/blob/c566103a603d6976bcdb459a307a4a600db547a1/internal/controller/utils_test.go#L96): We already use a factory for other objects, but we are not applying the builder pattern. It would be good to migrate to it.  
- It would be nice to place these in a separate file (e.g., `factories_test.go`).

jvanz:
@kubewarden/kubewarden-developers I've rewrite all the factories function from the `internal/controller/utils_test.go`. Let me know what do you think.

right after pushing the code I had another idea. Which is using a factory method with variable argument to define the objects configuration. For example:

```go
admissionPolicyFactory(withName("foo"), withPolicyServer("default"))
// or
admissionPolicyFactory(withName("foo"), withPolicyServer("default"), withMutating(true))
```

Something similar to what other go projects do to customize configuration. But I'm not sure what do you prefer. I do have have any string preference on either options. I just want to share all the paths with you.

internal/controller/factories_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+func newAdmissionPolicyFactory() *admissionPolicyFactory {
+	return &admissionPolicyFactory{
+		newName("validating-policy"),
+		"",
+		"",
+		false,
+		[]admissionregistrationv1.RuleWithOperations{},
+		"registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
+	}
```
flavio, line 30: I prefer to be explicit when initializing a struct:

```golang
return &admissionPolicyFactory{
  name: newName("validating-policy"),
  namespace: "",
  // and so on
}
```

internal/controller/factories_test.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func newClusterAdmissionPolicyFactory() *clusterAdmissionPolicyFactory {
+	return &clusterAdmissionPolicyFactory{
+		newName("validating-cluster-policy"),
+		"",
+		false,
+		[]admissionregistrationv1.RuleWithOperations{},
+		"registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
+	}
```
flavio, line 101: same here

internal/controller/factories_test.go
```diff
@@ -0,0 +1,318 @@
...
+	namespace    string
+	policyServer string
+}
+
+func newAdmissionPolicyGroupFactory() *admissionPolicyGroupFactory {
+	return &admissionPolicyGroupFactory{
+		newName("validating-policygroup"),
+		"",
+		"",
+	}
```
flavio, line 195: same here

internal/controller/factories_test.go
```diff
@@ -0,0 +1,318 @@
...
+	name         string
+	policyServer string
+}
+
+func newClusterAdmissionPolicyGroupFactory() *clusterAdmissionPolicyGroupFactory {
+	return &clusterAdmissionPolicyGroupFactory{
+		newName("validating-policygroup"),
+		"",
+	}
+}
```
flavio, line 256: same here

flavio (changes requested):

fabriziosestito:
> ```go
> admissionPolicyFactory(withName("foo"), withPolicyServer("default"))
> // or
> admissionPolicyFactory(withName("foo"), withPolicyServer("default"), withMutating(true))
> ```

I think that what we have now is clearer than functional options

fabriziosestito (approved):

flavio (approved):
LGTM, thanks for the changes

PR #957, opened 2025-01-10
Webhook integration tests and clean up MatchCondition validation code 

### LINKED ISSUE #824: Refactor policy validation on create and update, including tests

jvanz:
## Description

Adds tests to ensure that the function used by the webhooks server is calling the correct validation functions and moves the code from policy_validation_matchconditions.go file into policy_validation.go file, removing the unnecessary code.

Fix #824 

## Test

```shell
make test 
```


jvanz:
@fabriziosestito this is not exactly what I want to show you because I still have the new tests under the controller directory. But I could not finish the changes yet. Therefore, I'll leave it as it is now and continue next day.

api/policies/v1/zz_generated.deepcopy.go
```diff
@@ -55,6 +55,33 @@ func (in *AdmissionPolicy) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *AdmissionPolicyFactory) DeepCopyInto(out *AdmissionPolicyFactory) {
```
fabriziosestito, line 59: we need to exclude the factory types from the codegen

fabriziosestito (changes requested):
Overall LGTM. 
I like the approach of testing the method directly without starting the webhook, it simplifies the tests. This is ok since we have e2e tests and are sure that webhooks are correctly wired. 

If we want to exclude factories from the final binary, we could use the build flag approach, similar to what we used in [audit-scanner's test-utils](https://github.com/kubewarden/audit-scanner/blob/eb1a168845c3a7a409d66eda69be7805e536c5e1/internal/testutils/testutils.go#L1).

 @kubewarden/kubewarden-developers wdyt?
Keep in mind that the LSP may require some configuration on our machines, as it runs without the testing build flag and will not recognize the factory code.

fabriziosestito (approved):

PR #984, opened 2025-01-27
fix: AdmissionPolicyGroup must not have context aware resources

flavio:
Prior to this commit, it was possible to declare AdmissionPolicyGroup resources with policies that had context aware resources.

Context awareness allows a policy to query the Kubernetes API with `list` and `get` verbs. The queries are run using the ServiceAccount of the Policy Server instance that hosts the policy.

AdmissionPolicyGroup are namespaced CRDs which can be created by non-admin Kubernetes users. The ability to deploy AdmissionPolicyGroup with context aware resources exposes the cluster to potential leaks of data.
For example, a non-admin user, might create an AdmissionPolicyGroup with one of its policies that has access to Kubernetes Secrets. The policy would then be able to `get` Secrets that are not available to the non-admin user (if the Policy Server is running with a ServiceAccount that has been explicitly configured to have `get` access to all the Secrets of the cluster).

This commit fixes the by removing the `ContextAwareResources` field from the `AdmissionPolicyGroup`.

## What happens after this commit is deployed

The existing AdmissionPolicyGroup will be automatically updated and they will lose their `.spec.policies.[*].ContextAwareResources` field (if set).

The kubewarden controller will also automatically reconcile the Policy Server configuration.


pjbgf (approved):

viccuad (approved):

flavio:
I've rebased against the `main` branch

Issue #993, opened 2025-02-04
Add mTLS support


fabriziosestito:
The controller should take a new flag/environment variable that contains the name of the CA used by the client.

The certificate must be injected inside of the Pod. This can be done by extending our helm chart to mount the CA certificate from a ConfigMap into the Pod.

Our controller is built using controller-runtime, which means we can use [this](https://github.com/kubernetes-sigs/controller-runtime/pull/802/files) feature.

The kubewarden-controller is also in charge of creating the Policy Server Deployment objects. When mTLS is enabled for the kubewarden-controller, this setting should be automatically propagated also to the Policy Server deployments.

AC:

 - [x] introduce a new cli flag that contains the full path to the CA certificate used to verify the incoming connections to the webhooks
  - [x] change the validating and mutating webhook server hosted by kubewarden-controller to make use of this certificate
  - [x] generate Policy Server deployments that have mTLS enabled:
    - [x] The ConfigMap mounted by the kubewarden-controller is also mounted by the Policy Server
    - [x] The Policy Server is started with mTLS enabled
  - [ ] Add the proper labels required by the NetworkPolicy that blocks connection from the host to the Policy Server service (see [these policies](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/rancher-webhook-hardening))

fabriziosestito:
> Add the proper labels required by the NetworkPolicy that blocks connection from the host to the Policy Server service (see [these policies](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/rancher-webhook-hardening))
@kubewarden/kubewarden-developers we already use `app.kubernetes.io/part-of: kubewarden`, 
is this enough?

PR #1000, opened 2025-02-06
feat: add mTLS support

fabriziosestito:
## Description

This PR introduces the `--client-ca-config-map` flag.  
When specified, the setup function configures the webhook server with the client CA.  
Additionally, it mounts the ConfigMap in the reconciled PolicyServer deployments and sets the `KUBEWARDEN_CLIENT_CA_FILE` environment variable to the CA path.

Fix: #993

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -72,6 +72,9 @@ const (
 	CARootPrivateKey = "ca.key"
 	OldCARootCert    = "old-ca.crt"
 
+	// Client CA ConfigMap.
+	ClientCACert = "client-ca.crt"
+
 	// Certs.
 	CertExpirationYears  = 10
 	CACertExpiration     = 10 * 365 * 24 * time.Hour
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -14,6 +14,7 @@ const (
 	PolicyServerPort                                = 8443
 	PolicyServerMetricsPortEnvVar                   = "KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT"
 	PolicyServerMetricsPort                         = 8080
+	PolicyServerReadinessProbePort                  = 8081
 	PolicyServerReadinessProbe                      = "/readiness"
 	PolicyServerLogFmtEnvVar                        = "KUBEWARDEN_LOG_FMT"
 
--- a/internal/controller/policyserver_controller_deployment.go
+++ b/internal/controller/policyserver_controller_deployment.go
@@ -29,7 +29,9 @@ const (
 	sourcesVolumeName                = "sources"
 	verificationConfigVolumeName     = "verification"
 	kubewardenCAVolumeName           = "kubewarden-ca-cert"
+	kubewardenCAVolumePath           = "/ca"
 	clientCAVolumeName               = "client-ca-cert"
+	clientCAVolumePath               = "/client-ca"
 	secretsContainerPath             = "/pki"
 	imagePullSecretVolumeName        = "imagepullsecret"
 	dockerConfigJSONPolicyServerPath = "/home/kubewarden/.docker"
@@ -317,6 +319,16 @@ func (r *PolicyServerReconciler) configureMutualTLS(ctx context.Context, policyS
 	)
 
 	admissionContainer := &policyServerDeployment.Spec.Template.Spec.Containers[0]
+
+	admissionContainer.VolumeMounts = append(
+		admissionContainer.VolumeMounts,
+		corev1.VolumeMount{
+			Name:      kubewardenCAVolumeName,
+			MountPath: kubewardenCAVolumePath,
+			ReadOnly:  true,
+		},
+	)
+
 	if r.ClientCAConfigMapName != "" {
 		if err := r.Client.Get(ctx, types.NamespacedName{Name: r.ClientCAConfigMapName, Namespace: r.DeploymentsNamespace}, &corev1.ConfigMap{}); err != nil {
 			return fmt.Errorf("failed to fetch client CA config map: %w", err)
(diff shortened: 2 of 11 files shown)


internal/controller/policyserver_controller_deployment.go
```diff
@@ -343,6 +344,34 @@ func (r *PolicyServerReconciler) adaptDeploymentSettingsForPolicyServer(policySe
...
+					},
+				},
+			},
+		)
+
+		admissionContainer := &policyServerDeployment.Spec.Template.Spec.Containers[0]
+		admissionContainer.Env = append(admissionContainer.Env, corev1.EnvVar{
+			Name:  "KUBEWARDEN_CLIENT_CA_FILE",
+			Value: constants.ClientCACert,
+		})
```
flavio, line 373: Here we might have a problem. We have to inject two CAs: the one used by the Kubernetes API server and the one used by `audit-scanner`. Inside of policy-server we're planning to allow the same cli flag to be repeated multiple times. I don't think there's a way to achieve the same thing (provide multiple values) using the environment variable.

Maybe we have to change the deployment to use cli flags for that.

internal/constants/constants.go
```diff
@@ -72,6 +72,9 @@ const (
 	CARootPrivateKey = "ca.key"
 	OldCARootCert    = "old-ca.crt"
 
+	// Client CA ConfigMap.
+	ClientCACert = "client-ca.crt"
```
flavio, line 76: I would rename it to make it obvious that's the CA of the Kubernetes API server client certificate. Just to distinguish it from the CA used by `audit-scanner`

internal/controller/policyserver_controller_deployment.go
```diff
@@ -27,6 +27,7 @@ const (
 	policiesVolumeName               = "policies"
 	sourcesVolumeName                = "sources"
 	verificationConfigVolumeName     = "verification"
+	clientCAVolumeName               = "client-ca-cert"
```
flavio, line 30: Rename it to make it obvious that's the k8s API server CA. Also, always mount the internal Kubewarden CA, that's required to verify the requests issued by `audit-scanner`

flavio: Also, unrelated with this code line, we have to update the definition of the PolicyServer deployment to consume the `readiness` probe from a different port.

The Policy Server is always going to listen on two different ports:
- One without TLS for the readiness probe
- On with TLS (and eventually mTLS) for all the incoming requests

internal/controller/policyserver_controller_deployment.go
fabriziosestito: We could parse both the env variable and the cli flag with  the [value_delimiter](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_delimiter) set.

internal/constants/constants.go
fabriziosestito: I’d keep this variable as it is because it matches the webhook option name in controller-runtime. There's no need to separate audit-scanner and API server keys since the controller doesn’t use the audit-scanner concept and will just inject the internal CA for general use.

flavio: You're right, good points!

jvanz (approved):

internal/controller/policyserver_controller_deployment.go
```diff
@@ -345,6 +252,181 @@ func (r *PolicyServerReconciler) adaptDeploymentSettingsForPolicyServer(policySe
...
+							Key:  constants.CARootCert,
+							Path: constants.CARootCert,
+						},
+					},
+				},
+			},
+		},
+	)
+
+	admissionContainer := &policyServerDeployment.Spec.Template.Spec.Containers[0]
```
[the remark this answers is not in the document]
flavio: Can you explain what you're afraid of?

Are you concerned that something like OTEL changes the Deployment definition to inject its own container sidecar and, while doing that, the container number 0 is no longer Policy Server but the sidecar?

We always build our desired state of Deployment, which has only one container (ours). The mutating admission controllers will then mutate it afterwards

internal/controller/policyserver_controller_deployment.go
```diff
@@ -345,6 +252,181 @@ func (r *PolicyServerReconciler) adaptDeploymentSettingsForPolicyServer(policySe
...
+	if r.ClientCAConfigMapName != "" {
+		if err := r.Client.Get(ctx, types.NamespacedName{Name: r.ClientCAConfigMapName, Namespace: r.DeploymentsNamespace}, &corev1.ConfigMap{}); err != nil {
+			return fmt.Errorf("failed to fetch client CA config map: %w", err)
+		}
```
flavio, line 323: I think it's nice for us to bubble up this error. We could also have ignored that, the container would have stayed in status "pending" forever because the ConfigMap could not be mounted

internal/controller/policyserver_controller_test.go
```diff
@@ -660,6 +658,37 @@ var _ = Describe("PolicyServer controller", func() {
...
+
+			container := deployment.Spec.Template.Spec.Containers[0]
+
+			By("specifing the client ca certificate")
+			Expect(container.Env).To(ContainElement(MatchFields(IgnoreExtras, Fields{
+				"Name":  Equal("KUBEWARDEN_CLIENT_CA_FILE"),
+				"Value": Equal(fmt.Sprintf("%s,%s", constants.CARootCert, constants.ClientCACert)),
+			})))
+
+			Expect(deployment.Spec.Template.Spec.Volumes).To(ContainElement(MatchFields(IgnoreExtras, Fields{
```
flavio, line 676: We are not checking the Kubewarden's root CA is mounted into the container. Are we?

fabriziosestito: woops sorry! I have not commited the test

internal/controller/policyserver_controller_deployment.go
fabriziosestito: Yes, it is a better UX to bubble up the error. We do the same for the CA/certs secret in the other suberconciler.

jvanz (approved):

flavio (approved):
ship it!

viccuad (approved):
LGTM!

jvanz (approved):

viccuad (approved):

flavio (approved):

fabriziosestito:
blocked, waiting for https://github.com/kubewarden/policy-server/issues/1078

viccuad:
Merging, policy-server main ready and tests pass :rocket:

fabriziosestito:
@viccuad this was passing because of the dropme commit: https://github.com/kubewarden/kubewarden-controller/pull/1000/commits/2e07b8f1d7ecd8401fe1bafd41c44611c6b0bcfd, we still need to work on the multiple CAs

viccuad:
Argh, sorry there. Understood. Feel free to revert if needed..

viccuad:
Moving it back to blocked even if it is merged to keep track of it.

PR #1005, opened 2025-02-11
build(deps): disable cel-go bot updates 

fabriziosestito:
## Description

We should align the `cel-go` version with the one used in `k8s.io/apiserver`. Since `cel-go` is a direct dependency, a replace directive is required, and the Renovate bot should exclude it from updates

.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,9 @@ require (
 	sigs.k8s.io/controller-runtime v0.20.1
 )
 
+// CEL needs to be pinned to the same version as the one used by the k8s.io/apiserver package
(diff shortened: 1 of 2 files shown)


flavio (approved):
I just wonder if we're going to notice the upgrades done to the CEL version done by the future k8s releases

jvanz:
> I just wonder if we're going to notice the upgrades done to the CEL version done by the future k8s releases

Maybe we can add an item in the release checklist to check if there is a new CEL version

jvanz (approved):

viccuad (approved):

PR #1011, opened 2025-02-19
fix: mount and configure internal CA only with mTLS enabled

fabriziosestito:
We always mount and configure the internal CA in the PolicyServer deployment.
This is incorrect because it causes the webhook endpoints to require certificates issued by the internal CA.
This PR fixes the issue by enabling the internal CA and mTLS communication with the audit scanner only when a client CA is provided.

### DIFF
--- a/internal/controller/policyserver_controller_deployment.go
+++ b/internal/controller/policyserver_controller_deployment.go
@@ -300,35 +300,6 @@ func configureLabelsAndAnnotations(policyServerDeployment *appsv1.Deployment, po
 }
 
 func (r *PolicyServerReconciler) configureMutualTLS(ctx context.Context, policyServerDeployment *appsv1.Deployment) error {
-	policyServerDeployment.Spec.Template.Spec.Volumes = append(
-		policyServerDeployment.Spec.Template.Spec.Volumes,
-		corev1.Volume{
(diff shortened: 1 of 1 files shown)


internal/controller/policyserver_controller_deployment.go
```diff
@@ -372,10 +363,6 @@ func (r *PolicyServerReconciler) configureMutualTLS(ctx context.Context, policyS
 		return nil
 	}
 
-	admissionContainer.Env = append(admissionContainer.Env, corev1.EnvVar{
-		Name:  "KUBEWARDEN_CLIENT_CA_FILE",
-		Value: filepath.Join(kubewardenCAVolumePath, constants.CARootCert),
-	})
```
flavio, line 378: Don't we still need to set this environment variable?

fabriziosestito: no we don't, it is configured in the `if` block above alongside the user provided client CA

flavio: true, it was hidden by GH review :+1:

flavio (approved):
LGTM

viccuad (approved):
LGTM

jvanz (approved):

PR #1027, opened 2025-02-28
fix: client-ca filepath

fabriziosestito:
## Description

Fixes the client-ca file path to align with the server certificate location of the webhook server. Since Kubernetes does not allow a ConfigMap volume mount for client-ca to share the same path as the TLS certificates Secret, a subpath is used instead.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"flag"
 	"os"
+	"path/filepath"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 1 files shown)


jvanz (approved):

cmd/main.go
```diff
@@ -198,7 +199,7 @@ func setupManager(deploymentsNamespace string, metricsAddr string, probeAddr str
 
 	clientCAName := ""
 	if enableMutualTLS {
-		clientCAName = constants.ClientCACert
+		clientCAName = filepath.Join("client-ca", constants.ClientCACert)
```
viccuad, line 202:
```suggestion
                // The WebhookServer shares the same CertDir for both the server
                // certificate and the client CA certificate. We expect the ClientCACert
                // in the "client-ca"  sub-folder from the ConfigMap, since one cannot
                // mount several Secrets/ConfigMaps under the same path.
 		clientCAName = filepath.Join("client-ca", constants.ClientCACert)
```

viccuad (approved):
LGTM, just a quibble comment on why.

cmd/main.go
viccuad: Adding commit, as suggested if it was needed on Slack.

PR #1033, opened 2025-03-05
fix: expose Policy Server over port 443

flavio:
Leave the PolicyServer pods to listen on port 8443, but change the Service definition to map port 443 (of the service) to port 8443 (of the pod). Also, configure the `(Mutating|Validating)WebhookConfiguration` to reach the webhook server by using port 443.

This is done because when mTLS is enabled, the cluster operator has to provide a `kubeconfig` snippet that specifies which identity has to be used when interacting with webhook servers. By having Policy Server exposed over port 443 the `kubeconfig` configuration is simplified.

Without this change, the `kubeconfig` file would look like that:

```yaml
apiVersion: v1
kind: Config
users:
- name: 'kubewarden-controller-webhook-service.kubewarden.svc'
  user:
    client-certificate: /etc/mtls/domain.crt
    client-key: /etc/mtls/domain.key
- name: "*.kubewarden.svc:8443"
  user:
    client-certificate: /etc/mtls/domain.crt
    client-key: /etc/mtls/domain.key
```

With this change the `kubeconfig` would look like that:

```yaml
apiVersion: v1
kind: Config
users:
- name: '*.kubewarden.svc'
  user:
    client-certificate: /etc/mtls/domain.crt
    client-key: /etc/mtls/domain.key
```

**Important:** the service is still going to listen on the legacy 8443 port for a couple of Kubewarden's releases. Forcing a sudden switch from port 8443 to 443 could cause a temporary disservice inside of a cluster that is being upgraded. There's going to be a short window during which the Service will expose the Policy Server on port 443, but the webhook configuration is still using port 8443. We want API's servers requests to be regularly processed during the time taken by the controller to reconcile the webhook configurations.

In a couple of months we can change the Service to expose Policy Server only on port 443.


jvanz (approved):

viccuad (approved):
LGTM. We could open a card already for 1.25 on the removal of the legacy port.

flavio:
I've created https://github.com/kubewarden/kubewarden-controller/issues/1035 to keep track of the technical debt

PR #1037, opened 2025-03-06
policy server use kubernetes recommended labels

flavio:
All the resources associated with Policy Server are now using the labels recommended by Kubernetes, see [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).

### DIFF
--- a/internal/controller/policyserver_controller_deployment.go
+++ b/internal/controller/policyserver_controller_deployment.go
@@ -297,6 +297,10 @@ func configureLabelsAndAnnotations(policyServerDeployment *appsv1.Deployment, po
 	}
 	policyServerDeployment.Labels[constants.AppLabelKey] = policyServer.AppLabel()
 	policyServerDeployment.Labels[constants.PolicyServerLabelKey] = policyServer.Name
+
+	for key, value := range policyServer.CommonLabels() {
+		policyServerDeployment.Labels[key] = value
+	}
 }
 
 func (r *PolicyServerReconciler) configureMutualTLS(ctx context.Context, policyServerDeployment *appsv1.Deployment) error {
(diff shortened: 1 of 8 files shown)


viccuad (approved):
LGTM

internal/controller/policyserver_controller_deployment.go
```diff
@@ -297,6 +297,10 @@ func configureLabelsAndAnnotations(policyServerDeployment *appsv1.Deployment, po
 	}
 	policyServerDeployment.Labels[constants.AppLabelKey] = policyServer.AppLabel()
 	policyServerDeployment.Labels[constants.PolicyServerLabelKey] = policyServer.Name
```
viccuad, line 299: We should add a card to schedule removing the old labels in the future.

jvanz (approved):

internal/controller/policyserver_controller_deployment.go
flavio: Even better, we have a dedicated PR: https://github.com/kubewarden/kubewarden-controller/pull/1038

This is currently under the blocked column and is assigned to Kubewarden 1.25

internal/controller/policyserver_controller_test.go
```diff
@@ -87,6 +87,10 @@ var _ = Describe("PolicyServer controller", func() {
 					"TolerationSeconds": PointTo(Equal(tolerationSeconds)),
 				}),
 			}))
+
+			for k, v := range policyServer.CommonLabels() {
+				Expect(deployment.Spec.Template.ObjectMeta.Labels).To(HaveKeyWithValue(k, v))
+			}
```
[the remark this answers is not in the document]
fabriziosestito: I don't think there is a better way to do this since gomega `Keys` is used to match literal keys. 

However, once we remove the legacy labels, we could be strict in this assertion:

```go
Expect(deployment.Spec.Template.ObjectMeta.Labels).To(Equal(policyServer.CommonLabels()))

```

fabriziosestito (approved):

PR #1043, opened 2025-03-13
dev: fix tilt environment

flavio:
Do not configure the registry that holds the development image of kubewarden-controller as the place from which all the images of the `kubewarden-controller` helm chart have to be pulled from. This would cause the deployment to fail, since all the other images (like kubectl, audit-scanner) are not mirrored on the local registry.
Instead, patch the image attribute of the Deployment.

Add the ability to set labels of the `kubewarden` namespace created by Tilt. This allows to test different PSA profiles and tune our deployment.

### DIFF
--- a/tilt-settings.yaml.example
+++ b/tilt-settings.yaml.example
@@ -3,3 +3,9 @@ image: <your github handle>/kubewarden-controller
 helm_charts_path: ../helm-charts/
 audit_scanner_path: ../audit-scanner/
 controller_values_file: your-kubewarden-controller-helm-chart-values.yaml
+
+# This can be used to add labels to the `kubewarden` namespace.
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):

tilt-settings.yaml.example
```diff
@@ -3,3 +3,5 @@ image: <your github handle>/kubewarden-controller
 helm_charts_path: ../helm-charts/
 audit_scanner_path: ../audit-scanner/
 controller_values_file: your-kubewarden-controller-helm-chart-values.yaml
+namespace_labels:
```
fabriziosestito, line 6: could you add a comment to describe why we would need to use this setting?

flavio: done

Issue #1019, opened 2025-02-26
Implement Policy Lifecycle RFC


fabriziosestito:
This epic tracks the implementation of the [Policy Lifecycle RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0022-policy-lifecycle.md).

# Summary

This change addresses a critical issue with the Policy Server.
Currently, a policy can fail to load for a variety of reasons. For example, an inability to pull from the registry or incorrect settings. This causes the Policy Server to exit with an error, causing the pod to enter a CrashLoopBackOff state.
Resolving this requires inspecting the error messages from the Policy Server pod and fixing the underlying issue.
However, a user might have the permissions to create a policy but lack the necessary permissions to check the logs of the Policy Server. This presents challenges to diagnosis and resolution of problems.
Additionally, when updating a policy, there is no status reported to indicate the failure. The policy remains in the Active state even while the Policy Server is stuck in a crash loop.
This creates a significant risk, as new Policy Server pods might fail to start, and the old ones running the previous functional configuration could be lost if their node becomes unavailable.
This situation can disrupt the cluster, as all incoming admission requests will be denied in the absence of operational Policy Server instances.

The proposed changes aim to address these issues by introducing a new policy lifecycle that includes the following features:

- Hot reload of policies: The Policy Server will be able to load a new policy without requiring a restart.
- Policy validation: The Policy Server will validate the policy before loading it, ensuring correct pre-compilation and loading.
- Policy status monitoring: The Policy Server will report the status of the policy, indicating whether it was loaded successfully or if an error occurred.
- Policy versioning: The Policy Server will keep running with a previous version of the policy in case of issues with the latest policy.
- Policy rollback: The user will be able to rollback to a previous policy version in case of issues with the latest policy.

fabriziosestito:
> I  just reread the RFC and this epic. It looks good overall. I just have a question about the PolicyRevision CRDs. In the RFC, we state that we will use a single CRD for both cluster and namespaced policies. However, in the epic, we have a sub-issue to create two PolicyRevision CRDs: one for cluster-wide policies and another for namespaced ones. This approach is listed as an "alternative" in the RFC document. I cannot remember if we decided to adopt the alternative solution. If so, the epic looks fine to me. Otherwise, we need to update the sub-issue.

We decided to go with the Namespaced and cluster-scoped alternative, but the RFC was not updated. @flavio do you remember the decision?

fabriziosestito:
[...]

PR #1084, opened 2025-04-14
   fix: handle immutable field updates.

### LINKED ISSUE #1081: Default policy server pod is not restarted by controller [upgrade]
I install kubewarden 1.22 (or 1.23). Then I upgrade to charts from main and use `latest` images.

### Upgrade chart arguments:
```
DEFAULTS_ARGS="--set policyServer.image.tag=latest"
CONTROLLER_ARGS="--set image.tag=latest --set auditScanner.image.tag=latest"
```

### Problem:
When I use `--set image.tag=latest` on controller then
- policy-server is not restarted when I deploy new policy
- policy-server is not restarted after upgrade

In controller logs:
```json
{
  "level": "error",
  "ts": "2025-04-10T12:44:16Z",
  "msg": "Reconciler error",
  "controller": "policyserver",
  "controllerGroup": "policies.kubewarden.io",
  "controllerKind": "PolicyServer",
  "PolicyServer": {
    "name": "default"
  },
  "namespace": "",
  "name": "default",
  "reconcileID": "362434d1-68a6-48b4-a4c4-5de755d7da4e",
  "error": "error reconciling policy-server deployment: Deployment.apps \"policy-server-default\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"app.kubernetes.io/instance\":\"policy-server-default\", \"app.kubernetes.io/part-of\":\"kubewarden\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable",
  "stacktrace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:347\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255"
}
```

Policy-server restarts as expected if I don't use latest controller image.
Also policy-server restarts work ok if I install with latest images directly, without upgrade.

Probably caused by https://github.com/kubewarden/kubewarden-controller/pull/1038 that was merged yesterday.
Failing nightly job: https://github.com/kubewarden/helm-charts/actions/runs/14366897328/job/40281921279

jvanz:
## Description

In a recent change the controller start to add all the recommended label
in the resources along side with the previous "app" label. To goal is to
remove the "app" label in a near future. With that in mind, in the
commit c43fad86d7ab6336c635d7f0d1ab9e874b5a6fd2 we remove the usage of
the "app" label and update all the selector to use only the recommended
labels. However, the label selector used in the spec of the deployment is
not mutable. This means that we cannot update/patch this field. We must
recreate the deployment. But if we do that, we can cause down time while
the old deployment is being deleted and the new one is not ready.
Furthermore, we cannot create the new one with the expected name before
delete the old deployment.

This commit change the deployment reconciler to handle the update of
immutable fields by creating a new deployment, wait it to be ready,
delete the old deployment, and then restore the right name by creating
another deployment with the latest configuration and expected name. And
delete the temporary deployment in the end. Therefore, we ensure that
during the whole update process some policy server will be ready to
anwser request from the control plane.
    

Fix #1081 

## Test

1. Install Kubewarden 1.23
2. Upgrade to latest container image (controller should use an image with the changes from this branch)


jvanz:
Moving to draft. I think there is a better solution that does not require more releases to remove the app label

jvanz:
@kubewarden/kubewarden-developers, I aware codecov will complain about the missing tests. To fix that I need to add unit test for each new function. However, the proper way to test this with e2e. Therefore, I would like to give you the change to review the logic/idea before starting test to the function that may change soon.

jvanz:
Okay, an update after the daily call with the @kubewarden/kubewarden-developers. We've decided to hold off on this for now. We plan to revisit the [policy lifecycle](https://github.com/kubewarden/kubewarden-controller/issues/1019) in future releases, which will be a major breaking change, and we'll use that opportunity to remove the old "app" label from the deployment. Consequently, I'll close this PR and open a new one that keeps the "app" label alongside all the other changes, adding the standard label to the resources and the deployment.

PR #1136, opened 2025-06-03
chore(linter): enable wrapcheck linter.

jvanz:
## Description

The wrapcheck linter is disable in the default linter configuration we adopted. But we want to enable it. This commit re enable the linter and fix all the errors it reported.


viccuad (approved):
LGTM, thanks.

.golangci.yml
```diff
@@ -130,7 +130,7 @@ linters:
     #- prealloc # [premature optimization, but can be used in some cases] finds slice declarations that could potentially be preallocated
     #- tagalign # checks that struct tags are well aligned
     #- varnamelen # [great idea, but too many false positives] checks that the length of a variable's name matches its scope
-    #- wrapcheck # checks that errors returned from external packages are wrapped
+    - wrapcheck # checks that errors returned from external packages are wrapped
```
flavio: Don't uncomment it here, write it below around line 163. It's important to keep our customization well define, it makes life easier when moving to the next version of the golden config

Makefile
```diff
@@ -202,7 +202,7 @@ GINKGO ?= $(LOCALBIN)/ginkgo-$(GINKGO_VERSION)
 KUSTOMIZE_VERSION ?= v5.4.1
 CONTROLLER_TOOLS_VERSION ?= v0.16.1
 ENVTEST_VERSION ?= release-0.18
-GOLANGCI_LINT_VERSION ?= v2.0.2
```
flavio, line 205: Don't forget to update also https://github.com/kubewarden/kubewarden-controller/blob/749f376729cbbeb2ea129d9da8b01ae1cbed3ccf/.github/workflows/ci.yml#L62

This is such a pain... it would be great to have an updatecli automation for that, but I don't know how easy it would be

cmd/main.go
```diff
@@ -252,7 +252,10 @@ func setupManager(deploymentsNamespace string, metricsAddr string, probeAddr str
 			ClientCAName: clientCAName,
 		}),
 	})
-	return mgr, err
+	if err != nil {
+		return mgr, errors.Join(errors.New("failed to setup manager"), err)
```
flavio: Do not `errors.Join` to wrap errors, this has a different purpose.

You should use `fmt.Errorf` with the `%w` directive. This allows to use `errors.Is(err, otherErrorType)`, plus IMHO provides a better way to tailor the final user message

jvanz: Ok, I'll do the change. But, JFYI, `errors.Is` also works with errors returned by `Join` [method](https://pkg.go.dev/errors#Join)

internal/controller/admissionpolicy_controller.go
```diff
@@ -61,7 +61,10 @@ type AdmissionPolicyReconciler struct {
 func (r *AdmissionPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var admissionPolicy policiesv1.AdmissionPolicy
 	if err := r.Get(ctx, req.NamespacedName, &admissionPolicy); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, errors.Join(errors.New("failed to get admission policy"), err)
```
flavio: No join please

internal/controller/admissionpolicygroup_controller.go
```diff
@@ -61,7 +61,10 @@ type AdmissionPolicyGroupReconciler struct {
 func (r *AdmissionPolicyGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var admissionPolicyGroup policiesv1.AdmissionPolicyGroup
 	if err := r.Get(ctx, req.NamespacedName, &admissionPolicyGroup); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, errors.Join(errors.New("failed to get admission policy group"), err)
```
flavio: No join please

internal/controller/clusteradmissionpolicy_controller.go
```diff
@@ -61,7 +61,10 @@ type ClusterAdmissionPolicyReconciler struct {
 func (r *ClusterAdmissionPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var clusterAdmissionPolicy policiesv1.ClusterAdmissionPolicy
 	if err := r.Get(ctx, req.NamespacedName, &clusterAdmissionPolicy); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, errors.Join(errors.New("failed to get cluster admission policy"), err)
```
flavio: No join please

internal/controller/clusteradmissionpolicygroup_controller.go
```diff
@@ -61,7 +61,10 @@ type ClusterAdmissionPolicyGroupReconciler struct {
 func (r *ClusterAdmissionPolicyGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var clusterAdmissionPolicy policiesv1.ClusterAdmissionPolicyGroup
 	if err := r.Get(ctx, req.NamespacedName, &clusterAdmissionPolicy); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, errors.Join(errors.New("failed to get cluster admission policy group"), err)
```
flavio: No join please

internal/controller/policyserver_controller.go
```diff
@@ -83,7 +83,10 @@ type TelemetryConfiguration struct {
 func (r *PolicyServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var policyServer policiesv1.PolicyServer
 	if err := r.Get(ctx, req.NamespacedName, &policyServer); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, errors.Join(errors.New("failed to get policy server"), err)
```
flavio: No join please

internal/controller/policyserver_controller_cert_secret.go
```diff
@@ -51,7 +51,7 @@ func (r *PolicyServerReconciler) reconcilePolicyServerCertSecret(ctx context.Con
 			var caCert, caPrivateKey []byte
 			caCert, caPrivateKey, err = certs.ExtractCARootFromSecret(caSecret)
 			if err != nil {
-				return err
+				return errors.Join(fmt.Errorf("failed to extract the CA from secret: %s", caSecret.GetName()), err)
```
flavio: No join please

flavio (changes requested):

cmd/main.go
flavio: thanks for pointing that out, I didn't realize that. I still prefer to use `fmt` because it provides IMHO a nicer output

flavio:
True, then let's leave the join until we figure it out

flavio:
summoning someone opinionated @fabriziosestito :smile:

jvanz:
@kubewarden/kubewarden-developers In the end of the day, both are equivalent. They both wrap the errors in the same way. However, in our case, maybe `fmt.ErrorF` is a better option. As far as I can see, we wrap error for two main reason:

- Make linter happy
- Write nicer error messages.

As we in the majority of the cases we  log the error and continue, and we do not have an exposed API to third party tools, the error chain does not care that much. And we do not need to worry too much about exposing implementation details (internal errors). Therefore, for our code base, it's more a matter of taste.

jvanz:
@flavio, as I mentioned in my previous [comment](https://github.com/kubewarden/kubewarden-controller/pull/1136#issuecomment-2940859664), after considering our use case, I agree with using fmt.Errorf for our codebase. Because of this, I've changed the errors.Join calls to fmt.Errorf. I initially used errors.Join because it was the most common method in the file where I started fixing linter errors, and I continued using it in other files without reviewing other error-wrapping approaches.

PR #1140, opened 2025-06-09
feat: custom error message.

### LINKED ISSUE #1139: Add support for custom error message in the policy CRDs
This is part of the implementation to allow users to define custom error messages for policies. This issue includes all the changes in the controller code base adding a new spec field to define a custom errors message, the policy server configmap generation including this new field, and the tests to cover these changes. 

### Acceptance criteria 
  - [ ] Update the `ClusterAdmissionPolicy` and the `AdmissionPolicy` CRDs: add the `spec.message` attribute. This attribute of type string and is optional
  - [ ] No work has to be done with regards to `ClusterAdmissionPolicyGroup` and `AdmissionPolicyGroup` since they already have a `message` field.
  - [ ] Update the Policy Server's ConfigMap generation to include the `spec.message` value if present

jvanz:
## Description

Adds a new message field to allow users define a custom error message for ClusterAdmissionPolicy and AdmissionPolicy types. This also involves of moving the code to access this field to merge with the current message field of the policy group types.

Fix #1139


api/policies/v1/factories.go
```diff
@@ -16,9 +16,10 @@ import (
 )
 
 const (
-	integrationTestsFinalizer   = "integration-tests-safety-net-finalizer"
-	defaultKubewardenRepository = "ghcr.io/kubewarden/policy-server"
-	maxNameSuffixLength         = 8
+	integrationTestsFinalizer      = "integration-tests-safety-net-finalizer"
+	defaultKubewardenRepository    = "ghcr.io/kubewarden/policy-server"
+	maxNameSuffixLength            = 8
+	defaultPolicyGroupErrorMessage = "error message"
```
flavio: maybe we can do something like that:

```golang
defaultPolicyGroupMessage = "policy group default rejection message"
```

api/policies/v1/policy_types.go
```diff
@@ -149,6 +149,12 @@ type PolicySpec struct {
 	// +optional
 	// +kubebuilder:default:=10
 	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
+
+	// Message is  used to specify the message that will be returned when
+	// the policy is rejected. The specific policy results will be
+	// returned in the warning field of the response.
```
flavio: We should do the same thing we do with the group policies. We are not using the warnings fields because they have some limitations (the length of the message and some rules around the IDs of the messages).

We should add them to the `.status.details.causes`. See [here](https://docs.kubewarden.io/explanations/policy-groups#the-message-attribute-and-the-response-format).

Can you update the inline docs on this PR?

```suggestion
	// Message overrides the rejection message of the policy.
	// When provided, the policy's rejection message can be found
	// inside of the `.status.details.causes` field of the
	// AdmissionResponse object.
```

api/policies/v1/factories.go
```diff
@@ -16,9 +16,10 @@ import (
 )
 
 const (
-	integrationTestsFinalizer   = "integration-tests-safety-net-finalizer"
-	defaultKubewardenRepository = "ghcr.io/kubewarden/policy-server"
-	maxNameSuffixLength         = 8
+	integrationTestsFinalizer      = "integration-tests-safety-net-finalizer"
+	defaultKubewardenRepository    = "ghcr.io/kubewarden/policy-server"
+	maxNameSuffixLength            = 8
+	defaultPolicyGroupErrorMessage = "policy group default rejection message"
```
flavio: `defaultPolicyGroupRejectionMessage`

flavio (approved):

PR #1142, opened 2025-06-13
fix: remove legacy policy server service port from configuration.

jvanz:
Removes the old policy server service port that was used before adding
port 443. This port was kept to avoid issues during the upgrade to
Kubewarden v1.23. Now, after several releases since switching to port
443, we can remove the old port from the service configuration.

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>


flavio (approved):
Let's not forget to add a mention inside of the blog post and the release notes

Issue #1165, opened 2025-07-24
Implement new `spec.timeoutEvalSeconds` for policies


viccuad:
### Acceptance criteria

- Add new optional `spec.timeoutEvalSeconds` for all policies and policygroups sub-policies. Contrary to `spec.timeoutSeconds` which specifies the timeout for the associated webhook, this new field specifies the timeout for the evaluation of the policy.
- PolicyGroups sub-policies can each have a `spec.timeoutEvalSeconds`, but the whole PolicyGroup cannot have a global timeoutEvalSeconds. This allows simplifying validation of policies in groups having longer timeoutEvalSeconds than the PolicyGroup itself, plus makes the UX of sub-policies closer to normal policies.
- Validate the old timeoutSeconds so it doesn't exceed the webhook timeout from Kubernetes.
- Validate the new timeoutSeconds so it doesn't exceed the webhook timeout from Kubernetes.
- Validate the new timeoutEvalSeconds so it doesn't exceed old timeoutSeconds.
- Provide tests as needed.
- Update reference and docs.kubewarden.io as needed, particularly the policies.yaml file for policy-server.

flavio:
While in the beginning I was a bit skeptical about having each member of a policy group to have its own timeout, I now changed my mind. I think we can go ahead with that.

PR #1187, opened 2025-08-25
feat: Add new spec.timeoutEvalSeconds for policies and policy group members 

viccuad:
## Description

Closes https://github.com/kubewarden/kubewarden-controller/issues/1165

- Add a new `spec.timeoutEvalSeconds` for (Cluster)AdmissionPolicies and (Cluster)AdmissionPolicyGroup policy members. These new keys specify the timeout for the policy evaluation. After the timeout passes, the policy evaluation call will fail based on the failure policy. The timeout value must be between 1 and 30 seconds.
- In the case of (Cluster)AdmissionPolicyGroups, only the members have this new field, now the group itself.
- Add validations for timeoutSeconds and timeoutEvalSeconds: neither timeoutSeconds nor timeoutEvalSeconds can be greater that the max webhook timeout from kubernetes (hardcoded at 30s).

## Test

Added 2 unit tests, for (Cluster)AdmissionPolicies and (Cluster)AdmissionPolicyGroups, of the validating functions.

### DIFF
--- a/internal/controller/policyserver_controller_test.go
+++ b/internal/controller/policyserver_controller_test.go
@@ -264,6 +264,8 @@ var _ = Describe("PolicyServer controller", func() {
 		})
 
 		It("should create the policy server configmap with the assigned policies", func() {
+			timeoutEvalSeconds := int32(5)
+
 			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).Build()
 			createPolicyServerAndWaitForItsService(ctx, policyServer)
 
@@ -275,6 +277,7 @@ var _ = Describe("PolicyServer controller", func() {
 
 			clusterAdmissionPolicy := policiesv1.NewClusterAdmissionPolicyFactory().
 				WithPolicyServer(policyServerName).
+				WithTimeoutEvalSeconds(&timeoutEvalSeconds).
 				WithContextAwareResources([]policiesv1.ContextAwareResource{
 					{
 						APIVersion: "v1",
@@ -298,7 +301,8 @@ var _ = Describe("PolicyServer controller", func() {
 				WithMembers(policiesv1.PolicyGroupMembersWithContext{
 					"pod_privileged": {
 						PolicyGroupMember: policiesv1.PolicyGroupMember{
-							Module: "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
+							Module:             "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
+							TimeoutEvalSeconds: &timeoutEvalSeconds,
 						},
 						ContextAwareResources: []policiesv1.ContextAwareResource{
 							{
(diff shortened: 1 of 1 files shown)


api/policies/v1/policy_validation_test.go
```diff
@@ -596,3 +597,73 @@ func TestValidatePolicyModeField(t *testing.T) {
...
+					found := false
+					for _, err := range errs {
+						if strings.Contains(err.Error(), expected) {
+							found = true
+							break
+						}
+					}
+					require.True(t, found, "expected error containing %q, got %v", expected, errs)
+				}
+			}
```
jvanz, line 666: You can do a similar logic using the [filter](https://github.com/kubernetes/apimachinery/blob/v0.33.4/pkg/util/validation/field/errors.go#L366) function and checking the len of the returned list with the require package. To be honest, I'm not sure if this will be better and make the code more readable, but I think it's nice the share the filter function here to future changes. The same applies for the policy group test. ;)

Therefore, I do not consider this comment a blocker to merge this PR. Just sharing a idea for our collective memory. ;)

viccuad: That's a neat trick! Gave it a try, but then one cannot match the error message as the filter drops the errors.
I will keep it as it is, given that there's 2 errors with the same error field, so dropping an error for that error field means we don't exactly know which of the 2 errors was emitted.

api/policies/v1/policygroup_validation.go
```diff
@@ -138,3 +140,40 @@ func validatePolicyGroupExpressionField(policyGroup PolicyGroup) *field.Error {
 
 	return nil
 }
+
+// validatePolicyGroupMembersTimeouts checks the timeouts so that:
+//   - the group's timeoutSeconds is not greater than the Kubernetes webhook max timeout (30s).
+//   - each member's timeoutEvalSeconds is not greater than the group's
+//     timeoutSeconds nor the Kubernetes webhook max timeout (30s).
+func validatePolicyGroupMembersTimeouts(policyGroup PolicyGroup) field.ErrorList {
```
flavio, line 150: I think this is good enough, but if we wanted to be fancy we could also ensure the sum of all the member's timeoutEvalSeconds is not going to be greater than the group's timeoutSeconds nor the k8's max timeout value

internal/controller/policyserver_controller_test.go
```diff
@@ -263,7 +263,9 @@ var _ = Describe("PolicyServer controller", func() {
 			})))
 		})
 
-		It("should create the policy server configmap with the assigned policies", func() {
+		FIt("should create the policy server configmap with the assigned policies", func() {
```
flavio: I think the `FIt` is a leftover from your development time, you should go back to `It`

flavio (changes requested):

jvanz (approved):

internal/controller/policyserver_controller_test.go
viccuad: Woops.. rebased.

api/policies/v1/policygroup_validation.go
viccuad: Good point. Rebased and added that logic, plus the unit tests that check it.

flavio: thanks!

flavio (approved):

Issue #1209, opened 2025-09-25
timeoutEvalSeconds does not check minimum value


kravciak:
In the description we mention `The timeout value must be between 1 and 30 seconds.`
We check maximum value, but not minimum:

This allows me to set following values, which produces policy-server error.
```
timeoutEvalSeconds: 0
Error: WebAssembly failure: cannot rehydrate PolicyEvaluatorPre: unable to rehydrate wapc module:
cannot build Wapc host: Initialization failed: Initialization failed: wapc_init init interrupted, execution deadline exceeded

timeoutEvalSeconds: -1
Error: error while loading policies from "/config/policies.yml":
data did not match any variant of untagged enum PolicyOrPolicyGroup
```

### Expected
We should limit minimum value on crds.

PR #1211, opened 2025-09-25
Check timeouts for mininum value

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/1209

## Test

Added several unit tests.


jvanz (approved):
LGTM. But I have one thing in mind. Can we add a test in the e2e test to avoid a regression? As this validation is in the CRDs, there is not way to do it in the current integration tests.  😬

api/policies/v1/policy_types.go
```diff
@@ -152,13 +152,17 @@ type PolicySpec struct {
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=30
 	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
 
 	// TimeoutEvalSeconds specifies the timeout for the policy evaluation. After
 	// the timeout passes, the policy evaluation call will fail based on the
 	// failure policy.
 	// The timeout value must be between 1 and 30 seconds.
 	// +optional
+	// +kubebuilder:validation:Minimum=1
```
flavio: I would raise it to be 2 seconds, which is the default value used by Policy Server when nothing is specified by the user.

1 second could be a bit too tight for certain policies

flavio (changes requested):

api/policies/v1/policy_types.go
viccuad: Rebased and raised to 2 seconds.

viccuad:
Rebased onto main for merging.

jvanz:
> I think it would be nice, but let's not try to cover all the possible cases. We're relying on a feature of Kubernetes, we can assume this is tested properly. We just have to make sure we didn't put the wrong annotation in there

> I agree, I prefer the kubebuilder tags. We could check that they work (or for their presence) in e2e tests but it seems overkill. If we prefer that, I pushed the previous changes to #1213.

When I've suggested the e2e test is not to test the validation itself. It's to ensure that we will not remove the kubebuilder tag by accident (e.g. rebase, conflict resolving, etc). However, when I've wrote my review feedback I forgot that the tests installed the CRDs from the [local directory](https://github.com/kubewarden/kubewarden-controller/blob/b55af9f5f5f15f8e28c27a305257be99a7025dc4/internal/controller/suite_test.go#L73). That's why I've suggested the e2e test. 

However, as the CRDs installed in the tests are from the local directory. We can add a very simple Ginkgo test that try to deploy a invalid policy settings and check if the validation is performed. Therefore, we ensure that will will not remove the kubebuilder tag by accident. Thus, e2e test and #1213 are not necessary.

flavio (approved):

PR #1214, opened 2025-09-26
test: Add integration tests to check all kubebuilder:validation tags

viccuad:
## Description

Add 2 tests that cover all current kubebuilder:validation tags.

The Kubebuilder tags are on the types:

`PolicySpec{}`: shared by AdmissionPolicies and ClusterAdmissionPolicies
`PolicyGroupmember{}`, `GroupSpec{}`: shared by AdmissionPolicyGroups and ClusterAdmissionPolicyGroups

Hence create 1 testcase per namespaced and clusterwide policy type.

## Test


api/policies/v1/factories.go
```diff
@@ -441,6 +447,24 @@ func (f *ClusterAdmissionPolicyGroupFactory) WithTimeoutSeconds(timeout *int32)
 	return f
 }
 
+func (f *ClusterAdmissionPolicyGroupFactory) WithTimeoutEvalSeconds(timeout *int32) *ClusterAdmissionPolicyGroupFactory {
+	member := f.policyMembers["pod_privileged"]
```
viccuad, line 445: These are not following the factory pattern, but I think it's simpler than to create a factory for policyMembers at this point in time.

internal/controller/clusteradmissionpolicy_controller_test.go
```diff
@@ -33,6 +33,33 @@ import (
...
 	ctx := context.Background()
 
+	When("creating a wrong Policy should fail because of CRD validation", Ordered, func() {
+		// this test triggers all the kubebuilder:validation tags for PolicySpec{}
+
+		var policyServerName string
+		var policyName string
+		var policy *policiesv1.ClusterAdmissionPolicy
+
+		BeforeAll(func() {
```
jvanz: Why using `BeforeAll` instead of `It`?

jvanz (changes requested):
I have a question about how you wrote the test.

internal/controller/clusteradmissionpolicy_controller_test.go
viccuad: Good point, this should indeed by a It(). Refactored, and separated the cases further (so we actually test each of the cases instead of all of them failing at once).

viccuad:
Ready for review. This is a doozy.

Removed the test cases verifying that `spec.module` is required. That they are required is working as expected, checked the code, checked the resulting crds (see [here](https://github.com/kubewarden/kubewarden-controller/blob/2e6d8e39b61da0cea2627d12dbaac522da981af6/config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml#L489)).

The thing is, this kubebuilder/controller-gen markers (`+required` or  `+kubebuilder:validation:Required`) have no effect. The `spec.module` field is set as required, as it doesn't have any json tags such as `omitempty`.

I have checked the code, checked the testcases, verified that the testcases TestEnv correctly consumes the CRDs definition, verified that we are running these specific tests as `in-cluster` so the k8s api is the one verifying the CR instantiation, etc. Yet I cannot trigger the testcase and have it correctly fail when a `spec.module` is not passed.

Hence, I have removed these `spec.module` tests are left the others, which correctly test the kubebuilder markers.

In any case, since we check the CRDs YAMLs into git, if we were to mistag the `spec.module` we would realize then.

api/policies/v1/factories.go
```diff
@@ -217,6 +217,11 @@ func (f *ClusterAdmissionPolicyFactory) WithTimeoutEvalSeconds(timeout *int32) *
 	return f
 }
 
+func (f *ClusterAdmissionPolicyFactory) WithNoModule() *ClusterAdmissionPolicyFactory {
+	f.module = ""
+	return f
+}
+
```
jvanz: As we do not test this field. We can remove this function

api/policies/v1/factories.go
```diff
@@ -441,6 +447,24 @@ func (f *ClusterAdmissionPolicyGroupFactory) WithTimeoutSeconds(timeout *int32)
...
+	member := f.policyMembers["pod_privileged"]
+	member.TimeoutEvalSeconds = timeout
+	f.policyMembers["pod_privileged"] = member
+	return f
+}
+
+func (f *ClusterAdmissionPolicyGroupFactory) WithNoModule() *ClusterAdmissionPolicyGroupFactory {
+	f.module = ""
+	return f
+}
```
jvanz: As we do not test this field. We can remove this function

api/policies/v1/factories.go
```diff
@@ -441,6 +447,24 @@ func (f *ClusterAdmissionPolicyGroupFactory) WithTimeoutSeconds(timeout *int32)
...
+
+func (f *ClusterAdmissionPolicyGroupFactory) WithNoGroupModule() *ClusterAdmissionPolicyGroupFactory {
+	member := f.policyMembers["pod_privileged"]
+	member.Module = ""
+	return f
+}
```
jvanz: As we do not test this field. We can remove this function

viccuad:
Thanks! Rebased and dropped those unneeded functions, plus reworded the commit message as its scope was reduced.

PR #1235, opened 2025-10-27
fix(ci): add second cosign signature for compatibility

jvanz:
## Description

In order to allow old cosign version and other verification tools to verify the signature it's necessary to add a second signature using the old format. The default format before cosign v3 changed the default signature bundle.

Updates the CI to sign the blobs using only cosign v3 signature bundle format.


flavio (approved):
I like this PR. I like the strategy of signing with v2 and v3 format inside of the OCI registry, but sign only with v3 the blobs that are going to be made available as assets of the GitHub Release

PR #1375, opened 2026-01-20
label-based test execution

flavio:
This PR modernizes the CI trigger mechanism by replacing commit message-based triggers with GitHub labels:

- Removed: `[ci full]` and `[full ci]` commit message triggers
- Added: Three new labels for on-demand test execution:
  - `ci-rust`: Runs all Rust jobs
  - `ci-go`: Runs all Go jobs
  - `ci-full`: Runs complete test suite (both Rust and Go)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,15 +77,6 @@ jobs:
             fi
           fi
 
-          # Check if commit message contains [ci full] or [full ci] to run all checks
-          COMMIT_MSG='${{ github.event.head_commit.message }}'
-          if echo "$COMMIT_MSG" | grep -qiE '\[(ci full|full ci)\]'; then
-            echo "Commit message contains [ci full] or [full ci], running all checks"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -10,6 +10,7 @@ on:
         default: true
   push:
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
viccuad, line 13: how does this work, does this mean that every time we relabel a PR the CI would run again?

What happens if we add or remove an unrelated PR and the detect changes job doesn't match anything, does that mean that now there's a new workflow run that doesn't run any jobs (and is the last, so it's shown on UI)?

flavio: it would run again. The detector would then run only the test that matter

jvanz: If it has the label, the CI will not check the changed files. Which means that the CI will run no matter the changes. So, the unrelated PR will run the CI even if there is not change related to the "requested CI to run".

Issue #1358, opened 2026-01-19
monorepo: ensure charts are pushed to central repository


flavio:
updatecli: assess what the "push helm chart to central repo" does, evaluate if we can make a plugin that is shared between kubewarden admission controller and the sbomscanner project

viccuad:
@kubewarden/kubewarden-developers some background on how we intend to reuse and simplify the admission controller release following the pattern of SBOMScanner, with a bit of changes.

As we talked on the meeting, we are going to try and move all the updatecli automation for release changes earlier in the process, This means that for release, we will:

1. We trigger the updatecli workflow. This runs updatecli, and opens a PR against kubewarden/kubewarden-controller with the following changes:
    - Bumps of helm-charts
    - Bumps of Cargo.* files
    - Bumps of READMEs.
    - Anything needed.
  This PR contains all the changes needed, happens prior to tags, can be reverted, opened against maintenance branches, etc.

2. Developers review and merge this PR. Upon merge, we either create a tag automatically, or we leave the tag to the developers.

3. Pushing the tag triggers a release job as it is right now. This release job builds all the images and artifacts, and publishes the GH release in kubewarden/kubewarden-controller.

4. The kubewarden/helm-charts repo gets a workflow dispatch to consume the newly changed files of the helm-charts from the kubewarden/kubewarden-controller repo, ideally by just checking out the tag. This workflow opens a PR against kubewarden/helm-charts, with these new changes.

5. Developers review and merge the helm-chart PR, triggering the release of the Helm charts.

PR #1380, opened 2026-01-22
ci(feat): Add updatecli pipeline to bump everything for release

viccuad:
## Description

Relates to https://github.com/kubewarden/kubewarden-controller/issues/1358

Refactor and augment the updatecli pipeline:
- rename `helm-chart-update.yaml -> open-release-pr.yaml`.
- Bump Helm chart `appVersion`, `version`, values.yaml keys.
- Bump kubewarden-controller README.md and schema.
- Bump `Cargo.*` versions.
- Open PR against the kubewarden-controller repo with the changes.

This updatecli pipeline takes the title of the last GH draft release to find the future stack version, and uses it to bump all needed things. Then opens a PR against the repo with all changes. Upon merge of that automated PR, one would create the tag.

## Test

Tested on my fork, see:
https://github.com/viccuad/kubewarden-controller/pull/3/files

### DIFF
--- a/updatecli/updatecli.release.d/open-release-pr.yaml
+++ b/updatecli/updatecli.release.d/open-release-pr.yaml
@@ -21,7 +21,6 @@ sources:
         # release-drafter, as its title has the prospective semver that we want
         # to tag
         draft: true
-      versionfilter:
         kind: latest
   newestReleaseTagVersion:
     name: Extract newest tag from newest GitHub release
@@ -33,7 +32,8 @@ sources:
       repository: "{{ .github.repo }}"
       token: "{{ requiredEnv .github.token }}"
       key: "tagname"
-      versionfilter:
+      typefilter:
+        draft: false
         kind: latest
 
   controllerChartVersion:


updatecli/updatecli.release.d/open-release-pr.yaml
```diff
@@ -0,0 +1,226 @@
...
+    # used for displaying a gh compare URL in the open PR
+    # TODO continue on fail (we may not have a tag)
+    kind: githubrelease
+    spec:
+      owner: "{{ .github.owner }}"
+      repository: "{{ .github.repo }}"
+      token: "{{ requiredEnv .github.token }}"
+      key: "tagname"
+      versionfilter:
+        kind: latest
```
viccuad, line 37: @olblak not important, but with these here I get:
```
DEBUG: version filter of kind "latest" not supported. Feel free to open an issue explaining your need
```

Which is contrary to what is documented here: https://www.updatecli.io/docs/plugins/resource/github_release/#_version_filter, as `latest` is the default and supposedly is supported.

updatecli/updatecli.release.d/open-release-pr.yaml
```diff
@@ -0,0 +1,226 @@
...
+
+targets:
+  update_kubewarden_crds_helm_version:
+    scmid: default
+    name: Update Helm chart kubewarden-crds version
+    kind: yaml
+    sourceid: crdsChartVersion
+    spec:
+      file: charts/kubewarden-crds/Chart.yaml
+      key: $.version
```
viccuad, line 123: I expect that the bumps of the chart `version` will be bugged, as if we do several RCs they will get bumped several times.

To fix this, we will need to follow the approach of having 1 updatecli release per case, as we do in the helm-charts repo..

https://github.com/kubewarden/helm-charts/tree/main/updatecli

updatecli/updatecli.release.d/open-release-pr.yaml
```diff
@@ -0,0 +1,226 @@
...
+        #           {{ range .pr.reviewers }}
+        - "{{ . }}"
+      #           {{ end }}
+      #     {{ end }}
+      #     {{ if .pr.labels }}
+      labels:
+        #         {{ range .pr.labels }}
+        - "{{ . }}"
+      #           {{ end }}
+      #     {{ end }}
```
flavio, line 112: I think we can get rid of the commented lines, these look like leftovers

viccuad: They are templating to pass the reviewers and labels values from `values.yaml`. This is also how it is done for the policies PR.

updatecli/updatecli.release.d/open-release-pr.yaml
jvanz: Maybe should we use the `typefilter.latest` flag?

updatecli/DEVELOPING.md
```diff
@@ -0,0 +1,28 @@
...
+as needed.
+
+Change `updatecli/values.yaml` as needed to target forks.
+
+```console
+$ cd kubewarden-controller/
+$ export UPDATECLI_GITHUB_TOKEN=<your token>
+$ clear; updatecli apply --config updatecli/updatecli.release.d/open-release-pr.yaml \
+  --values updatecli/values.yaml \
+  --debug --clean=true
```
jvanz, line 13: Maybe we can have CI to call the updatecli script. Therefore, no need to generate personal tokens. Something similar to what we have in the policies repository.

updatecli/updatecli.release.d/open-release-pr.yaml
```diff
@@ -0,0 +1,226 @@
...
+    transformers:
+      - semverinc: minor
+    spec:
+      file: charts/kubewarden-controller/Chart.yaml
+      key: $.version
+  crdsChartVersion:
+    name: "Get latest kubewarden-crds helm version"
+    kind: yaml
+    transformers:
+      - semverinc: minor
```
jvanz, line 51: Are we not be able to do minor/major releases? Shouldn't we expose the type of version bump using a values field?

updatecli/updatecli.release.d/open-release-pr.yaml
flavio: oh sorry...

updatecli/DEVELOPING.md
jvanz: Nerver mind, I just saw you other [PR](https://github.com/kubewarden/kubewarden-controller/pull/1384/changes)

I would say add an example on how to trigger that CI using `gh` cli tool. I let you decide if you want to do that on this PR or on the another.

updatecli/updatecli.release.d/open-release-pr.yaml
viccuad: We should be able to do major/minor/patch releases. I think using a values field would be nice for this.

For this PR, I decided to not add that functionality; I was scared of the test matrix and the need for several updatecli pipelines (as we already do in helm-charts). We would need to port the functionality from the helm-charts.

Is it ok if we punt on this functionality for now?

jvanz: Yes, that's fine.

updatecli/updatecli.release.d/open-release-pr.yaml
viccuad: Indeed, having both `typefilter.latest` and `typefilter.draft` works!
Because the last release of the repo is both `draft` and `latest` if one edits it right now to see.

I gave it a try to this locally and opened a PR against my fork: https://github.com/viccuad/kubewarden-controller/pull/3/files

viccuad: Fixed with https://github.com/kubewarden/kubewarden-controller/pull/1380/commits/9d3482ed91aa25666ae3fa161be82a191b7b7d3c

Now no more DEBUG complaints.

PR #1399, opened 2026-01-27
chore(automation): move renovate config to monorepo

flavio:
Move the configuration of renovate into the monorepo. While doing that, disable the automatic merge for cargo lock files.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -43,7 +43,7 @@
       "groupName": "GitHub Actions"
     },
     {
-      "_comment": "bundle dependencies updates together",
+      "description": "bundle dependencies updates together",
       "matchUpdateTypes": [
         "patch",
         "minor",


renovate.json
```diff
@@ -1,25 +1,66 @@
...
     },
     {
-      "matchManagers": ["github-actions"],
+      "matchManagers": [
+        "github-actions"
+      ],
       "groupName": "GitHub Actions"
+    },
+    {
+      "_comment": "bundle dependencies updates together",
```
jvanz: This is not valid anymore. You can rename it to "description". I had to do that for the [policies](https://github.com/kubewarden/policies/pull/58) repository as well.

jvanz:
```suggestion
      "description": "bundle dependencies updates together",
```

Issue #1407, opened 2026-01-30
Move the updatecli that lives inside of `kubewarden/charts` that bumps helm charts dependencies into the monorepo


jvanz:
As part of the monorepo consolidation effort (see #1269), the Kubewarden project currently has updatecli configurations in the kubewarden/charts repository that automatically bump Helm chart dependencies. With the transition to a monorepo architecture in kubewarden-controller, these updatecli workflows need to be migrated to prevent conflicts and ensure proper dependency management. The goal is to move these updatecli configurations and GitHub Actions   workflows into the monorepo so that dependency updates are handled within the unified repository structure, maintaining automated bumps for policies, policy-reporter, kuberlr, and other chart dependencies.

 ### Acceptance Criteria

  - Identify and document all updatecli configurations in kubewarden/charts that bump Helm chart dependencies
  - Migrate updatecli manifests from kubewarden/helm-charts to the kubewarden-controller monorepo in the appropriate directory structure
  - Update or create GitHub Actions workflows in the monorepo to run updatecli for dependency bumps
  - Merge the migrated updatecli script with the existing golang version bump script in the repository to consolidate dependency management
  - Verify that dependency bump PRs are automatically created and properly tested in the monorepo
  - Update or deprecate the corresponding workflows in kubewarden/charts to avoid conflicts

PR #1409, opened 2026-01-30
fix: Update Hauler OIDC regex

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/issues/902.

- deps: Bump Hauler tags/OIDC regexps to match new repos. I am bumping tags too so the Hauler manifest is consistent and working.

## Test

Tested it by running `hauler store sync --filename hauler_manifest.yaml` locally.


viccuad:
This makes testing https://github.com/kubewarden/kubewarden-controller/issues/1407 a bit more annoying as one would need to downgrade tags. I am happy rebasing this PR after we merge the automation to bump dependencies.

PR #1410, opened 2026-01-30
chore(deps): migrate updatecli from helm-charts to monorepo

jvanz:
## Description

This change consolidates all dependency management into the kubewarden-controller monorepo by migrating updatecli configurations from the separate kubewarden/helm-charts repository. The migration introduces automated weekly updates   for Go language versions, Helm chart dependencies, policy images, and the Hauler manifest - all unified into a single workflow that runs every Monday at 3:30 AM. This replaces the previous fragmented approach where Go updates ran   hourly and chart dependencies were managed separately in a different repository.

  The consolidation simplifies maintenance by eliminating redundant workflows and compose files, centralizing all dependency update logic in the update-deps.yaml configuration. The updatecli manifests now handle Go version updates in go.mod and Dockerfiles, Helm chart dependencies including policy-reporter and openreports, policy module images in the defaults chart, and automatic synchronization of the Hauler manifest for air-gapped deployments. This unified approach ensures all dependencies are updated together, reducing the risk of version mismatches and making the dependency update process more predictable and maintainable.

Fix https://github.com/kubewarden/kubewarden-controller/issues/1407

Example generated by this script:

https://github.com/jvanz/kubewarden-controller/pull/136
https://github.com/jvanz/kubewarden-controller/pull/135


updatecli/updatecli.d/update-chart-deps.yaml
```diff
@@ -0,0 +1,121 @@
...
+    kind: "github/pullrequest"
+    scmid: "default"
+    spec:
+      automerge: false
+      mergemethod: squash
+      description: |
+        Automatic update of Helm chart dependencies: Helm charts, policies, kuberlr-kubectl image, and Hauler manifest
+        This PR has been created by automation.
+
+        NOTE: REMEMBER TO SQUASH MERGE
```
viccuad: We could add an admontion (like [here](https://github.com/kubewarden/kubewarden-controller/blob/main/updatecli/updatecli.release.d/open-release-pr.yaml#L268)).

updatecli/updatecli.d/update-hauler-manifest.yaml
```diff
@@ -0,0 +1,162 @@
+name: Update Hauler manifest with versions from Helm chart
```
viccuad, line 1: if the 2 PRs are opened at the same time, and the one from `update-chart-deps` has bumps, does that mean that the hauler PR will be out of sync always?

jvanz: The `update-chart-deps.yaml` and `update-hauler-manifest.yaml`  share the same `pipelineid`. It will be only one PR.  You can see this in this [PR](https://github.com/jvanz/kubewarden-controller/pull/136) description. There are section for each of the policies in the compose file. But both are in the same PR

jvanz: The second PR mentioned in the [PR](https://github.com/jvanz/kubewarden-controller/pull/135) description is for golang updates only. It's not related to the Helm and hauler manifest updates.

jvanz:
@viccuad, sorry, I just realized that I forgot to update the commit message considering the latest changes that I did. I just updated the PR and commit descriptions.

updatecli/updatecli.d/update-hauler-manifest.yaml
viccuad: Thanks for the clarification!

PR #1432, opened 2026-02-03
build: Fix darwin x86_64 build

viccuad:
## Description

- build: Target macos 11 to fix build by setting `MACOSX_DEPLOYMENT_TARGET=11.0`
- deps(wasmtime): Disable stack-switching feature in Cargo.toml
- ci: Add workflow_dispatch to manually trigger build-kwct]

## Test

Tested on my fork:
https://github.com/viccuad/kubewarden-controller/actions/runs/21629843944/job/62339408886

### DIFF
--- a/.github/workflows/build-kwctl.yml
+++ b/.github/workflows/build-kwctl.yml
@@ -7,6 +7,13 @@ on:
         required: false
         type: boolean
         default: false
+  workflow_dispatch:
+    inputs:
+      force_build:
+        description: "Force build even if no kwctl-related changes detected"
+        required: false
+        type: boolean
+        default: false
   push:
     branches:
       - "main"


Cargo.toml
```diff
@@ -31,5 +31,30 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["fmt"] }
 url = { version = "2.5", features = ["serde"] }
 wasi-common = "41.0"
-wasmtime = "41.0"
+wasmtime = { version = "41.0", default-features = false, features = [
```
flavio, line 37: I don't like the fact we are disabling stack switching for everything, just to make kwctl happy on osx. Can we reduce the blast radius of this change?

flavio: Another, simpler option, could be to temporarily disable builds for OSX and x86_64. At least until https://github.com/bytecodealliance/wasmtime/issues/12217 is solved for real

flavio: Ok, this is a feature that could be required by some wasm guests. We don't need it right now, I think it's safe to go with this approach, but please leave a comment for that.

viccuad: I tried a more targeted approach, but even if disabling the feature on burrego and policy-evaluator, other dependencies such as wasmtime-provider pull wasmtime with the default features.

Hence I have rebased with the original approach that disables the feature workspace-wise, and added some comments.

PR #1440, opened 2026-02-03
    fix(ci): pin updatecli version.

jvanz:
## Description

The latest updatecli version in use in the CI workflow has an issue that causes problems in the conditions that verifies the chart position in the Hauler manifest. To workaround that, the updatecli version is pinned to a version where the issue does not happen. The updatecli reports: "something wrong happen".  This prevents the script to further update the `hauler_manifest.yaml` file

This is a workaround for our CI for now. But I'll bisect updatecli versions and see what's the root cause later.


viccuad (approved):
Thanks for this! Will have an eye out too.

PR #1455, opened 2026-02-06
ci(open-release-pr): Obtain desired version from workflow dispatch

viccuad:
## Description

Fix  https://github.com/kubewarden/kubewarden-controller/issues/1454
Fix https://github.com/kubewarden/kubewarden-controller/issues/1470

Before: the desired version is read from the GH release draft title,
which doesn't make any sense as the title gets overwritten and it never
matches the RCs we want to do.

Now: The workflow has an input for the version, which is passed as env
var to the updatecli pipeline.

Document current release process in CONTRIBUTING.md

## Test

Tested on my fork:

```
export RELEASE_VERSION=v1.32.1-rc1-viccuad; export UPDATECLI_GITHUB_OWNER=viccuad; clear; updatecli apply --config updatecli/updatecli.release.d/open-release-pr.yaml --values updatecli/values.yaml --debug --clean=true
```

https://github.com/viccuad/kubewarden-controller/pull/6

### DIFF
--- a/updatecli/updatecli.release.d/open-release-pr.yaml
+++ b/updatecli/updatecli.release.d/open-release-pr.yaml
@@ -146,6 +146,15 @@ targets:
     spec:
       file: charts/kubewarden-controller/Chart.yaml
       key: $.appVersion
+  update_kubewarden_controller_auto_install_annotation:
+    name: "Update kubewarden-controller auto-install annotation"
+    kind: yaml
+    sourceid: crdsChartVersion
+    scmid: "default"
+    spec:
+      file: "file://charts/kubewarden-controller/Chart.yaml"
+      key: 'annotations.catalog\.cattle\.io/auto-install'
+      value: 'kubewarden-crds={{ source "crdsChartVersion" }}'
   update_kubewarden_defaults_helm_version:
     scmid: default
     name: Update Helm chart kubewarden-defaults version
@@ -162,6 +171,15 @@ targets:
     spec:
       file: charts/kubewarden-defaults/Chart.yaml
       key: $.appVersion
+  update_kubewarden_defaults_auto_install_annotation:
+    name: "Update kubewarden-defaults auto-install annotation"
+    kind: yaml
+    sourceid: crdsChartVersion
+    scmid: "default"
+    spec:
+      file: "file://charts/kubewarden-defaults/Chart.yaml"
+      key: 'annotations.catalog\.cattle\.io/auto-install'
+      value: 'kubewarden-crds={{ source "crdsChartVersion" }}'
   update_controller_image_tag:
     scmid: default
     name: Update Helm chart kubewarden-controller image tag


.github/workflows/open-release-pr.yml
```diff
@@ -37,6 +42,7 @@ jobs:
         env:
           UPDATECLI_GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}
           UPDATECLI_GITHUB_OWNER: ${{ github.repository_owner }}
+          RELEASE_VERSION: ${{ github.event.inputs.version }}
```
jvanz: Afaik, you can use `${{ inputs.version }}`. It's not necessary the prefix.

updatecli/updatecli.release.d/open-release-pr.yaml
```diff
@@ -1,27 +1,22 @@
...
 # the kubewarden-controller repo.
 
 sources:
-  newestGitHubDraftReleaseVersion:
-    name: Extract version from newest GitHub draft release from release drafter
-    kind: githubrelease
+  desiredReleaseVersion:
+    # TODO(viccuad) in the future, we can drop this source and consume the env var
+    # directly where needed with requiredEnv or a Go template var, for example:
+    # {{ $desiredReleaseVersion := env "RELEASE_VERSION" }}
```
jvanz: Open an issue for that. It looks like a `good first issue` kind of issue. ;)

viccuad: I was trying to be minimal in case of other PRs with changes for the Hauler manifest of the chart versions, but at the end I changed the source name too. I do like though that it's a source, feels more visible than just consuming `sourceId: {{ requiredEnv "RELEASE_VERSION" }}`  on each of the targets.

WDYT? What do you prefer?

jvanz: I'm fine using source. But if this is the case, let's remove the `TODO` comment.

viccuad: Turns out I cannot use the following:

```yaml
  update_kubewarden_crds_helm_appversion:
    scmid: default
    name: Update Helm chart kubewarden-crds appVersion
    kind: yaml
    source: '{{ requiredEnv "RELEASE_VERSION" }}'
    spec:
      file: charts/kubewarden-crds/Chart.yaml
      key: $.appVersion

```
As I get an error where it expects a `sourceid` and not a `source`:
```
ERROR: empty 'sourceid' for target "update_kubewarden_crds_helm_appversion"
```

I have removed the comment then.

jvanz: Ah, I see. But I think you can use `disablesourceinput` and define the `value` field in the spec. Something like this: https://github.com/updatecli/updatecli/blob/467cfeb997027c581aa7354fb0818ad8d2559087/e2e/updatecli.d/success.d/xml.yaml#L83-L91

jvanz:
> Moving to blocked, will wait on #1456 being merged to rebase as needed.

This is now unblocked! :)

jvanz (approved):

updatecli/updatecli.release.d/open-release-pr.yaml
viccuad: That looks indeed cleaner! thanks

viccuad: Argh, so we use the source as we trim the prefix for the `Cargo.*` changes. I could just create 2 env vars, and do the trimming on a `kind: shell`. But it's cleaner as it stands now, so I'm leaving the source.

PR #1482, opened 2026-02-11
ci: add Cargo.toml version check to release workflow

### LINKED ISSUE #1477: Add Cargo.toml version validation for releases
The kwctl standalone repository included a `version-check` CI job that validated the version in `Cargo.toml` matched the git tag during releases, preventing mismatched versions from being published. This check was removed during the monorepo migration and is not currently present in the monorepo CI.

This validation is critical for release quality, ensuring that tagged releases accurately reflect the version declared in the crate manifest. The check should be added to the release workflow to catch version mismatches before artifacts are built and published.

ABHIGYAN-MOHANTA:
Added a version-check job to validate Cargo.toml version against tag version during release.

## Description

Fix #1477

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ permissions: read-all
 
 jobs:
   version-check:
-    name: Check Cargo.toml version
+    name: Check Cargo.toml files version
     runs-on: ubuntu-latest
     steps:
       - name: Download source code


.github/workflows/release.yml
```diff
@@ -8,6 +8,23 @@ on:
...
+  version-check:
+    name: Check Cargo.toml version
+    if: github.ref_type == 'tag'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download source code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Check cargo file version
+        run: |
+          CARGO_VERSION=$(sed -n 's/^version[[:space:]]*=[[:space:]]*"\(.*\)"/\1/p' crates/kwctl/Cargo.toml)
```
viccuad, line 19: Nice to have: check `crates/policy-server/Cargo.toml` too for the same expected TAG_VERSION.
(In that case, please update the name of the job too, at least it should be "files").

.github/workflows/release.yml
```diff
@@ -80,6 +97,7 @@ jobs:
 
   build-kwctl:
     name: Build kwctl binaries
+    needs: version-check
```
viccuad: Instead of making this job depend on `version-check`, I would prefer to make the `ci` job depend on `version-check`. That means we abort earlier in the process.

.github/workflows/release.yml
```diff
@@ -8,6 +8,23 @@ on:
 permissions: read-all
 
 jobs:
+  version-check:
+    name: Check Cargo.toml version
+    if: github.ref_type == 'tag'
```
viccuad: We can safely drop the `if`, the workflow only runs on pushed tags.

Issue #1272, opened 2025-11-25
[policy-server]: Allow defining Sigstore trust root to be used


jvanz:
Recently we change the [kwctl](https://github.com/kubewarden/kwctl/pull/1423) to allow user to define a [ClientTrustConfig](https://github.com/sigstore/protobuf-specs/blob/4d38e4482bf67c7ab86bf2f61e8d79010ac0974e/protos/sigstore_trustroot.proto#L341) json file defining which instance of the Sigstore should be used to verify the policies. We should do that same for policy-server. Therefore, enabling a policy-server to verify policies in a user define Sigstore instance. 

The Sigstore trust root is created in a function which should be updated to check if there is a ClientTrustConfig json file is defined and, if it is, use to build the Sigstore trust configuration returned by the function. 

https://github.com/kubewarden/policy-server/blob/570cc6b3e71f85daf3b51f1a59e078346ac56b5d/src/lib.rs#L322-L331

### Acceptance criteria
- Add a new CLI flag in the policy server to define the path where the `ClientTrustConfig` file is stored
- Update policy server CRD to have a new field to define a config map with the `ClientTrustConfig` json configuration
- Update the controller to mount the config map inside the policy server deployment and add a cli flag in pointing to it
- Change the configuration building to load this new cli flag on it
- Change the function [`create_sigstore_trustroot`](https://github.com/kubewarden/policy-server/blob/570cc6b3e71f85daf3b51f1a59e078346ac56b5d/src/lib.rs#L322) to load the `ClientTrustConfig` when defined in the configuration
- Add e2e tests to the new CLI flag

Issue #1483, opened 2026-02-11
Rename CRD and RBAC files to match kubebuilder/controller-gen defaults


jvanz:
The CRD and RBAC file names in the kubewarden-controller repository currently differ from the default names generated by kubebuilder/controller-gen. This mismatch requires manual intervention after running `make generate-controller`, forcing developers to manually merge the generated files with the custom-named files. This creates friction in the development workflow and introduces potential for errors.

The proposal is to rename these files to match the kubebuilder/controller-gen defaults, enabling automated generation without manual merging. For RBAC files that require custom labels and annotations, the generation process can be adjusted to inject these customizations automatically.

   ### Acceptance Criteria

- Rename CRD files to match kubebuilder/controller-gen default naming convention
- Rename RBAC files (`role.yaml`, `rbac.yaml`) to match kubebuilder defaults
- Update Helm chart references to point to the renamed files
- Adjust code generation to automatically add required labels and annotations to RBAC files
- Verify `make generate-controller` completes without requiring manual file merging
- Update documentation to reflect the new file naming convention
- Check how the CRDs inline html for docs.kubewarden.io. If needed, simplify and drop those inlined html patches.

PR #1485, opened 2026-02-12
feat(policy-server): add support for custom Sigstore trust roots.

jvanz:
## Description
    
Add sigstore-trust-config CLI flag and SigstoreTrustConfig CRD field to enable policy-server to work with private Sigstore instances. Users can now reference a ConfigMap containing custom Sigstore trust root configuration via the PolicyServer resource.

The implementation includes controller logic to mount the ConfigMap, webhook validation to verify the ConfigMap exists, and corresponding updates to the Helm chart. Security documentation has been added to clarify RBAC requirements and ConfigMap namespace restrictions.

Fix https://github.com/kubewarden/kubewarden-controller/issues/1272

During the working on this feature, I've faced some issue on how the manifest are generated and the need of manual work to added them in the kubewarden-crds helm chart. For this reason, I've did some changes on the default version used to generate the files and change where the default location for the files. Therefore, we will not commit file with duplicate CRDs definition and make clear that we need to do some manual work to add new updates in the CRDs. Furthermore, this simplify a little bit on how I can create the ginkgo test to verify this feature. There is already an open issue to address this [issue](https://github.com/kubewarden/kubewarden-controller/issues/1483) in the manifests generation. Thus, I do not see a big problem on doing that for now. 

## Test

I've run the integration test in my fork and force an verification issue: https://github.com/jvanz/kubewarden-controller/actions/runs/21946736441/job/63386244966

### DIFF
--- a/api/policies/v1/policyserver_webhook.go
+++ b/api/policies/v1/policyserver_webhook.go
@@ -147,11 +147,11 @@ func validateImagePullSecret(ctx context.Context, k8sClient client.Client, image
 		Name:      imagePullSecret,
 	}, secret)
 	if err != nil {
-		return fmt.Errorf("cannot get spec.ImagePullSecret: %w", err)
+		return fmt.Errorf("cannot get spec.imagePullSecret: %w", err)
 	}
 
 	if secret.Type != "kubernetes.io/dockerconfigjson" {
-		return fmt.Errorf("spec.ImagePullSecret secret \"%s\" is not of type kubernetes.io/dockerconfigjson", secret.Name)
+		return fmt.Errorf("spec.imagePullSecret secret \"%s\" is not of type kubernetes.io/dockerconfigjson", secret.Name)
 	}
 
 	return nil
@@ -166,11 +166,11 @@ func validateSigstoreTrustConfig(ctx context.Context, k8sClient client.Client, s
 		Name:      sigstoreTrustConfig,
 	}, configMap)
 	if err != nil {
-		return fmt.Errorf("cannot get spec.SigstoreTrustConfig ConfigMap: %w", err)
+		return fmt.Errorf("cannot get spec.sigstoreTrustConfig ConfigMap: %w", err)
 	}
 
 	if _, ok := configMap.Data[constants.PolicyServerSigstoreTrustConfigEntry]; !ok {
-		return fmt.Errorf("spec.SigstoreTrustConfig ConfigMap \"%s\" does not contain required key \"%s\"", sigstoreTrustConfig, constants.PolicyServerSigstoreTrustConfigEntry)
+		return fmt.Errorf("spec.sigstoreTrustConfig ConfigMap \"%s\" does not contain required key \"%s\"", sigstoreTrustConfig, constants.PolicyServerSigstoreTrustConfigEntry)
 	}
 
 	return nil
--- a/api/policies/v1/policyserver_webhook_test.go
+++ b/api/policies/v1/policyserver_webhook_test.go
@@ -237,7 +237,7 @@ func TestPolicyServerValidateSigstoreTrustConfig(t *testing.T) {
 		{
 			name:      "missing ConfigMap",
 			configMap: nil,
-			error:     "cannot get spec.SigstoreTrustConfig ConfigMap",
+			error:     "cannot get spec.sigstoreTrustConfig ConfigMap",
 		},
 		{
 			name: "ConfigMap missing required key",
@@ -263,9 +263,8 @@ func TestPolicyServerValidateSigstoreTrustConfig(t *testing.T) {
 				require.NoError(t, err)
 			}
 
-			configMapName := "sigstore-config"
 			policyServer := NewPolicyServerFactory().
-				WithSigstoreTrustConfigMap(configMapName).
+				WithSigstoreTrustConfigMap("sigstore-config").
 				Build()
 
 			policyServerValidator := policyServerValidator{


charts/kubewarden-crds/templates/admissionpolicies.yaml
```diff
@@ -473,368 +474,3 @@ spec:
     storage: true
     subresources:
       status: {}
-  - additionalPrinterColumns:
```
jvanz, line 476: FYI: https://github.com/kubewarden/kubewarden-controller/pull/1485/changes/bd190922c3e7019a1398e69859e42b717d5ebe6e

Makefile
```diff
@@ -142,7 +142,7 @@ generate-controller: manifests  ## Generate code containing DeepCopy, DeepCopyIn
 
 .PHONY: manifests
 manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects. We use yq to modify the generated files to match our naming and labels conventions.
-	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=controller-role crd webhook paths="./api/policies/v1"  paths="./internal/controller" output:crd:artifacts:config=charts/kubewarden-crds/templates output:rbac:artifacts:config=charts/kubewarden-controller/templates
+	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=controller-role crd webhook paths="./api/policies/v1"  paths="./internal/controller" output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac
```
jvanz, line 149: FYI: https://github.com/kubewarden/kubewarden-controller/pull/1485/changes/bd190922c3e7019a1398e69859e42b717d5ebe6e

internal/controller/suite_test.go
```diff
@@ -129,6 +137,22 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+
+	err = (&policiesv1.ClusterAdmissionPolicy{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = (&policiesv1.AdmissionPolicyGroup{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = (&policiesv1.ClusterAdmissionPolicyGroup{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).ToNot(HaveOccurred())
+
```
jvanz: I've added the webhooks tests here because envtest tests are faster in comparison with the e2e ones that requires a real cluster

api/policies/v1/policyserver_types.go
```diff
@@ -88,12 +88,23 @@ type PolicyServerSpec struct {
...
+	// namespace as the controller deployment), containing Sigstore verification
+	// configuration. The configuration must be under a key named
+	// verification-config in the Configmap.
 	// +optional
 	VerificationConfig string `json:"verificationConfig,omitempty"`
 
+	// Name of SigstoreTrustConfig configmap in the kubewarden namespace (same
+	// namespace as the controller deployment), containing Sigstore trust
+	// configuration (ClientTrustConfig JSON). The configuration must be under a
+	// key named sigstore-trust-config in the Configmap. This is used to configure
```
flavio:
```suggestion
	// key named sigstore-trust-config in the ConfigMap. This is used to configure
```

charts/kubewarden-crds/templates/admissionpolicies.yaml
flavio: thanks for the clarification

charts/kubewarden-defaults/values.yaml
```diff
@@ -121,9 +121,14 @@ policyServer:
...
   serviceAccountName: policy-server
-  # verificationConfig: your_configmap
   # Configmap containing a Sigstore verification configuration under a key
   # named `verification-config`. Must be on the same ns as the PolicyServer.
+  # verificationConfig: your_configmap
+  
+  # Configmap containing a Sigstore trust configuration (ClientTrustConfig JSON)
+  # under a key named `sigstore-trust-config`. Must be on the same ns as the
+  # PolicyServer. This is used to configure a custom Sigstore instance.
+  # sigstoreTrustConfig: your_configmap
```
flavio, line 131: nit:

```suggestion
  # sigstoreTrustConfig: your_configmap

```

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
flavio, line 1: Do we still need these files with the new repo structure? I see sbomscanner doesn't have them.

config/crd/bases/policies.kubewarden.io_policyservers.yaml
flavio, line 1: Do we still need these files with the new repo structure? I see sbomscanner doesn't have them.

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
flavio, line 1: Do we still need these files with the new repo structure? I see sbomscanner doesn't have them.

crates/policy-server/src/cli.rs
```diff
@@ -124,6 +124,13 @@ pub(crate) fn build_cli() -> Command {
             .env("KUBEWARDEN_SIGSTORE_CACHE_DIR")
             .help("Directory used to cache sigstore data"),
 
+        Arg::new("sigstore-trust-config")
+            .long("sigstore-trust-config")
+            .value_name("SIGSTORE_TRUST_CONFIG_PATH")
+            .env("KUBEWARDEN_SIGSTORE_TRUST_CONFIG_PATH")
+            .value_parser(clap::builder::PathBufValueParser::new())
+            .help("JSON-formatted file conforming to the ClientTrustConfig message in the Sigstore protobuf specs. This file configures the entire Sigstore instance state, including the URIs used to access the CA and artifact transparency services as well as the cryptographic root of trust itself"),
```
flavio: The help message is really mouthful:

```suggestion
            .help("Path to a JSON Sigstore ClientTrustConfig file"),
```

crates/policy-server/src/lib.rs
```diff
@@ -329,3 +336,49 @@ async fn create_sigstore_trustroot(config: &Config) -> Result<Arc<SigstoreTrustR
...
+    let trust_root = client_trust_config.trusted_root.ok_or_else(|| {
+        anyhow!(
+            "Sigstore trust config file {} missing trusted_root field",
+            pki_file.display()
+        )
+    })?;
+
+    let trust_root_bytes = serde_json::to_vec(&trust_root)?;
+    // Safety: The trust root data comes from a user-provided ConfigMap.
+    // RBAC should be used to restrict access to this ConfigMap.
```
flavio: nit, I think we stated that already too many times :) 
```suggestion
````

File: crates/policy-server/tests/sigstore.rs [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+use policy_evaluator::{
+    admission_response_handler::policy_mode::PolicyMode,
+    policy_fetcher::{sources::read_sources_file, verify::config::VerificationConfigV1},
+};
+#[cfg(feature = "sigstore-testing")]
+use policy_server::config::PolicyOrPolicyGroup;
+#[cfg(feature = "sigstore-testing")]
+use std::collections::BTreeSet;
+#[cfg(feature = "sigstore-testing")]
+use tower::ServiceExt;
```
flavio: there's no need to put all the `use` statements behind the `cfg` check. We can always import them, this is not going to be a big deal.

Also, reorganize the imports to:

- put `std` library imports first
- leave an empty line
- put external imports
- leave an empty line
- put project imports (including `mod common`)
- group imports `std::{collections::{BTreeSet, HashMap},...}`

crates/policy-server/tests/sigstore.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// This test is behind a feature flag because it requires a sigstore testing environment to run
+/// properly. In CI this is done by the sigstore/scaffolding/actions/setup action and some
+/// additional configuration. Locally, one can use the script and documentation available in the
+/// same repository.
```
flavio: It could be nice to point to the path of the documentation mentioned.

crates/policy-server/tests/sigstore.rs
```diff
@@ -0,0 +1,132 @@
...
+/// information to find the local sigstore instance. It follows the ClientTrustConfig format. See the
+/// spec here:
+/// https://github.com/sigstore/protobuf-specs/blob/4d38e4482bf67c7ab86bf2f61e8d79010ac0974e/protos/sigstore_trustroot.proto#L341
+/// The verification_config.yaml file should contain the verification configuration for the policy,
+/// it can be generated using `kwctl scaffold verification-config` command. The verification
+/// options in the file should match the way the policy was signed in the local sigstore instance.
+/// The test also checks if there is a sources.yaml file. If it exists, it is used by the policy
+/// server to handle insecure registries.
+#[tokio::test]
+#[cfg(feature = "sigstore-testing")]
```
flavio: this is the only `cfg` guard that we should leave around.

crates/policy-server/tests/sigstore.rs
```diff
@@ -0,0 +1,132 @@
...
+        "trust_config.json not found at workspace root"
+    );
+
+    let verification_config_path = workspace_root.join("verification_config.yaml");
+    assert!(
+        verification_config_path.exists(),
+        "verification_config.yaml not found at workspace root"
+    );
+
+    // if there is a sources.yaml file, the test consider that the registry does not support https
```
flavio: Revisit the inline explanation, there's no "if there's a source...". The file is expected to be there, otherwise the assertion on the next line will fail.

crates/policy-server/tests/sigstore.rs
```diff
@@ -0,0 +1,132 @@
...
+    // request. Therefore, the sources file should be copied to the tempdir as well and the command
+    // should be run with the --sources-path argument.
+    let source_sources_yaml = workspace_root.join("sources.yaml");
+    assert!(
+        source_sources_yaml.exists(),
+        "sources.yaml not found at workspace root"
+    );
+
+    // Load verification config
+    let verification_config_content = std::fs::read_to_string(&verification_config_path)
```
flavio: nit:

```rust
use std::fs;
//...

fs::read_to_string()
```

internal/controller/policyserver_controller_test.go
```diff
@@ -1286,4 +1287,234 @@ var _ = Describe("PolicyServer controller", func() {
...
+
+			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).Build()
+			policyServer.Spec.SigstoreTrustConfig = configMapName
+
+			err := k8sClient.Create(ctx, policyServer)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("does not contain required key"))
+		})
+
+		It("should accept PolicyServer update changing SigstoreTrustConfig", func() {
```
flavio: This test is making sure the validating webhook accepts the UPDATE request, as long as the new configmap satisfies the requirements.

It doesn't however make sure the actual `Deployment` is consuming the new value.

internal/controller/policyserver_controller_test.go
```diff
@@ -1286,4 +1287,234 @@ var _ = Describe("PolicyServer controller", func() {
 			})
 		})
 	})
+
+	When("configuring SigstoreTrustConfig", func() {
```
flavio: Each test creates a PolicyServer and waits for the resource to be reconciled. This is slowing down everything.

Take for example the test that makes sure the ConfigMap is mounted as a volume and the other test that makes sure the env variable is set accordingly.

We could refactor the test to something like:

- Have a variable shared by the different tests that holds the name of the PolicyServer being created
- Have the `it` blocks focus only on doing assertions

internal/controller/policyserver_controller_test.go
```diff
@@ -1286,4 +1287,234 @@ var _ = Describe("PolicyServer controller", func() {
...
+		It("should create PolicyServer with valid SigstoreTrustConfig ConfigMap", func() {
```
flavio: I have a foggy memory about the project tests... This looks like a test for the validating webhook. Can't we move that over there? I hope the webhook tests are faster than these ones.

UPDATE: see my comment below

internal/controller/policyserver_controller_test.go
```diff
@@ -1286,4 +1287,234 @@ var _ = Describe("PolicyServer controller", func() {
...
+			configMapName := newName("sigstore-config")
+			createConfigMapWithSigstoreTrustConfig(ctx, configMapName)
+
+			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).WithSigstoreTrustConfigMap(configMapName).Build()
+
+			createPolicyServerAndWaitForItsService(ctx, policyServer)
+		})
+
+		It("should reject PolicyServer with missing SigstoreTrustConfig ConfigMap", func() {
```
flavio: I have a foggy memory about the project tests... This looks like a test for the validating webhook. Can't we move that over there? I hope the webhook tests are faster than these ones.

UPDATE: see my comment below

internal/controller/policyserver_controller_test.go
```diff
@@ -1286,4 +1287,234 @@ var _ = Describe("PolicyServer controller", func() {
...
+			nonExistentConfigMap := "nonexistent-sigstore-config"
+
+			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).WithSigstoreTrustConfigMap(nonExistentConfigMap).Build()
+
+			err := k8sClient.Create(ctx, policyServer)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("cannot get spec.SigstoreTrustConfig ConfigMap"))
+		})
+
+		It("should reject PolicyServer with SigstoreTrustConfig ConfigMap missing required key", func() {
```
flavio: I have a foggy memory about the project tests... This looks like a test for the validating webhook. Can't we move that over there? I hope the webhook tests are faster than these ones.


UPDATE: see my comment below

internal/controller/policyserver_controller_test.go
flavio: If we do have webhook tests somewhere else, I would move the test that ensures we can update the value over there. Here I would leave the code that tests that a new Deployment is started, using the new configmap.

The test logic could also strip the Sigstore PKI custom config from the PolicyServer definition, and just make sure the new Deployment is missing all the data associated with that (volumes, env vars)

config/rbac/role.yaml
flavio, line 1: Do we still need this file?

internal/controller/suite_test.go
flavio: I see, a better approach would be test them with regular unit tests. SBOMscanner does that (see [here](https://github.com/kubewarden/sbomscanner/tree/d4c8060767c08c0c2e00cf0ac9877da2c83332ee/internal/webhook/v1alpha1)).

Makefile
flavio: I've just seen:

> Modified Makefile to output generated files to config/crd/bases and config/rbac directories instead of Helm chart templates, enabling their use in controller tests with envtest. This change should be revisited once a decision is made about whether to continue using the config directory structure or remove it in favor of generating directly into chart templates.

I think we should get rid of the `config` directory. I like how SBOMscanner doesn't make use of it, yet it allows the usage of CRDs and such inside of the env tests.

Let's see what is being done there.

I like how the previous `Makefile` didn't require any extra step to keep in sync the CRDs inside of our chart without the need of doing any extra step.

internal/controller/suite_test.go
flavio: Wait, I've seen  we also have unit tests for the webhooks under `api/policies/v1/policyserver_webhook_test.go`.

Why do we need all of that? Can't we extend these tests to cover the new validation that has been introduced?

api/policies/v1/policyserver_types.go
flavio: BTW, can you also update the CRD docs as part of this PR?

flavio (changes requested):
Overall LGTM. I left some comments.

I gotta say, I love the monorepo 😄

Makefile
viccuad: I agree with dropping `config` (and saving `config/samples/` under `examples/` for now). I agree too on not having an additional step to keep the CRDs in sync.

crates/policy-server/tests/sigstore.rs
jvanz: We need the `cfg` otherwise the compilation will fail because the "imports are not in use". I have another solution that I'll push soon.

internal/controller/suite_test.go
jvanz: The sigstore configuration validation requires to fetch data from the cluster. It checks if the configmap exists and if the required key is there. This is not possible in the current state of the unit tests. For that, we need to add a mocking infrastructure.

jvanz: What do you prefer? Add the mock of the k8s client or use the envtest in place?

Makefile
jvanz: > During the working on this feature, I've faced some issue on how the manifest are generated and the need of manual work to added them in the kubewarden-crds helm chart. For this reason, I've did some changes on the default version used to generate the files and change where the default location for the files. Therefore, we will not commit file with duplicate CRDs definition and make clear that we need to do some manual work to add new updates in the CRDs. Furthermore, this simplify a little bit on how I can create the ginkgo test to verify this feature. There is already an open issue to address this https://github.com/kubewarden/kubewarden-controller/issues/1483 in the manifests generation. Thus, I do not see a big problem on doing that for now.

As I said in the description, it's not just a matter of generating in a different location. We need to adapt the files manually or using a script. As far as I can see, that's why kube-builder use kustomize. It allow this kind of manipulation. 

For example, the rbac/role one need an update to add some defaults [labels](https://github.com/kubewarden/kubewarden-controller/blob/b2c7281197788ed213029750bdd6dc3a9e074dff/charts/kubewarden-controller/templates/rbac.yaml#L7). Because of that, I've opened an issue to address that. Here, I'm just updating the required files to make the feature work. And leave the final solution for the manifest generation for a future PR.

config/crd/bases/policies.kubewarden.io_admissionpolicies.yaml
jvanz: Let's keep the discussion about these file in a single thread: https://github.com/kubewarden/kubewarden-controller/pull/1485#discussion_r2801473177

config/crd/bases/policies.kubewarden.io_policyservers.yaml
jvanz: Let's keep the discussion about these file in a single thread: https://github.com/kubewarden/kubewarden-controller/pull/1485#discussion_r2801473177

config/crd/bases/policies.kubewarden.io_clusteradmissionpolicies.yaml
jvanz: Let's keep the discussion about these file in a single thread: https://github.com/kubewarden/kubewarden-controller/pull/1485#discussion_r2801473177

config/rbac/role.yaml
jvanz: Let's keep the discussion about these file in a single thread: https://github.com/kubewarden/kubewarden-controller/pull/1485#discussion_r2801473177

api/policies/v1/policyserver_webhook.go
```diff
@@ -151,6 +157,25 @@ func validateImagePullSecret(ctx context.Context, k8sClient client.Client, image
...
+// validateSigstoreTrustConfig validates that the specified PolicyServer sigstoreTrustConfig ConfigMap exists
+// and contains the required key.
+func validateSigstoreTrustConfig(ctx context.Context, k8sClient client.Client, sigstoreTrustConfig string, deploymentsNamespace string) error {
+	configMap := &corev1.ConfigMap{}
+	err := k8sClient.Get(ctx, client.ObjectKey{
+		Namespace: deploymentsNamespace,
+		Name:      sigstoreTrustConfig,
+	}, configMap)
+	if err != nil {
+		return fmt.Errorf("cannot get spec.SigstoreTrustConfig ConfigMap: %w", err)
```
viccuad:
```suggestion
		return fmt.Errorf("cannot get spec.sigstoreTrustConfig ConfigMap: %w", err)
```

These are user-facing errors, that should match the capitalization of the spec fields. FWIW, other errors such as in `validateImagePullSecret()` have the same typo.

api/policies/v1/policyserver_webhook.go
```diff
@@ -151,6 +157,25 @@ func validateImagePullSecret(ctx context.Context, k8sClient client.Client, image
...
+	}
+
+	if _, ok := configMap.Data[constants.PolicyServerSigstoreTrustConfigEntry]; !ok {
+		return fmt.Errorf("spec.SigstoreTrustConfig ConfigMap \"%s\" does not contain required key \"%s\"", sigstoreTrustConfig, constants.PolicyServerSigstoreTrustConfigEntry)
```
viccuad:
```suggestion
		return fmt.Errorf("spec.sigstoreTrustConfig ConfigMap \"%s\" does not contain required key \"%s\"", sigstoreTrustConfig, constants.PolicyServerSigstoreTrustConfigEntry)
```

internal/controller/suite_test.go
jvanz: Never mind, I'm moving the webhook validation tests to unit test.

Makefile
jvanz: As we discussed in our daily call, we will fix the CRDs in a future PR.

internal/controller/policyserver_controller_test.go
jvanz: I've changed the test organization to follow pattern from the file where we have creation/updating/delete contexts. It's not exactly what you asked here. But it also reduce the amount of policy server creation.

flavio (approved):
LGTM, thanks for the changes

crates/policy-server/Cargo.toml
```diff
@@ -65,6 +66,7 @@ inotify      = "0.11"
 tokio-stream = "0.1.18"
 
 [dev-dependencies]
+aws-lc-rs      = "1.15"
```
flavio: why have you added this as a dev dependency? This should be automatically pulled in by higher order dependencies (like rustls)

crates/policy-server/Cargo.toml
```diff
@@ -82,4 +84,5 @@ reqwest = { version = "0.13", default-features = false, features = [
 ] }
 
 [features]
-otel_tests = []
+otel_tests       = []
+sigstore-testing = ["rustls/aws-lc-rs"]
```
flavio: I'm also confused about this one, we should always end up using aws-lc-rs as crypto backend of rustls. AFAIK this is the [default upstream](https://github.com/rustls/rustls/blob/b47bf5466a282cd63db8fc73a533ea05c1a90d5e/rustls/Cargo.toml#L18-L20)

crates/policy-server/tests/sigstore.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// options in the file should match the way the policy was signed in the local sigstore instance.
+    /// The test also checks if there is a sources.yaml file. If it exists, it is used by the policy
+    /// server to handle insecure registries.
+    #[tokio::test]
+    async fn test_sigstore_trust_config() {
+        // Initialize rustls crypto provider before any rustls operations
+        // This is needed because axum-server uses tls-rustls-no-provider
+        let _ = rustls::crypto::CryptoProvider::install_default(
+            rustls::crypto::aws_lc_rs::default_provider(),
+        );
```
flavio: This shouldn't be needed anymore since [this PR got merged](https://github.com/kubewarden/kubewarden-controller/pull/1498).

Maybe this is a leftover and is also what is causing you to have to add the dependencies I've outlined inside of the `Cargo.toml` file.

jvanz: No, it's not a leftover. This is required to make it works. The cargo changes that you've outlined are leftover of previous try to fix the test. Those changes are the first try to make the test works. Because the test was complaining about the missing provider that can be defined with a feature or calling this function. However, forcing the feature only did not work. Therefore, I've added the code here as well. Because of that, I've removed the changes from the Cargo.toml file in the last update.

I think that the code in the policy server is not failing because some of our dependencies are initializing the provider before the application doing the TLS operations. And, as in the test the code path is a little different, we hit this issue. Now I'm thinking more about that. Maybe it's a good idea to add a similar code in the main of the policy server. Just to ensure that the crypto provider is actually what we want and do not rely on the dependency chain resolution or initialization from some of our dependencies. We already do that in [kwctl](https://github.com/kubewarden/kubewarden-controller/blob/476f8232e5c8cf5081e54b18318590b186003f56/crates/kwctl/src/main.rs#L113).

flavio: Thanks for the explanation. rust crypto can be pretty messy :disappointed:

Issue #1300, opened 2025-11-06
[kwctl][policy-server]: handle HTTP proxy


davejab:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When attempting to pull existing policy from ECR via proxy DNS lookup appears to fail:

```
$ kwctl -v pull registry://111111111111.dkr.ecr.eu-west-1.amazonaws.com/mypolicy:v0.0.1

2025-11-06T11:21:14.664644Z DEBUG policy_fetcher: pulling policy url=Url { scheme: "registry", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("111111111111.dkr.ecr.eu-west-1.amazonaws.com")), port: None, path: "/mypolicy:v0.0.1", query: None, fragment: None }
2025-11-06T11:21:14.664781Z DEBUG policy_fetcher::registry: fetching policy image=Reference { registry: "111111111111.dkr.ecr.eu-west-1.amazonaws.com", mirror_registry: None, repository: "mypolicy", tag: Some("v0.0.1"), digest: None } client_protocol=Https(SystemCa)
2025-11-06T11:21:14.670087Z DEBUG oci_client::client: Pulling image: Reference { registry: "111111111111.dkr.ecr.eu-west-1.amazonaws.com", mirror_registry: None, repository: "mypolicy", tag: Some("v0.0.1"), digest: None }
2025-11-06T11:21:14.670144Z DEBUG oci_client::client: Pulling image manifest from https://111111111111.dkr.ecr.eu-west-1.amazonaws.com/v2/mypolicy/manifests/v0.0.1
2025-11-06T11:21:14.670179Z DEBUG oci_client::token_cache: Fetching token key.registry=111111111111.dkr.ecr.eu-west-1.amazonaws.com key.repository=mypolicy key.operation=Pull miss=true
2025-11-06T11:21:14.670196Z DEBUG oci_client::client: Authorizing for image: Reference { registry: "111111111111.dkr.ecr.eu-west-1.amazonaws.com", mirror_registry: None, repository: "mypolicy", tag: Some("v0.0.1"), digest: None }
2025-11-06T11:21:14.670208Z DEBUG oci_client::client: url="https://111111111111.dkr.ecr.eu-west-1.amazonaws.com/v2/"
2025-11-06T11:21:14.670251Z DEBUG reqwest::connect: starting new connection: https://111111111111.dkr.ecr.eu-west-1.amazonaws.com/
[...]

PR #1487, opened 2026-02-13
feat(kwctl,policy-server): Honour proxy env vars

viccuad:
## Description

Part of  https://github.com/kubewarden/kubewarden-controller/issues/1300

Honour HTTP_PROXY,HTTPS_PROXY,NO_PROXY env vars and their lowercase counterparts.
Implement proxy features for policy-fetcher related features, such as pull, push, pull-and-run.

Support for proxy env vars for context-aware policy calls will come later.

Document on readmes and on `kwctl` CLI commands.

## Test

Add lowercase unit tests.
Add kwctl/policy-server integration tests.

### DIFF
--- a/crates/policy-fetcher/Cargo.toml
+++ b/crates/policy-fetcher/Cargo.toml
@@ -51,6 +51,7 @@ x509-parser = "0.18"
 anyhow = { workspace = true }
 rcgen = "0.14"
 rstest = { workspace = true }
+temp-env = "0.3"
 tempfile = { workspace = true }
 testcontainers = { version = "0.27", default-features = false, features = [
   "aws-lc-rs",
--- a/crates/policy-fetcher/src/lib.rs
+++ b/crates/policy-fetcher/src/lib.rs
@@ -570,4 +570,34 @@ mod tests {
             !success
         );
     }
+
+    #[test]
+    fn test_get_proxy_env_vars_case_insensitivity() {
+        temp_env::with_vars(
+            [
+                ("http_proxy", Some("http://lowercase-http")),
+                ("https_proxy", Some("http://lowercase-https")),
+                ("no_proxy", Some("localhost,127.0.0.1")),
+            ],
+            || {
+                let (http, https, no) = get_proxy_env_vars();
+                assert_eq!(http, Some("http://lowercase-http".to_string()));
+                assert_eq!(https, Some("http://lowercase-https".to_string()));
+                assert_eq!(no, Some("localhost,127.0.0.1".to_string()));
+            },
+        );
+        temp_env::with_vars(
+            [
+                ("HTTP_PROXY", Some("http://uppercase-http")),
+                ("HTTPS_PROXY", Some("http://uppercase-https")),
+                ("NO_PROXY", Some("our.example")),
+            ],
+            || {
+                let (http, https, no) = get_proxy_env_vars();
+                assert_eq!(http, Some("http://uppercase-http".to_string()));
+                assert_eq!(https, Some("http://uppercase-https".to_string()));
+                assert_eq!(no, Some("our.example".to_string()));
+            },
+        );
+    }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4624,6 +4624,7 @@ dependencies = [
  "serde_yaml",
  "sha2",
  "sigstore",
+ "temp-env",
  "tempfile",
  "testcontainers",
  "textwrap",


crates/kwctl/src/cli.rs
```diff
@@ -121,6 +121,15 @@ fn subcommand_pull() -> Command {
...
     Command::new("pull")
         .about("Pulls a Kubewarden policy from a given URI")
+        .long_about(
+            r#"Pulls a Kubewarden policy from a given URI
+
+It respects standard proxy environment variables when downloading policies:
+- HTTP_PROXY or http_proxy: proxy server for HTTP requests
+- HTTPS_PROXY or https_proxy: proxy server for HTTPS requests
+- NO_PROXY or no_proxy: comma-separated list of hosts to exclude from proxying
+"#,
```
jvanz: Shouldn't we add this documentation in the root command? If I execute the `run` command, which will pull the policy if it is not present, it will use this env var as well, right?

viccuad: Indeed! rebased and fixed.

crates/kwctl/src/cli.rs
```diff
@@ -188,6 +201,15 @@ fn subcommand_verify() -> Command {
...
     Command::new("verify")
         .about("Verify a Kubewarden policy from a given URI using Sigstore")
+        .long_about(
+            r#"Verify a Kubewarden policy from a given URI using Sigstore
+
+It respects standard proxy environment variables when downloading policies:
+- HTTP_PROXY or http_proxy: proxy server for HTTP requests
+- HTTPS_PROXY or https_proxy: proxy server for HTTPS requests
+- NO_PROXY or no_proxy: comma-separated list of hosts to exclude from proxying
+"#,
```
flavio: needs to be replaced with the static string.

Also, not related, but also `push` should handle these variables.

crates/kwctl/tests/proxy.rs
```diff
@@ -0,0 +1,49 @@
...
+};
+
+mod common;
+
+const POLICY_URI: &str = "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5";
+
+#[rstest]
+#[case::uppercase("HTTP_PROXY", "HTTPS_PROXY")]
+#[case::https_proxy_only("", "HTTPS_PROXY")]
+#[case::lowercase("http_proxy", "https_proxy")]
```
flavio: I think that testing the handling of uppercase/lowercase handling should be done inside of a cheap unit test.

Here we should focus on testing the http and https handling.

It would be nice to ensure the policy was actually pulled through the proxy. Can we for example look for some logs inside of the proxy?

flavio: Also, it would be nice to test the NO_PROXY setting.

crates/policy-fetcher/src/lib.rs
```diff
@@ -220,6 +220,22 @@ pub(crate) fn host_and_port(url: &Url) -> std::result::Result<String, errors::In
     ))
 }
 
+/// Helper function, reads proxy configuration from environment variables.
+/// Checks both uppercase and lowercase variants (HTTP_PROXY/http_proxy, etc.)
+pub(crate) fn get_proxy_env_vars() -> (Option<String>, Option<String>, Option<String>) {
```
flavio, line 225: This is the place to write the unit tests that verifies upper/lower case management.

crates/policy-server/tests/integration_test.rs
```diff
@@ -1300,3 +1301,69 @@ fn build_request_client(
     }
     builder.build().expect("failed to build client")
 }
+
+#[rstest]
+#[case::uppercase("HTTP_PROXY", "HTTPS_PROXY")]
+#[case::https_proxy_only("", "HTTPS_PROXY")]
+#[case::lowercase("http_proxy", "https_proxy")]
```
flavio: Same here

crates/kwctl/tests/proxy.rs
flavio: BTW, you're never going to test the HTTP proxy with our official images, since they are all hosted over HTTPS.

You could try to test HTTP proxy by using a local registry or by hosting the policy with a traditional web server.

crates/kwctl/src/cli.rs
viccuad: Done

crates/kwctl/tests/proxy.rs
viccuad: Indeed. Moved lowercase tests to policy-fetcher unit tests, added correct integration tests with a proxy and a local OCI registry for the rest.

crates/policy-fetcher/Cargo.toml
```diff
@@ -8,7 +8,7 @@ authors = [
 ]
 edition = "2024"
 name = "policy-fetcher"
-version = "0.11.0"
+version = "0.12.0"
```
jvanz, line 11: Now that we have all the crates in the same repository, I'm wondering if we shouldn't update this version together with the other crates in the updatecli script. This is blocker and I don't want to overthink this... Just sharing an idea that came to mind while reviewing this PR.

jvanz (approved):

PR #1497, opened 2026-02-16
ci: Add coverage for rust code

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/1347

Add new `make coverage-rust` that runs all Rust tests and saves the coverage locally `under coverage/rust/*` in html format.

Add new `coverage-rust` job that runs all Rust tests in CI and submits the result to Codecov.

## Test

Tested locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,6 +290,29 @@ jobs:
         run: sudo apt install -y bats
       - name: Run integration tests
         run: make -C crates/policy-evaluator integration-tests
+        
+  coverage-rust:
+    name: coverage-rust
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@0e76c5c569f13f7eb21e8e5b26fe710062b57b62 # v2.65.13
+        with:
+          tool: cargo-llvm-cov
+      - name: Install cosign # this is needed by some of the e2e tests
+        uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
+      - run: cargo llvm-cov --ignore-run-fail --doctests --lcov --output-path coverage/rust/lcov.info
+      - name: Upload Rust test coverage to Codecov
+        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_ORG_TOKEN }}
+        with:
+          name: rust-tests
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -290,6 +290,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@0e76c5c569f13f7eb21e8e5b26fe710062b57b62 # v2.65.13
+        with:
+          tool: cargo-llvm-cov
+      - name: Install cosign # this is needed by some of the e2e tests
+        uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
+      - run: cargo llvm-cov --ignore-run-fail --doctests --lcov --output-path coverage/rust/lcov.info
+      - name: Upload Rust test coverage to Codecov
```
viccuad, line 307: As it stands, the `cargo llvm-cov` is configured to not fail if there's tests that have failed, and we always submit the coverage data.

This is contrary to what we do for go, as there the push of the data happens after all tests have been successful.

It would be possible to change this and only submit on full rust test runs that are successful, or run the coverage tests in parallel, and only submit the coverage data if the rust test are successful. I'm not sure I much value in that.

.github/workflows/ci.yml
```diff
@@ -290,6 +290,29 @@ jobs:
...
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@0e76c5c569f13f7eb21e8e5b26fe710062b57b62 # v2.65.13
+        with:
+          tool: cargo-llvm-cov
+      - name: Install cosign # this is needed by some of the e2e tests
+        uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
+      - run: cargo llvm-cov --ignore-run-fail --doctests --lcov --output-path coverage/rust/lcov.info
```
jvanz, line 306: Why not calling the Makefile target?

.github/workflows/ci.yml
jvanz: I'm fine with the current approach.

flavio: I'm fine with the current approach. I also wonder if we shouldn't just the rust test suite with coverage turned on. Basically, avoid to introduce a new job for "rust with coverage".

.github/workflows/ci.yml
viccuad: Summarizing off-band discussion:

The CI invocation and local invocation differ, CI produces lcov.info and local html for human consumption. Tried producing both on the same run but it's not possible. Hence, leaving the CI invocation as is.

PR #1498, opened 2026-02-16
fix: correct PEM header for ECDSA private keys

flavio:
The certificate generation code was creating ECDSA private keys but incorrectly labeling them as 'RSA PRIVATE KEY' in the PEM header.

This caused the PolicyServer to fail parsing keys after switching from rustls-pemfile to rustls_pki_types. The latter crate is more strict when parsing PEM contents.


flavio:
There was a second issue, I'm about to push the fixes.

Issue #1499, opened 2026-02-18
Upgrade of kubewarden fail from 1.32.1 -> Latest


kravciak:
After I upgrade kubewarden I get error message in policy-server logs:

```bash
# Install current kubewarden
~ make clean install VERSION=1.32.1

# Upgrade to kubewarden with latest images
~ make upgrade LATEST=1
```

```bash
~ k get pods -n kubewarden
NAME                                     READY   STATUS             RESTARTS      AGE
kubewarden-controller-6cc957dd7b-xr9jp   1/1     Running            0             3m8s
policy-server-default-596584d5b-f46kj    1/1     Running            0             4m25s
policy-server-default-76bd654948-fvdq6   0/1     CrashLoopBackOff   4 (59s ago)   2m37s

~ k logs -n kubewarden policy-server-default-76bd654948-fvdq6
2026-02-18T14:34:27.980112Z  INFO sigstore::cosign::client_builder: Cannot parse Rekor public key with id cf1199155bddd051268d1f16ac5c0c75c009f6fb5a63f4177f8e18d7051e3fa0: Pkcs8 spki error : Ecdsa-P256 from der bytes to public key failed: unknown/unsupported algorithm OID: 1.2.840.10045.2.1
2026-02-18T14:34:27.980203Z  INFO policy_server::policy_downloader: policies download download_dir="/tmp" policies_count=0 status="init"
2026-02-18T14:34:27.984266Z  INFO policy_server: policy timeout protection is enabled execution_limit_seconds=2
Error: unexpected error: failed to parse private key as RSA, ECDSA, or EdDSA
```

Related to https://github.com/kubewarden/kubewarden-controller/pull/1498

jvanz:
Nice catch indeed! I'm also fine with option 2 and the type change. Go for it!

PR #1501, opened 2026-02-19
feat: Version the format of policy-server cert Secrets via annot

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/1499

- Add new annotation `kubewarden/cert-format-version`, set to a numeric value that informs of our desired Secret format version.

  On reconciliation, if the new annotation is missing or its version is
not matching, reissue the certificate and recreate the Secret.

  This allows us to perform migrations of the policy-server cert Secret
format, including the cert and key inside. For example, this ensures that the migration to the new tls.key header
string is performed seamlessly.

- Set policy-server cert secret type to TLS. This doesn't have any functional change, our Secret already was
compliant with this type. This will get reconciled on upgrade to the new cert-format-version 1.

## Test

Added new unit tests to reconcile when annotation `kubewarden/cert-format-version`.

### DIFF
--- a/internal/controller/policyserver_controller_cert_secret_test.go
+++ b/internal/controller/policyserver_controller_cert_secret_test.go
@@ -1,0 +1,102 @@
+package controller
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	policiesv1 "github.com/kubewarden/kubewarden-controller/api/policies/v1"
+	"github.com/kubewarden/kubewarden-controller/internal/constants"
+)
+
+var _ = Describe("PolicyServer certificate secret reconciliation", func() {
+	ctx := context.Background()
+	var policyServerName string
+
+	BeforeEach(func() {
+		policyServerName = newName("policy-server-testcert")
+	})
+
+	When("a PolicyServer is created", func() {
+		It("should create the cert secret with certificate data, format annotation, and correct type and labels", func() {
+			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).Build()
+			createPolicyServerAndWaitForItsService(ctx, policyServer)
+
+			Eventually(func() error {
+				secret, err := getTestPolicyServerSecret(ctx, policyServerName)
+				if err != nil {
+					return err
+				}
+
+				By("verifying the secret type")
+				Expect(secret.Type).To(Equal(corev1.SecretTypeOpaque))
+
+				By("verifying the certificate data was generated")
+				Expect(secret.Data).To(HaveKey(constants.ServerCert))
+				Expect(secret.Data[constants.ServerCert]).ToNot(BeEmpty())
+				Expect(secret.Data).To(HaveKey(constants.ServerPrivateKey))
+				Expect(secret.Data[constants.ServerPrivateKey]).ToNot(BeEmpty())
+
+				By("verifying the format version annotation was added")
+				Expect(secret.Annotations).To(HaveKeyWithValue(
+					constants.ServerCertSecretFormatAnnotation,
+					constants.ServerCertSecretFormatVersion,
+				))
+
+				By("verifying the labels were set")
+				Expect(secret.Labels).To(HaveKeyWithValue(constants.PartOfLabelKey, constants.PartOfLabelValue))
+				Expect(secret.Labels).To(HaveKeyWithValue(constants.ComponentLabelKey, constants.ComponentPolicyServerLabelValue))
+
+				return nil
+			}, timeout, pollInterval).Should(Succeed())
+		})
+	})
+
+	When("a cert secret without the format annotation already exists", func() {
+		It("should regenerate the certificate and add the format annotation", func() {
+			By("pre-creating the cert secret without the format annotation")
+			preExistingSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      getPolicyServerNameWithPrefix(policyServerName),
+					Namespace: deploymentsNamespace,
+				},
+				Type: corev1.SecretTypeOpaque,
+				Data: map[string][]byte{
+					constants.ServerCert:       []byte("old-cert-data"),
+					constants.ServerPrivateKey: []byte("old-key-data"),
+				},
+			}
+			Expect(k8sClient.Create(ctx, preExistingSecret)).To(Succeed())
+
+			policyServer := policiesv1.NewPolicyServerFactory().WithName(policyServerName).Build()
+			createPolicyServerAndWaitForItsService(ctx, policyServer)
+
+			Eventually(func() error {
+				secret, err := getTestPolicyServerSecret(ctx, policyServerName)
+				if err != nil {
+					return err
+				}
+
+				By("verifying the format version annotation was added")
+				Expect(secret.Annotations).To(HaveKeyWithValue(
+					constants.ServerCertSecretFormatAnnotation,
+					constants.ServerCertSecretFormatVersion,
+				))
+
+				By("verifying the certificate was regenerated")
+				Expect(secret.Data[constants.ServerCert]).ToNot(Equal([]byte("old-cert-data")))
+
+				By("verifying the secret type and labels")
+				Expect(secret.Type).To(Equal(corev1.SecretTypeOpaque))
+				Expect(secret.Labels).To(HaveKeyWithValue(constants.PartOfLabelKey, constants.PartOfLabelValue))
+				Expect(secret.Labels).To(HaveKeyWithValue(constants.ComponentLabelKey, constants.ComponentPolicyServerLabelValue))
+
+				return nil
+			}, timeout, pollInterval).Should(Succeed())
+		})
+	})
+})


internal/constants/constants.go
```diff
@@ -59,9 +59,11 @@ const (
...
 	MetricsShutdownTimeout            = 5 * time.Second
 
-	WebhookServerCertSecretName = "kubewarden-webhook-server-cert" //nolint:gosec // This is not a credential
-	ServerCert                  = "tls.crt"
-	ServerPrivateKey            = "tls.key"
+	WebhookServerCertSecretName      = "kubewarden-webhook-server-cert" //nolint:gosec // This is not a credential
+	ServerCert                       = "tls.crt"
+	ServerPrivateKey                 = "tls.key"
+	ServerCertSecretFormatVersion    = "1"
+	ServerCertSecretFormatAnnotation = "kubewarden/cert-format-version" //nolint:gosec // This is not a credential
```
jvanz: As far as I can remember, it's a good practice use a domain prefix in the annotations name. Something like: 
```suggestion
	ServerCertSecretFormatAnnotation = "kubewarden.io/cert-format-version" //nolint:gosec // This is not a credential
```

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+		policyServer = &policiesv1.PolicyServer{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: testPolicyServerName,
+			},
+		}
+
+		// Set up scheme
+		scheme = runtime.NewScheme()
+		Expect(corev1.AddToScheme(scheme)).To(Succeed())
+		Expect(policiesv1.AddToScheme(scheme)).To(Succeed())
```
jvanz: Why do you need to do this if this is done in the suite_test.go?

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+	When("a secret without format annotation exists", func() {
+		It("should regenerate the certificate and add the annotation", func() {
+			By("creating an old cert without the annotation")
+			now := time.Now()
+			oldCert, oldPrivateKey, err := certs.GenerateCert(
+				caCert,
+				caPrivateKey,
+				now,
+				now.Add(constants.ServerCertExpiration),
+				"policy-server-default.test-namespace.svc",
```
flavio: make use of the contents of the variables you defined earlier to build this string

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	policiesv1 "github.com/kubewarden/kubewarden-controller/api/policies/v1"
+	"github.com/kubewarden/kubewarden-controller/internal/certs"
+	"github.com/kubewarden/kubewarden-controller/internal/constants"
+)
+
+var _ = Describe("PolicyServer certificate secret reconciliation", func() {
+	ctx := context.Background()
+	const testNamespace = "test-namespace"
```
jvanz: There is a test namespace defined in the suite_test.go. Unless there is a reason for this, I think we can use that namespace.

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+				now,
+				now.Add(constants.ServerCertExpiration),
+				"policy-server-default.test-namespace.svc",
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("creating policy server secret WITHOUT annotation (simulates old format)")
+			oldSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "policy-server-default",
```
flavio: I don't remember the full code right now, but if we have a helper that builds the name of the secret starting from the name of the policy-server, then lets make use of it.

internal/controller/policyserver_controller_cert_secret_test.go
jvanz: Small suggestion, maybe we should build this URL to use the test namespace defined above (or from the suite_test.go). Just to avoid remember that we need to update this string if the namespace changed. 

I do not consider this suggestion a blocker to merge this PR.

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+					// Intentionally no annotations
+				},
+				Type: corev1.SecretTypeOpaque, // old type
+				Data: map[string][]byte{
+					constants.ServerCert:       oldCert,
+					constants.ServerPrivateKey: oldPrivateKey,
+				},
+			}
+
+			fakeClient := fake.NewClientBuilder().
```
jvanz: Cannot we use the `k8sClient` from suite_test.go? It already have the scheme configuration done.

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+				WithScheme(scheme).
+				WithObjects(caSecret, oldSecret, policyServer).
+				Build()
+
+			reconciler := &PolicyServerReconciler{
```
jvanz: The reconciler is already configured in the test suite preparation in `suite_test.go`. Do we need to do it again here?

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+				constants.PartOfLabelValue,
+			))
+			Expect(updatedSecret.Labels).To(HaveKeyWithValue(
+				constants.ComponentLabelKey,
+				constants.ComponentPolicyServerLabelValue,
+			))
+		})
+	})
+
+	When("a secret with the correct format annotation exists", func() {
```
jvanz: This test has the same request about k8s client and reconciler from the previous one. But I'll not repeat myself to save space in the UI. ;) 

Furthermore, this test look like generated by copilot. This is not a issue on itself. But I think we can ask it to take in consideration the partner of other tests and what is done in `suite_test.go` file. With this additional information in the prompt, I think it will not generate this initialization code for the reconciler and fakeClient that is already done in the test suite initialization.

internal/controller/policyserver_controller_cert_secret.go
```diff
@@ -66,11 +74,13 @@ func (r *PolicyServerReconciler) reconcilePolicyServerCertSecret(ctx context.Con
 				return fmt.Errorf("cannot generate policy-server %s certificate: %w", policyServer.NameWithPrefix(), err)
 			}
 
-			policyServerSecret.Type = corev1.SecretTypeOpaque
+			policyServerSecret.Type = corev1.SecretTypeTLS
```
jvanz: I was researching about this change and I think we have a blocker here. Kubernetes does not allow type change. In other words, the reconciler will fail when trying to patch/update the secret because we are changing the type. We need to delete the secret and recreated. 

I've ran an script trying to update a secret:

```console
=== Creating test namespace test-secret-type-20852 ===
namespace/test-secret-type-20852 created

=== Step 1: Create an Opaque secret with tls.crt and tls.key ===
secret/test-secret-type-change created
Type: Opaque

=== Step 2: Try to change the secret type from Opaque to kubernetes.io/tls via kubectl apply ===
Warning: resource secrets/test-secret-type-change is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
The Secret "test-secret-type-change" is invalid: type: Invalid value: "kubernetes.io/tls": field is immutable
RESULT: Type change via apply FAILED

=== Step 3: Try to change the secret type via kubectl replace ===
The Secret "test-secret-type-change" is invalid: type: Invalid value: "kubernetes.io/tls": field is immutable
RESULT: Type change via replace FAILED

=== Step 4: Try to patch the secret type via kubectl patch (strategic merge) ===
The Secret "test-secret-type-change" is invalid: type: Invalid value: "kubernetes.io/tls": field is immutable
RESULT: Type change via patch FAILED

=== Step 5: Try to patch the secret type via server-side apply ===
The Secret "test-secret-type-change" is invalid: type: Invalid value: "kubernetes.io/tls": field is immutable
RESULT: Type change via server-side apply FAILED

=== Final secret state ===
Type: Opaque

=== Done ===
Cleaning up namespace test-secret-type-20852...

```



I think that's why we do not change the secret type in the past. Changing the secret types requires more complex upgrade logic. Because we need to delete  and recreate the secret 

I think we should create an tech-deb issue to address this. As stated in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets), this is a convenience:

> The TLS Secret type is provided only for convenience. You can create an Opaque type for credentials used for TLS authentication. However, using the defined and public Secret type (kubernetes.io/tls) helps ensure the consistency of Secret format in your project. The API server verifies if the required keys are set for a Secret of this type.

Therefore, I think we can keep this change as small as possible and resolve the type  change later. In the end of the day, there is no actual benefit of doing it and it will require more complex upgrade logic.

internal/controller/policyserver_controller_cert_secret_test.go
```diff
@@ -0,0 +1,247 @@
...
+			Expect(updatedSecret.Annotations).To(HaveKeyWithValue(
+				constants.ServerCertSecretFormatAnnotation,
+				constants.ServerCertSecretFormatVersion,
+			))
+
+			By("verifying the certificate data exists")
+			Expect(updatedSecret.Data).To(HaveKey(constants.ServerCert))
+			Expect(updatedSecret.Data[constants.ServerCert]).ToNot(BeEmpty())
+			Expect(updatedSecret.Data).To(HaveKey(constants.ServerPrivateKey))
+			Expect(updatedSecret.Data[constants.ServerPrivateKey]).ToNot(BeEmpty())
```
flavio: Here we have to make sure the old secret is not updated. I would compare the `metadata.resourceVersion` value and ensure it didn't change as a result of the reconciliation.

internal/controller/policyserver_controller_cert_secret.go
flavio: @jvanz thanks for the explanation! I agree, let's keep using the `opaque` type for the time being

viccuad:
Please review again.

I have dropped the change of the Secret `type: Opaque` for now and tackled some other things from the review.

I have redone the test correctly, this time reusing the test utils we already provide, which simplifies them a lot. Now we check for:
- Secret creation with correct annot for the first time
- When old Secret is missing the versioning annot, regenerate the
  certificate, and add the version annot (current upgrade path at time of writing)

internal/constants/constants.go
viccuad: Done.

PR #1503, opened 2026-02-20
feat(policy-evaluator): support field masks when querying k8s

flavio:
This commit allows policies to instruct the host to strip away unwanted fields when querying and listing Kubernetes resources.

This can be particularly useful when the Kubernetes resource being exchanged is particularly big, and the policy cares about a limited subset of its data.

The policy can now use a field mask to determine which fields are relevant. The field mask is expressed as a list of strings that can be dot separated.

For example: ['spec.containers.name', 'spec.containers.image'] applied to a Pod will remove all the fields, except for `spec.containers[]`, but the entries of that list will be object that have only the `name` and `image` keys.

This feature can lead to:

- less memory usage on the PolicyServer: when listing for resources, the PolicyServer creates a reflector, that keeps in memory all the objects of that type (eventually across the whole cluster)
- faster and more efficient policy evaluation: less data needs to be moved between the host and the wasm guest. All this data is also serialized by the host, and deserialized by the guest

Older policies will not notice this change, even the ones making use of Kubernetes host capabilities.

Depends on: https://github.com/kubewarden/policy-sdk-rust/pull/213

### DIFF
--- a/crates/policy-evaluator/src/callback_handler/kubernetes/field_mask.rs
+++ b/crates/policy-evaluator/src/callback_handler/kubernetes/field_mask.rs
@@ -134,7 +134,7 @@ mod tests {
             },
             "d": 3
         }),
-        vec!["a.b", "d"])
+        &["a.b", "d"])
     ]
     #[case::nested_objects(
         json!({
@@ -156,7 +156,7 @@ mod tests {
                 }
             }
         }),
-        vec!["user.name", "user.metadata.id"]
+        &["user.name", "user.metadata.id"]
     )]
     #[case::arrays(
         json!({
@@ -174,7 +174,7 @@ mod tests {
                 {"name": "Bob"}
             ]
         }),
-        vec!["users.name"]
+        &["users.name"]
     )]
     #[case::arrays_with_nested_objects(
         json!({
@@ -198,7 +198,7 @@ mod tests {
                 }
             ]
         }),
-        vec!["users.name", "users.metadata"]
+        &["users.name", "users.metadata"]
     )]
     #[case::keep_all_below(
         json!({
@@ -222,12 +222,12 @@ mod tests {
                 }
             }
         }),
-        vec!["user"]
+        &["user"]
     )]
     fn test_prune_in_place(
         #[case] mut input: Value,
         #[case] expected: Value,
-        #[case] mask: Vec<&str>,
+        #[case] mask: &[&str],
     ) {
         let root = FieldMaskNode::new(mask);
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -12,7 +12,7 @@ itertools = "0.14.0"
 k8s-openapi = { version = "0.27.0", default-features = false, features = [
   "v1_33",
 ] }
-kubewarden-policy-sdk = { version = "0.15.1" }
+kubewarden-policy-sdk = { git = "https://github.com/flavio/policy-sdk-rust", branch = "k8s-field-masks" }
```
jvanz: We need to invert this dependency. The SKD should depend on the crates on this repository. Therefore, the PR on this repo can be independent. It looks weird that core components depends on a SDK to expose core components features. ;)

crates/policy-evaluator/src/callback_handler/kubernetes/reflector.rs
jvanz: Maybe we need to update this documentation to mention the `field_masks` now.

crates/policy-evaluator/src/callback_handler/kubernetes/field_mask.rs
```diff
@@ -0,0 +1,260 @@
...
+
+/// Represents a node in a field mask tree.
+///
+/// This structure is used to define and operate on masks that specify which parts of a JSON object should
+/// be retained or pruned.
+#[derive(Default)]
+pub struct FieldMaskNode {
+    children: BTreeMap<String, FieldMaskNode>,
+    // If true, we keep everything below this point (e.g. user asked for "metadata")
+    is_terminal: bool,
```
jvanz: Cannot we remove this `is_terminal` field? If children is empty, it's a leaf node in the tree, right?

Cargo.toml
jvanz: This is not a blocker for this PR. It just something that come to my mind

crates/policy-evaluator/src/callback_handler/kubernetes/field_mask.rs
flavio: good point, it's useless and I've removed it

crates/policy-evaluator/src/callback_handler/kubernetes/reflector.rs
flavio: good point, done

Cargo.toml
flavio: I acknowledge this is painful, but we need that to inside of [policy-evaluator callback handler](https://github.com/kubewarden/kubewarden-controller/blob/main/crates/policy-evaluator/src/callback_requests.rs) as a way to convert from the callback types provided by the guest into our [`CallbackRequestType`](https://github.com/kubewarden/kubewarden-controller/blob/28efbe571fa6e749a22a6b697829e11476bbc7f8/crates/policy-evaluator/src/callback_requests.rs#L33) enum.

I don't think we could invert the dependency. It would also mean we would need to publish another crate on crates.io, and make the current SDK depend on that.

What I think would make everything more simpler would be to move the Rust SDK into this repository. We would archive the old repository and we would change the automation to tag its releases from here. That's a bit of work though... do we want to embrace it?

jvanz: As we discussed in our daily let keep it as it is.

PR #1512, opened 2026-02-25
feat: audit-scanner deletes old wgpolicyk8s reports when migrating to openreports

viccuad:
## Description

Relates to https://github.com/rancher/kubewarden-ui/issues/1324

When the kubewarden-controller Helm chart is deployed with `.Values.auditScanner.reportCRDsKind "openreports"`, we are performing a migration from old PolicyReports from wgpolicyk8s.io to new OpenReports from openreports.io.

Add the following feature to audit-scanner:
When audit-scanner is run with `--report-kind openreports`, it will delete all old wgpolicyk8s reports that are labelled as owned by kubewarden.

It does this efficiently by reusing existing deleteOldReports() functions that perform a `deletecollection` call (and not multiple `delete` calls), and by listing with limit=1 prior to attempting deletion.

Note:
The audit-scanner already deletes old reports, but only the type it's configured to create:
- If configured to create old policyreports, it deletes all old policyreports as they are stale and recreates them.
- If configured to create new openreports, it deletes all old openreports as they are stale and recreates them. But it doesn't delete old policyreports.

## Test

Added unit tests for this new function.
The function reuses existing delete functions that already have tests.

### DIFF
--- a/internal/audit-scanner/report/legacy_migration.go
+++ b/internal/audit-scanner/report/legacy_migration.go
@@ -43,46 +43,36 @@ func DeleteAllLegacyPolicyReports(ctx context.Context, c client.Client, logger *
 
 	clusterReportList := &wgpolicy.ClusterPolicyReportList{}
 	err = c.List(ctx, clusterReportList, listOpts)
-	if meta.IsNoMatchError(err) {
+	switch {
+	case meta.IsNoMatchError(err):
 		logger.DebugContext(ctx, "wgpolicyk8s.io CRDs not installed, skipping legacy clusterreport cleanup")
-	} else {
-		if err != nil {
-			return fmt.Errorf("failed to list legacy ClusterPolicyReports: %w", err)
+	case err != nil:
+		return fmt.Errorf("failed to list legacy ClusterPolicyReports: %w", err)
+	case len(clusterReportList.Items) > 0:
+		logger.InfoContext(ctx, "Deleting legacy wgpolicyk8s.io ClusterPolicyReports")
+		if err = store.DeleteOldClusterReports(ctx, ephemeralRunUID); err != nil {
+			return err
 		}
-		if len(clusterReportList.Items) > 0 {
-			logger.InfoContext(ctx, "Deleting legacy wgpolicyk8s.io ClusterPolicyReports")
-			err = store.DeleteOldClusterReports(ctx, ephemeralRunUID)
-			if err != nil {
-				return err
-			}
-		}
-
 	}
 
 	policyReportList := &wgpolicy.PolicyReportList{}
 	err = c.List(ctx, policyReportList, listOpts)
-	if meta.IsNoMatchError(err) {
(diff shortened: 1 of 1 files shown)


internal/audit-scanner/report/legacy_migration.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	// After first migration, we will perform 2 list calls, one for
+	// ClusterPolicyReport, another for PolicyReport, both with limit=1.
+	listOpts := &client.ListOptions{LabelSelector: labelSelector, Limit: 1}
+
+	clusterReportList := &wgpolicy.ClusterPolicyReportList{}
+	err = c.List(ctx, clusterReportList, listOpts)
+	if err != nil {
+		return fmt.Errorf("failed to list legacy ClusterPolicyReports: %w", err)
+	}
```
flavio: I would try to catch the error raised when a CRD is not defined. In that case I would ignore the error and continue.

internal/audit-scanner/report/legacy_migration.go
```diff
@@ -0,0 +1,80 @@
...
+		err = store.DeleteOldClusterReports(ctx, ephemeralRunUID)
+		if err != nil {
+			return err
+		}
+	}
+
+	policyReportList := &wgpolicy.PolicyReportList{}
+	err = c.List(ctx, policyReportList, listOpts)
+	if err != nil {
+		return fmt.Errorf("failed to list legacy PolicyReports: %w", err)
```
flavio, line 64: same here, but here we can safely return immediately with `nil`

internal/audit-scanner/report/legacy_migration.go
viccuad: Done

internal/audit-scanner/report/legacy_migration.go
viccuad: Done

internal/audit-scanner/report/legacy_migration.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// After first migration, we will perform 2 list calls, one for
+	// ClusterPolicyReport, another for PolicyReport, both with limit=1.
+	listOpts := &client.ListOptions{LabelSelector: labelSelector, Limit: 1}
+
+	clusterReportList := &wgpolicy.ClusterPolicyReportList{}
+	err = c.List(ctx, clusterReportList, listOpts)
+	if meta.IsNoMatchError(err) {
+		logger.InfoContext(ctx, "wgpolicyk8s.io CRDs not installed, skipping legacy report cleanup")
+		return nil
```
flavio: by returning here we will never try to remove the `PolicyReports`. However, I think it's reasonable, because someone who doesn't have the `ClusterPolicyReports` CRD defined won't almost certainly not have the `PolicyReport` CRD defined either.

PR #1513, opened 2026-02-25
    feat(controller): flag to allow policies in kubewarden namespace

### LINKED ISSUE #1201: Feature Request: Allow users to specify policies that also apply to kubewarden namespace

Currently the kubewarden-controller adds an exclusion rule to all webhookconfigurations to ignore the namespace in which kubewarden itself is running.

We would like a feature toggle to be added to allow policies to bypass this anti-lockout mechanism so that some resources can be protected/mutated even in the kubewarden namespace.

The implementation would ideally introduce a feature toggle on the clusteradmissionpolicy resource that is evaluated for true false in https://github.com/kubewarden/kubewarden-controller/blob/3c5e2fd8cc68c592f833539d42c5a92624839293/internal/controller/policy_subreconciler_webhook.go#L193 (with a default of apply the exclusion of kubewarden namespace).

### Solution you'd like

So flow would be:

```
kind: ClusterAdmissionPolicy
metadata:
  name: loadbalancerdeny
spec:
  module: registry://<registry>v0.0.1
  settings: {}
  rules:
  - apiGroups:
    - ''
    apiVersions:
    - v1
    resources:
    - namespaces
    operations:
    - CREATE
    - UPDATE
  mutating: false
  defaultexclusion: true
```
If you don't want the webhook to have the deployed namespace for kubewarden excluded.
This setting would default to false if not specified and thereby allow current users to experience existing behaviour of kubewarden.

_No response_

### Anything else?

_No response_

jvanz:
## Description

Adds a new field called "allowInsideKubewardenNamespace" in the policy spec. This field allow the policy to valuate resources in the Kuberwarden namespace. When this field is "true" the controller will NOT add the namespace selector to skip resources from the Kubewarden namespace. The field has default value "false" to keep backwards compatibility.

Fix #1201

### DIFF
--- a/api/policies/v1/admissionpolicy_types.go
+++ b/api/policies/v1/admissionpolicy_types.go
@@ -133,6 +133,10 @@ func (r *AdmissionPolicy) GetNamespaceSelector() *metav1.LabelSelector {
 	}
 }
 
+func (r *AdmissionPolicy) GetAllowInsideKubewardenNamespace() bool {
+	return false
+}
+
 func (r *AdmissionPolicy) GetObjectSelector() *metav1.LabelSelector {
 	return r.Spec.ObjectSelector
 }
--- a/api/policies/v1/admissionpolicygroup_types.go
+++ b/api/policies/v1/admissionpolicygroup_types.go
@@ -157,6 +157,10 @@ func (r *AdmissionPolicyGroup) GetNamespaceSelector() *metav1.LabelSelector {
 	}
 }
 
+func (r *AdmissionPolicyGroup) GetAllowInsideKubewardenNamespace() bool {
+	return false
+}
+
 func (r *AdmissionPolicyGroup) GetObjectSelector() *metav1.LabelSelector {
 	return r.Spec.ObjectSelector
 }
--- a/api/policies/v1/clusteradmissionpolicy_types.go
+++ b/api/policies/v1/clusteradmissionpolicy_types.go
@@ -88,6 +88,17 @@ type ClusterAdmissionPolicySpec struct {
 	// the policy is assigned to.
 	// +optional
 	ContextAwareResources []ContextAwareResource `json:"contextAwareResources,omitempty"`
+
+	// AllowInsideKubewardenNamespace controls whether the policy should also be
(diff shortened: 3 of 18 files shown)


Makefile
```diff
@@ -50,7 +50,7 @@ helm-unittest:
 
 .PHONY: test-e2e
 test-e2e: controller-image audit-scanner-image policy-server-image
-	$(GO_BUILD_ENV) go test ./e2e/ -v
+	$(GO_BUILD_ENV) go test ./e2e/ -v -timeout 20m
```
jvanz: This is necessary because the e2e tests are too slow and exceedes the default timeout (10m). To improve that I think we need to refactor them to reuse resource installed in the cluster. 

It looks to me that as we always create new policy servers for every test the controller starts to get slow because of the low resource amount in the test environment and many resource to reconcile. In this case, we have a worst scenario because the test reconfigure the controller. Which also trigger a full reconciliation of all previous resources running in the cluster. 

But as I don't want to do a full refactor of this tests. I just increase the timeout for now. I've doubled the time, but the test are not consuming the whole time. In my environment it took 11m.

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -53,6 +53,9 @@ spec:
         - --webhook-service-name={{ include "kubewarden-controller.fullname" . }}-webhook-service
         - --always-accept-admission-reviews-on-deployments-namespace
         - --zap-log-level={{ .Values.logLevel }}
+       {{- if .Values.allowClusterAdmissionPoliciesInDeploymentsNamespace }}
+        - --allow-cluster-admission-policies-in-deployments-namespace
```
flavio: nit: I would call this flag `--allow-cluster-admission-policies-in-kubewarden-namespace`

This makes more explicit the purpose

Makefile
flavio: I think we can remove the e2e tests you added to this PR. We're already testing the selector inside of `internal/controller/policy_subreconciler_webhook_test.go`. This is enough.

Generally speaking, we should not have e2e tests for everything because they are quite expensive.

flavio (approved):
I like the new design!

PR #1514, opened 2026-02-25
chore(deps): update wasmtime

flavio:
Update wasmtime dependency. This upgrade introduced a breaking change that had rippling effects across all our crates.

Now `wasmtime::error::Error` is no longer an alias to `anyhow::Error`. This required quite some changes across our error types and conversions.

BTW: the policy timeout protection is working as expected. This change didn't break it.

I had some failures locally when running the integration tests, but they were unrelated. It's about the flaky tests.


flavio:
the failure is unrelated :cry:

PR #1515, opened 2026-02-26
feat(kwctl,policy-server): Honor proxy configuration for context aware calls

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/1300
I recommend reviewing this per-commit.

Honour proxy configuration for context aware calls in kwctl, policy-server. This includes OCI registry traffic, Sigstore API calls.
 
To achieve this, 
Add a `ProxyConfig{}` struct. Refactor our approach on reading env vars by providing a `ProxyConfig::from_env()` and funnel all the proxy configuration through it.

This is needed, as policy-fetcher depends on reqwest with default features, and that includes `system-proxy` which allows it to read the env vars. But kwctl and policy-server depend on reqwest without this feature, hence the proxy configuration needs to be funneled to their clients.

To funnel the proxy configuration, extend the `Sources{}` struct with `Sources.proxies: Option(ProxyConfig)`.
When None, `from_env()` is used as fallback so existing callers are unaffected.
Also, add a  `Sources::proxies()` getter that returns the explicit config or falls back to env vars.
With this `Sources{}` struct changes, pass it or consume it where needed:
- policy-evaluator: when creating clients such as the sigstore-rs or the oci-client,
- policy-fetcher: onimplementations of `fetch()`
- kwctl, policy-server: masking the existing `Sources{}`

## Test

Added unit tests and integration tests as needed.
policy-evaluator has integration tests to see that proxies are  context aware calls. This mirror the testcontainer tests done for kwctl.
policy-fetcher has unit tests for the different `fetch()` implementations, as well as 

### Potential improvement

The `RawSources`  struct is left unimplemented, done in PR https://github.com/kubewarden/kubewarden-controller/pull/1516. This will provide the option to pass the proxy configurations via `sources.yaml` for kwctl, policy-server.

### DIFF
--- a/crates/policy-evaluator/tests/proxy.rs
+++ b/crates/policy-evaluator/tests/proxy.rs
@@ -4,7 +4,9 @@ mod common;
 // have Docker installed.
 #[cfg(not(target_os = "macos"))]
 mod proxy_tests {
-    use crate::common::setup_callback_handler;
+    use std::time::Duration;
+
+    use backon::{ExponentialBuilder, Retryable};
     use policy_evaluator::callback_requests::{CallbackRequest, CallbackRequestType};
     use policy_fetcher::{proxy::ProxyConfig, sources::Sources};
     use testcontainers::{
@@ -13,6 +15,8 @@ mod proxy_tests {
         runners::AsyncRunner,
     };
     use tokio::sync::oneshot;
+
+    use crate::common::setup_callback_handler;
 
     async fn start_proxy() -> (ContainerAsync<GenericImage>, u16) {
         let container = GenericImage::new("kalaksi/tinyproxy", "1.7")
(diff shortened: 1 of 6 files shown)


crates/policy-fetcher/src/sources.rs
```diff
@@ -393,4 +399,35 @@ Wm7DCfrPNGVwFWUQOmsPue9rZBgO
             assert_eq!(actual_cert, &expected_cert);
         }
     }
+
+    #[test]
+    fn test_proxy_config_explicit_takes_precedence_over_env() {
```
flavio, line 404: I think it would be nicer to set some proxy variable as using the env variables. Otherwise we have to assume that a non-working code would return `None` for proxies.

crates/policy-evaluator/tests/proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+                .unwrap_or_default();
+            let stderr = String::from_utf8(container.stderr_to_vec().await.unwrap_or_default())
+                .unwrap_or_default();
+            if stdout.contains(needle) || stderr.contains(needle) {
+                return true;
+            }
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
+        false
+    }
```
jvanz, line 58: I think we can use the backon crate here. This is **not** a must have and it's not blocking the PR. But I think this  would be nice to have. Furthermore, we are already using the crate in the policy-server integration [tests](https://github.com/kubewarden/kubewarden-controller/blob/fa24a6ee998c0b2966d4d728703f58a71cd511e1/crates/policy-server/tests/integration_test.rs#L804).

flavio: good point, I agree about using the crate. Also, I don't think this is a blocker

crates/policy-fetcher/src/sources.rs
viccuad: Indeed, done in https://github.com/kubewarden/kubewarden-controller/pull/1515/commits/816fbdc817415e15b61a5a2f64e6d7749fe938fd.

crates/policy-evaluator/tests/proxy.rs
viccuad: Indeed, I will implement this.

viccuad: Done for all tests with https://github.com/kubewarden/kubewarden-controller/pull/1515/commits/da4d73b6e7dc2859f52ea38a3a1f2cc3b7bc2091.

jvanz (approved):

PR #1520, opened 2026-02-26
refactor(charts): Change generation of RBAC templates

jvanz:
## Description

Updates the controller-gen command used to generated the RBAC manifests into the charts/kubewarden-controller directory. This commit also adds the kubebuilder markers to add the missing permissions in the controller roles that was added using manully created roles definitions. 

Fix https://github.com/kubewarden/kubewarden-controller/issues/1483

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -462,13 +462,30 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
-
       - name: enable git long paths on Windows
         if: matrix.os == 'windows-latest'
         run: git config --global core.longpaths true
 
       - name: Run cargo check
         run: make -C crates/kwctl check
+
+  check-manifests:
+    name: Check if the controller-gen generated manifests are up to date
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version: "1.26"
+          check-latest: true
+      - run: |
+          make manifests
+          if ! git diff --quiet -- charts/; then
+            echo "Changes detected in charts/. Please run 'make manifests' and commit the changes."
+            exit 1
+          fi
 
   # Rollup job for branch protection - single stable job name that depends on all checks
   ci-success:
@@ -496,6 +513,7 @@ jobs:
       - validate-hauler-manifest
       - kwctl-docs
       - check-kwctl-cross-platform
+      - check-manifests
     runs-on: ubuntu-latest
     steps:
       - name: Check all jobs status
--- a/.github/workflows/autolabeler.yml
+++ b/.github/workflows/autolabeler.yml
@@ -2,6 +2,7 @@ name: Auto label
 
 on:
   pull_request:
+    # Only following types are handled by the action, but one can default to all as well
     types: [opened, reopened, synchronize, edited]
 
 permissions: {}
@@ -11,8 +12,8 @@ jobs:
     # Skip fork PRs — the GITHUB_TOKEN is read-only and cannot add labels
     if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
     permissions:
-      pull-requests: write
       contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: release-drafter/release-drafter/autolabeler@139054aeaa9adc52ab36ddf67437541f039b88e2 # v7.1.1
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -102,7 +102,7 @@ jobs:
           persist-credentials: false
 
       - name: "Install kwctl"
-        uses: kubewarden/github-actions/kwctl-installer@f301a7874dd642510fff54a89e4329881bf871ef # v4.6.0
+        uses: kubewarden/github-actions/kwctl-installer@a03315e95ccf85c92e5d472824edeab0704f857b # v4.6.1
         with:
           KWCTL_VERSION: latest
 


charts/kubewarden-controller/templates/NOTES.txt
```diff
@@ -4,6 +4,22 @@ You can now start defining admission policies by using the cluster-wide
...
+
+  kubectl delete clusterrole kubewarden-controller-manager-cluster-role
+  kubectl delete clusterrole kubewarden-controller-metrics-reader
+  kubectl delete clusterrole kubewarden-controller-proxy-role
+  kubectl delete clusterrolebinding kubewarden-controller-proxy-rolebinding
+  kubectl delete rolebinding kubewarden-controller-leader-election-rolebinding -n {{ .Release.Namespace }}
+  kubectl delete rolebinding kubewarden-controller-manager-namespaced-rolebinding -n {{ .Release.Namespace }}
+  kubectl delete role kubewarden-controller-leader-election-role -n {{ .Release.Namespace }}
+  kubectl delete role kubewarden-controller-manager-namespaced-role -n {{ .Release.Namespace }}
+
```
jvanz: Do you want to move this to a hook? I'm not big fan of hooks. But this commands seems to be very simple. Maybe it's not a bad idea.

charts/kubewarden-controller/templates/controller-rbac-extras.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: kubewarden-controller-metrics-reader
+  labels:
+    {{- include "kubewarden-controller.labels" . | nindent 4 }}
+  annotations:
+    {{- include "kubewarden-controller.annotations" . | nindent 4 }}
+rules:
+- nonResourceURLs:
+  - /metrics
+  verbs:
+  - get
```
jvanz: I think we can remove this ClusterRole. I cannot see any place where it used. I'm keeping it for now just in case I'm missing something.

scripts/post-generate-manifests.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# ── 1. Rename CRD files ──────────────────────────────────────────────────────
+echo "Renaming CRD files in ${CRD_CHART_DIR}..."
+for src in "${CRD_CHART_DIR}"/policies.kubewarden.io_*.yaml; do
+    [[ -e "${src}" ]] || continue
+    filename="$(basename "${src}")"
+    dest="${CRD_CHART_DIR}/${filename#policies.kubewarden.io_}"
+    mv "${src}" "${dest}"
+    echo "  ${filename} -> $(basename "${dest}")"
+done
```
flavio: I'm also fine leaving them with the name generated by controller-gen

scripts/post-generate-manifests.sh
```diff
@@ -0,0 +1,79 @@
...
+{ print }
+' "${RBAC_FILE}" > "${RBAC_FILE}.tmp"
+
+mv "${RBAC_FILE}.tmp" "${RBAC_FILE}"
+
+# ── 3. Replace hardcoded namespace with Helm template expression ─────────────
+echo "Replacing hardcoded namespace in ${RBAC_FILE}..."
+sed -i 's/  namespace: kubewarden/  namespace: {{ .Release.Namespace }}/' "${RBAC_FILE}"
+echo "  Done."
+fi
```
flavio: I would prefer to reduce the amount of code and reuse what sbomscanner and the runtime-enforcer projects are doing:

https://github.com/kubewarden/sbomscanner/blob/0945621e258fb83aa933b58ca089f44e2e5eddf4/Makefile#L99-L105

charts/kubewarden-controller/templates/controller-rbac-extras.yaml
flavio: I think we need that to allow the prometheus SA to scrape our metrics. Let's leave it around

jvanz: Ok, let's leave it for now. But I'll open an issue to verify this. I have the impressions this is a leftover from some point in time.

flavio: I see a new contribution came from community that removes it. I think we can give it a try and remove this role

jvanz: Done

charts/kubewarden-controller/templates/NOTES.txt
flavio: Are you sure these roles are not automatically removed by helm? afterall these are resources that belonged to a previous installations, that are then removed by a new version of the chart

charts/kubewarden-controller/templates/controller-rbac-roles.yaml
```diff
@@ -0,0 +1,155 @@
...
+  resources:
+  - mutatingwebhookconfigurations
+  - validatingwebhookconfigurations
+  verbs:
+  - create
+  - delete
+  - list
+  - patch
+  - watch
+- apiGroups:
```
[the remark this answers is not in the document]
flavio: Good catch from copilot:

- https://github.com/kubewarden/kubewarden-controller/blob/42c079b24bb0215c0e3de73000b93c1c70166be2/internal/controller/policy_subreconciler_webhook.go#L95
- https://github.com/kubewarden/kubewarden-controller/blob/42c079b24bb0215c0e3de73000b93c1c70166be2/internal/controller/policy_subreconciler_webhook.go#L181

We have to update the annotations.

We also have to understand why all the tests are green despite this change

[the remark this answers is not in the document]
flavio: I've tasked an AI agent to find why the tests are green. This is the answer I got, which seems reasonable to me:

## Unit Tests — system:masters Bypasses All RBAC
envtest starts a real kube-apiserver with --authorization-mode=RBAC, so RBAC is technically enabled. However, controller-runtime's testEnv.Start() unconditionally returns a restConfig authenticated as a member of the system:masters group (envtest/server.go):
adminInfo := User{Name: "admin", Groups: []string{"system:masters"}}
system:masters is a hardcoded Kubernetes superuser group that bypasses the RBAC authorizer entirely — not via a ClusterRoleBinding, but at the authorizer level itself. No RBAC policy, no matter how restrictive, is evaluated for this identity. The manager and all reconcilers inherit these credentials, so every r.Get(...) call silently succeeds regardless of what the RBAC manifest says.
Additionally, policy_subreconciler_webhook_test.go doesn't test the reconcile methods at all — it only tests the namespaceSelector() helper in isolation with no client involved, so even in principle there's no API call to be blocked.

## E2E Tests — Test Harness Also Runs as system:masters
The e2e tests use a Kind cluster provisioned by sigs.k8s.io/e2e-framework. Kind's generated kubeconfig identifies the user as kubernetes-admin, which is also a member of system:masters. The entire test harness — namespace creation, Helm installs, assertion calls — runs under this identity, bypassing RBAC.
The controller's ServiceAccount does get the restricted Helm chart ClusterRole applied (the one missing get), but the tests only observe effects (e.g., "does a ValidatingWebhookConfiguration with this name now exist?"), not whether the controller reconciled without errors. They don't:
- Inspect the controller's logs for 403 errors
- Check the policy's .status conditions for reconcile failures
- Impersonate the controller's ServiceAccount to validate its permissions directly


I do agree we have to extend the e2e tests, but I do not agree with the remediation paths being proposed. I think we should understand how the `GET` operation is used, and check for changes that depend on that. I would not grep into the logs

Makefile
```diff
@@ -145,8 +146,13 @@ generate-controller: manifests  ## Generate code containing DeepCopy, DeepCopyIn
...
-	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=controller-role crd webhook paths="./api/policies/v1"  paths="./internal/controller" output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac
+manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=kubewarden-controller-manager,fileName=controller-rbac-roles.yaml crd webhook \
+			paths="./api/policies/v1"  paths="./internal/controller" paths="./cmd/controller" \
+			output:crd:artifacts:config=charts/kubewarden-crds/templates \
+			output:rbac:artifacts:config=charts/kubewarden-controller/templates
+	sed -i '/^metadata:/a\  labels:\n    {{- include "kubewarden-controller.labels" . | nindent 4 }}\n  annotations:\n    {{- include "kubewarden-controller.annotations" . | nindent 4 }}' charts/kubewarden-controller/templates/controller-rbac-roles.yaml
+	sed -i 's/  namespace: kubewarden/  namespace: {{ .Release.Namespace }}/' charts/kubewarden-controller/templates/controller-rbac-roles.yaml
```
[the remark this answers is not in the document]
jvanz: Nah... I see the potential issue. But we do not support commands from other platforms in the admission controller repositories. Thus, I think we are fine to keep it as it is for now.

charts/kubewarden-controller/templates/NOTES.txt
jvanz: Yeah, that's right. Helm will do the 3 way merge and remove this resources.

charts/kubewarden-controller/templates/controller-rbac-roles.yaml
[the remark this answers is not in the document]
jvanz: That's a good catch indeed. I'm on it. Let me update some test to avoid issues in the future.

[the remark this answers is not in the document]
jvanz: We hit a very subtle bug here. The get is run only when a policy is deleted. Taking a look in the code, the policy deletion should not work. Because the code expects to find the webhook before delete it:

https://github.com/kubewarden/kubewarden-controller/blob/42c079b24bb0215c0e3de73000b93c1c70166be2/internal/controller/policy_subreconciler_webhook.go#L93-L105

However, the webhooks are being removed. We never noticed this before because of cache. Which is loaded successfully because the watch/list permission are there. The get never fails because it never hit the API server. It always load data from the cache. To validate that, I changes the manager configuration disable cache for the webhook types:

```go
		Client: client.Options{
			Cache: &client.CacheOptions{
				DisableFor: []client.Object{
					&admissionregistrationv1.ValidatingWebhookConfiguration{},
					&admissionregistrationv1.MutatingWebhookConfiguration{},
				},
			},
		},


```

When I run a controller image based on main with the above change the reconciliation loop fail at the start:

```
{"level":"error","ts":"2026-03-11T19:43:14Z","msg":"Reconciler error","controller":"clusteradmissionpolicy","controllerGroup":"policies.kubewarden.io","controllerKind":"ClusterAdmissionPolicy","ClusterAdmissionPolicy":{"name":"do-not-share-host-paths"},"namespace":"","name":"do-not-share-host-paths","reconcileID":"6b6962fc-c4a2-44d4-b56d-d5dddfcdae77","error":"error reconciling validating webhook\ncannot reconcile validating webhook: validatingwebhookconfigurations.admissionregistration.k8s.io \"clusterwide-do-not-share-host-paths\" is forbidden: User \"system:serviceaccount:kubewarden:kubewarden-controller\" cannot get resource \"validatingwebhookconfigurations\" in API group \"admissionregistration.k8s.io\" at the cluster scope","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:495\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:438\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:313"}

```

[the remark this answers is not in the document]
jvanz: As first remediation for the out of synchrony of the permissions, I've added a CI step to detect when the controller-gen output diverge from the Helm chart files. Therefore, we spot if we forgot to update the markers or change the files without update the markers. 

This is **not** the fix for the tests. I think this will be more complex to configure because that may require some change in the test environment and ensure that the resources will not load from cache. I'll continue to think on that.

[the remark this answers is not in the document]
jvanz: I've [changed](https://github.com/kubewarden/kubewarden-controller/pull/1520/changes/6858280eb3f0ab76a7924bcfd72c2ae421642ecc) the ginkgo test to ensure that we use a more restricted RBAC role that mimics production environment. I've also disabled the cache for the webhook types. This should be enough to avoid a regression. Now, if I remove the `get` from the role used in the test, I can see the test failing.

internal/controller/suite_test.go
```diff
@@ -80,12 +96,194 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-	k8sManager, err := ctrl.NewManager(restConfig, ctrl.Options{
+	// The deploymentsNamespace must be created before the namespaced Role below.
+	err = k8sClient.Create(ctx, &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: deploymentsNamespace,
+		},
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	// Create RBAC resources that mirror the production Helm chart permissions .
```
viccuad: Here we are hardcoding RBAC, instead of making use of our RBAC files.

Can we instead of defining the structs by hand (why may drift in time), read from file, do a `decoder.Decode()` into an `Unstructured` struct, and then apply  with the `k8sClient.Create(ctx,obj)`?

jvanz: In my first try I do load the resources from Helm chart files. But that require removing templates calls from files or rendering the files first. If we need to do that I would add e2e test instead of the ginkgo ones. But those are slower as well. Thus, I think hardcode the role here can be a good balance between speed and convenience to detect issue early in the development process.

viccuad: Argh, true, I forgot that our output is templated.
I'm ok not doing this, but maybe we should track is as tech debt, I'm scared on forgetting that the tests work as such and slipping regressions in the future.

charts/kubewarden-controller/templates/controller-rbac-roles.yaml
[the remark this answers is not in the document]
flavio: Good debugging @jvanz! I've yet to finish reviewing the PR, but I would NOT want the cache to be disabled for webhook types, at least inside of the production code. I don't care about the tests.

internal/controller/suite_test.go
flavio: I'm concerned about these structs going out of sync with the actual RBAC managed by the helm chart.

Can't we leave things as they were and write a test inside of tests done with the e2e-framework? I don't care about them slow, I'm more concerned about having out of sync data.

viccuad: I'm convinced now on moving this test to `<root>/e2e/`. We can call a `helm template` which uses the default values, to be able to consume the templated YAML.

jvanz: If we move the test to e2e there is no reason to call `helm template`. We can just install the helm chart and run the tests.

jvanz: The only think I still new to refresh my memory is how to change the credentials used to run the commands. Because the default one has "super permissions". But I don't remember on top of my head an way to change it using the `e2e-framework`. Let me see how to do it.

jvanz: Ah, I remember now the main reason why I added this test. But let me refresh our memory about everything we're doing here to give full context. 

This PR originated to fix the manifest generation. While I was doing that, we noticed differences between the permissions from the kubebuilder markers and the actual permissions in the Helm chart files. Therefore, beyond adding the missing markers, we decided to investigate why we didn't spot that before in tests and fix it.

The reason we hadn't spotted this before is described in the analysis on this [thread](https://github.com/kubewarden/kubewarden-controller/pull/1520/changes#r2918030833). In summary, the missing `get` verb was being masked because `list`/`watch` populates the controller-runtime client cache. Therefore, `Get()` calls never reach the API server and RBAC is never evaluated for that verb. Other verbs (`create`, `update`, `delete`, `patch`) always hit the API server and would be caught. `list`/`watch` would crash the controller since the informer cache can't start.

That's why I disabled the cache for webhook configuration resources and used a custom RBAC-restricted user for the manager client in this [commit](https://github.com/kubewarden/kubewarden-controller/pull/1520/changes/0246f81dfb8b26fe17a9d947f2b70532e968c8d4). Both changes are required: `DisableFor` forces `Get()` to the API server, and the restricted user ensures RBAC is actually evaluated (the default envtest user is `system:masters`, which bypasses all RBAC).

So, to remove the hardcoding the RBAC rules we can move this verification to the e2e tests, where Helm installs the real chart and the controller runs under its actual ServiceAccount with production RBAC. However, this is not enough. Because the client in the controller still has the cache by default. Therefore, the issue of the missing `get` verb still can occur.  To fix that, we can add a `--disable-client-cache` flag to the controller that forces all `Get()` calls through the API server, and enable it in e2e via a Helm value. What do you think about this option?

flavio: Thanks for the recap.

> So, to remove the hardcoding the RBAC rules we can move this verification to the e2e tests, where Helm installs the real chart and the controller runs under its actual ServiceAccount with production RBAC. However, this is not enough. Because the client in the controller still has the cache by default. Therefore, the issue of the missing get verb still can occur. To fix that, we can add a --disable-client-cache flag to the controller that forces all Get() calls through the API server, and enable it in e2e via a Helm value. What do you think about this option?

Sounds like a good plan. Let's make sure we document why this flag is needed

viccuad: Indeed, thanks for the explanation. I also like the plan.

charts/kubewarden-controller/values.yaml
```diff
@@ -153,6 +153,11 @@ preDeleteHook:
 alwaysAcceptAdmissionReviewsOnDeploymentsNamespace: true
 # Verbosity of logging. Can be one of 'debug', 'info', 'error'.
 logLevel: info
+# Disable the controller-runtime client cache so that all calls go directly to
+# the API server. This is used on tests to ensures RBAC is fully evaluated for
+# every read operation, including the "get" verb which would otherwise be
+# silently served from cache. Should only be enabled in test environments.
+disableClientCache: false
```
flavio: I would not even add that to the `values.yaml`. Users should not be aware of this flag, since it's something we use only for testing

flavio: I'm fine with this approach, but there's also another approach we could take.

Instead of using a CLI flag we could rely on an environment variable. We could then extend the `values.yaml` to allow the user to inject custom env variables into the controller's deployment.

In this way we wouldn't leave "any trace" of this feature to the user. Meaning, it wouldn't show up inside of values.yaml, the deployment template, the cli help... It would be just an "obscure" env variable that we set.

.github/workflows/ci.yml
```diff
@@ -462,14 +462,29 @@
...
         run: git config --global core.longpaths true
 
       - name: Run cargo check
         run: make -C crates/kwctl check
 
+  check-manifests:
+    name: Check if the controller-gen generated manifests are up to date
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
jvanz: This is already fixed. See line 478

.github/workflows/ci.yml
```diff
@@ -462,14 +462,31 @@ jobs:
...
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version: "1.26"
+          check-latest: true
+      - run: |
+          make manifests
+          if ! git diff --quiet -- charts/; then
+            echo "Changes detected in charts/. Please run 'make manifests' and commit the changes."
```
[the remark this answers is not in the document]
jvanz: We do not use the `config` directory. I can ignore this suggestion

Makefile
```diff
@@ -146,8 +147,13 @@ generate-controller: manifests  ## Generate code containing DeepCopy, DeepCopyIn
 	$(GO_BUILD_ENV) $(CONTROLLER_GEN) object paths="./api/policies/v1"
 
 .PHONY: manifests
-manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects. We use yq to modify the generated files to match our naming and labels conventions.
-	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=controller-role crd webhook paths="./api/policies/v1"  paths="./internal/controller" output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac
+manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=kubewarden-controller-manager,fileName=controller-rbac-roles.yaml crd webhook \
+			paths="./api/policies/v1"  paths="./internal/controller" paths="./cmd/controller" \
+			output:crd:artifacts:config=charts/kubewarden-crds/templates \
```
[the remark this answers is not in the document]
jvanz: We plan to remove the config directory in the future. So, we can ignore this comment for  now.

cmd/controller/main.go
```diff
@@ -131,6 +145,14 @@ func main() {
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
+	if val, ok := os.LookupEnv(constants.DisableClientCacheEnvVar); ok {
+		parsed, err := strconv.ParseBool(val)
+		if err != nil {
+			setupLog.Error(err, "invalid value for "+constants.DisableClientCacheEnvVar+" using default", "value", val)
```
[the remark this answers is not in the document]
flavio: I think copilot has a point here, I would that its suggestion

internal/controller/client_interceptor.go
```diff
@@ -0,0 +1,114 @@
...
+// When the informer cache is disabled, controller-runtime forwards MatchingFields
+// to the API server as a field selector. The API server rejects this with 400
+// because CRDs do not support custom field selectors. The interceptor strips the
+// unsupported field selector, performs a plain List, and then filters the results
+// client-side using the Policy.GetPolicyServer() method — identical behaviour to
+// what the cache-backed path would have produced.
+//
+// All other List calls (no field selector, or an unrecognised field selector key)
+// are delegated to c unchanged.
+func NewCacheBypassClient(c client.WithWatch) client.WithWatch {
```
flavio: this is a bit convoluted, I'm feeling more inclined in not testing this kind of issue.

I would still keep the additional env change you did to the chart though.

@viccuad : what do you think?

viccuad: I agree, I side with dropping client_interceptor.go and the tests, and I find the `.Values.additionalEnvironmentVariables` a good addition.

jvanz: Ok, fine by me.

flavio (approved):

PR #1522, opened 2026-02-27
test(policy-evaluator,policy-server): Fix flaky tests

viccuad:
## Description

- For policy-evaluator, both proxy integration tests are exercising the same OCIManifest call. This means that the call gets cached by the `cached` crate.
  We want to test the hostcalls for those clients that we have configured (oci-client, sigstore). Yet those hostcalls are always cached (they don't provide a `disable_cache`).
  Therefore, just use a different OCI URI so it isn't cached.

- For policy-server, the proxy tests seem to fail trying to obtain the OCI module from the OCI registry as they all hit it at the same time. Running them serially seems to fix this problem.
  

## Test

I ran the test 30 times consecutively locally, it seems fine now.

### DIFF
--- a/crates/kwctl/tests/proxy.rs
+++ b/crates/kwctl/tests/proxy.rs
@@ -1,3 +1,4 @@
+use backon::{BlockingRetryable, ConstantBuilder};
 use predicates::str::contains;
 use rstest::rstest;
 use tempfile::tempdir;
@@ -7,7 +8,6 @@ use testcontainers::{
     runners::SyncRunner,
 };
 
-use backon::{BlockingRetryable, ConstantBuilder};
 use common::setup_command;
 mod common;
 


crates/kwctl/tests/proxy.rs
```diff
@@ -10,6 +7,7 @@ use testcontainers::{
     runners::SyncRunner,
 };
 
+use backon::{BlockingRetryable, ConstantBuilder};
```
flavio: super nit, this should be moved to be part of the previous block of imports

crates/kwctl/tests/proxy.rs
```diff
@@ -46,9 +45,9 @@ fn proxy_log_contains(container: &Container<GenericImage>, needle: &str) -> bool
     };
     check
         .retry(
-            ExponentialBuilder::default()
-                .with_min_delay(Duration::from_millis(100))
-                .with_max_times(5),
+            ConstantBuilder::default()
```
flavio, line 48: I like that, it's a better choice to reduce the execution time of these tests 👍

crates/policy-evaluator/tests/proxy.rs
```diff
@@ -82,7 +82,7 @@ mod proxy_tests {
                 },
                 response_channel: resp_tx,
             })
-            .expect("cannot send callback request");
+            .expect("cannot send callback request q");
```
jvanz: Typo?
```suggestion
            .expect("cannot send callback request");
```

viccuad: Woops, thanks!

crates/kwctl/tests/proxy.rs
viccuad: Done

PR #1542, opened 2026-03-03
fix(charts): allow evaluations in controller namespace.

### LINKED ISSUE #1540: allowInsideKubewardenNamespace has no effect
I install kubewarden v1.33.0-rc1 (default installation) and deploy pod-privileged policy from this yaml

```yaml
kind: ClusterAdmissionPolicy
metadata:
  annotations:
    io.kubewarden.policy.category: PSP
    io.kubewarden.policy.severity: medium
  name: pod-privileged-policy
spec:
  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v1.0.10
  settings: {}
  rules:
  - apiGroups:
    - ''
    apiVersions:
    - v1
    resources:
    - pods
    operations:
    - CREATE
  - apiGroups:
    - ''
    apiVersions:
    - v1
    resources:
    - replicationcontrollers
    operations:
    - CREATE
    - UPDATE
  - apiGroups:
    - apps
    apiVersions:
    - v1
    resources:
    - deployments
    - replicasets
    - statefulsets
    - daemonsets
    operations:
    - CREATE
    - UPDATE
  - apiGroups:
    - batch
    apiVersions:
    - v1
    resources:
    - jobs
    - cronjobs
    operations:
    - CREATE
    - UPDATE
  mutating: false
  allowInsideKubewardenNamespace: true
```

I expect both calls to be blocked, but policy is not blocking pods in kubewarden namespace.
```bash
~ k run privpod --image nginx --privileged -n kubewarden
pod/privpod created

~ k run privpod --image nginx --privileged 
Error from server: admission webhook "clusterwide-pod-privileged-policy.kubewarden.admission" denied the request: Privileged container is not allowed
```

jvanz:
## Description

In a recent change we added a new configuration field in cluster wide policies. This flag tells the controller not to add the namespace selector that skips evaluation of resources in the controller namespace. However, there is a second layer of security to avoid resource evaluations in some namespace configured in the policy-server. When the controller has the
"--always-accept-admission-reviews-on-deployments-namespace" CLI flag enabled, it adds the
"KUBEWARDEN_ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE" environment variable in the policy server deployment, configuring it to consider any request for that namespace (in this case the controller one) as accepted. This makes the policy configuration have no effect.

To fix this, this commit adds a new value field in the kubewarden-controller helm chart to allow cluster administrators to disable this CLI flag. Therefore, they would be able to run policies in the Kubewarden namespace.

Fix #1540

### DIFF
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -145,6 +145,12 @@ preDeleteHook:
     runAsNonRoot: true
     seccompProfile:
       type: RuntimeDefault
+# If true, the controller will always accept admission reviews in the
+# deployment namespace. It is recommended to keep this value true unless you
+# have a specific reason to disable it. This is a safety flag to avoid policy
+# evaluations that could interfere with the Kubewarden stack running in the
+# admission controller namespace.
+alwaysAcceptAdmissionReviewsOnDeploymentsNamespace: true
 # Verbosity of logging. Can be one of 'debug', 'info', 'error'.
 logLevel: info
 # open-telemetry options
--- a/charts/kubewarden-controller/templates/deployment.yaml
+++ b/charts/kubewarden-controller/templates/deployment.yaml
@@ -51,7 +51,9 @@ spec:
         - --leader-elect
         - --deployments-namespace={{ .Release.Namespace }}
         - --webhook-service-name={{ include "kubewarden-controller.fullname" . }}-webhook-service
+        {{- if .Values.alwaysAcceptAdmissionReviewsOnDeploymentsNamespace }}
         - --always-accept-admission-reviews-on-deployments-namespace
+        {{- end }}
         - --zap-log-level={{ .Values.logLevel }}
        {{- if .Values.mTLS.enable }}
         - --client-ca-configmap-name={{ .Values.mTLS.configMapName }}
--- a/charts/kubewarden-controller/tests/always_accept_admission_reviews_test.yaml
+++ b/charts/kubewarden-controller/tests/always_accept_admission_reviews_test.yaml
@@ -1,0 +1,25 @@
+suite: alwaysAcceptAdmissionReviewsOnDeploymentsNamespace flag
+templates:
+  - deployment.yaml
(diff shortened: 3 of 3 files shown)


charts/kubewarden-controller/values.yaml
```diff
@@ -145,6 +145,8 @@ preDeleteHook:
     runAsNonRoot: true
     seccompProfile:
       type: RuntimeDefault
+# If true, the controller will always accept admission reviews on the deployments namespace
```
flavio: Add a line explaining the safest choice is to leave that set to `true`.

jvanz: Done, if you want, review it again. ;)

jvanz:
> About naming, what do you think about this?
> 
> ```
> # Controller:
> alwaysAcceptAdmissionReviewsOnDeploymentsNamespace (keep as-is)
> 
> # Policy:
> spec.evaluateOnDeploymentsNamespace (for consistency with ^, or)
> spec.evaluateOnAdmissionControllerNamespace
> ```
> 
> To me `spec.allowInsideAdmissionControllerNamespace` feels safe, like it relaxes policy, reducing strictness.

I still would like to add the `allow` prefix. Just to make very clear that configuration does not automatically make the policy evaluate resources in the admission controller. But give the user the possibility to do so. But I'm fine with the `allowEvaluateOnAdmissionControllerNamespace`. I would not use the `Deployments` option. It sounds vague to me and does not help on the goal of the renaming. Which is make explicit that this is a settings related to the admission controller component of Kubewarden. 

I've opened another [PR](https://github.com/kubewarden/kubewarden-controller/pull/1544) to the renaming to separated the changes. Renaming generated a lot of changes due the generated manifest from the controller-gen. And I don't want to pollute this fix with that changes. Please, can you open this discussion there? (sorry for that) Therefore, we can update the appropriate PR if we want to. :)

viccuad (approved):
LGTM

PR #1556, opened 2026-03-05
fix(policy-server): initialize crypto provider.

jvanz:
## Description

When the policy-server starts, depending on the flag passed, the initialization code path is different. This is causing problems because when the "sigstore-trust-config" CLI flag is passed, the Kubernetes client is initialized earlier. As this client required a crypto provider ready to be used, the program crashes due the missing provider. This does not happen in other scenarios because the code path is a little bit different and some of our other dependencies initialize the crypto provider for us.

This commit fixes the issue following a similar approach that we use in kwctl where the crypto provider is initialized in the beginning of the main function.

### DIFF
--- a/crates/policy-server/src/main.rs
+++ b/crates/policy-server/src/main.rs
@@ -10,6 +10,7 @@ use clap::ArgMatches;
 use policy_server::PolicyServer;
 use policy_server::metrics::setup_metrics;
 use policy_server::tracing::setup_tracing;
+use rustls::crypto::aws_lc_rs::default_provider;
 
 #[tokio::main]
 async fn main() -> Result<()> {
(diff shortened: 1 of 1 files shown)


crates/policy-server/src/main.rs
```diff
@@ -18,6 +19,15 @@ async fn main() -> Result<()> {
...
     }
 
+    // This is necessary to ensure that we are using proper initializing the crypto provider.
+    // Without this initialization step, depending on which cli flags policy server has, the
+    // kubernetes client is initialized before the crypto provider causing a crash. In other
+    // scenarios the client is initialized after other dependencies which set the provider for us.
+    // Therefore, the following install_default() call ensure that we will have always a crypto
+    // provider initialized before any operation may require it.
+    if let Err(e) = default_provider().install_default() {
+        tracing::warn!("Failed to install rustls crypto provider: {:?}", e);
```
jvanz, line 33: Wait... I just noticed that the tracing is used before its init on line 33.

PR #1583, opened 2026-03-16
feat: foward image pull secrets

### LINKED ISSUE #1576: policy-server Deployments lack imagePullSecrets
It seems that after setting the following on the `kubewarden-controller` Helm chart values:
- `.Values.global.cattle.systemDefaultRegistry`. This takes effect for the `kubewarden-controller:v1.33.0` image, `policy-server:v1.33.0` image, `audit-scanner:v1.33.0` image.
- `.Values.imagePullSecrets`: This takes effect for the controller Deployment and the audit-scanner Deployment. But policy-server Deployments are missing this.

Note: not to confuse with the PolicyServer CRD `spec.imagePullSecrets`, which is for allowing the policy-server binary to pull policies from an authenticated registry.

### Acceptance criteria

- Fix the controller code, ensuring that policy-server Deployments get the same imagePullSecrets as the Controller Deployment or audit-scanner Deployment. For example, reuse the PolicyServer `spec.imagePullSecrets` also for the field  in the policy-server Deployment `deployment.Spec.Template.Spec.ImagePullSecrets`
- Update CRD docs for PolicyServer spec.imagePullSecrets to state that it is used for both the policy-server Deployment (that contains the `policy-server:v1.33.0` image) and for pulling policies.
- Add unit tests as needed
- Extend e2e airgap tests to test the full imagePullSecrets feature, by using an authenticated registry.

jvanz:
## Description

Adds a new controller CLI flag to allow used to define a list of secret that must be set in the policy server deployments to allow the container image to be downloaded from a private registry. Also update the Helm chart to expose this new CLI flag to users.

Fix #1576

### DIFF
--- a/api/policies/v1/policyserver_types.go
+++ b/api/policies/v1/policyserver_types.go
@@ -67,11 +67,8 @@ type PolicyServerSpec struct {
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
-	// Name of ImagePullSecret secret in the same namespace. It is mounted
-	// inside the policy-server container so that the policy-server binary can
-	// pull OCI policy artifacts from private registries at runtime. Note: this
-	// secret is not used to pull the policy-server container image itself; use
-	// the controller's --image-pull-secrets flag for that purpose.
+	// Name of ImagePullSecret secret in the same namespace. used for pulling
+	// policies from repositories.
 	// +optional
 	ImagePullSecret string `json:"imagePullSecret,omitempty"`
 
--- a/charts/kubewarden-controller/values.yaml
+++ b/charts/kubewarden-controller/values.yaml
@@ -109,9 +109,7 @@ global:
 nameOverride: ""
 # fullnameOverride completely replaces the generated release name
 fullnameOverride: ""
-# Secrets to pull container images from private registries. When set, the
-# secret names are passed to the controller via --image-pull-secrets and added
-# to the spec.template.spec.imagePullSecrets of every policy-server Deployment.
+# Secrets to pull container images from private registries
 imagePullSecrets: []
 # -- Additional labels to add to all resources
 additionalLabels: {}


api/policies/v1/policyserver_types.go
```diff
@@ -67,8 +67,12 @@ type PolicyServerSpec struct {
...
-	// Name of ImagePullSecret secret in the same namespace, used for pulling
-	// policies from repositories.
+	// Name of ImagePullSecret secret in the same namespace. It is used for two
+	// purposes: (1) as a Deployment-level imagePullSecrets entry so that
+	// Kubernetes can pull the policy-server container image from a private
+	// registry, and (2) mounted inside the policy-server container so that the
+	// policy-server binary can pull OCI policy artifacts from private
+	// registries at runtime.
 	// +optional
 	ImagePullSecret string `json:"imagePullSecret,omitempty"`
```
flavio, line 73: The name is singular, but then in the code we make use of the plural and we also merge multiple secrets into a single one (from what I've understood).

The contents of this secret is a docker config, which is a dictionary that has many entries.

Why do we need to have multiple secrets and complicate our codebase to merge them.

Can't we ask the user to provide a docker config that has all the different secrets configured:

```json
        "auths": {
                "dp.apps.rancher.io": {},
                "ghcr.io": {},
                "harbor.local": {},
                "https://index.docker.io/v1/": {},
                "kubewarden.jfrog.io": {},
                "kwbug.azurecr.io": {},
                "localhost:5000": {},
                "registry.svc.lan": {}
        }
```

viccuad: Agree, I would not merge secrets.

I would add the new controller flag with the imagePullSecret. In the Helm chart, use the same `.Values.global.imagePullSecret` as the pull secret for the controller Deployment, audit-scanner CronJob (already done) and policy-server Deployment. I think that adding a new controller flag for propagating this pull secret is fine.

charts/kubewarden-controller/values.yaml
```diff
@@ -109,7 +109,9 @@ global:
 nameOverride: ""
 # fullnameOverride completely replaces the generated release name
 fullnameOverride: ""
-# Secrets to pull container images from private registries
+# Secrets to pull container images from private registries. When set, the
+# secret names are passed to the controller via --image-pull-secrets and added
+# to the spec.template.spec.imagePullSecrets of every policy-server Deployment.
```
viccuad: I think we can remove this comment addition, it's an implementation detail, and is only partially true. This also happens for the `audit-scanner` CronJob and the controller Deployment.

api/policies/v1/policyserver_types.go
jvanz: Just to clarify one thing. We do not merge the secret in the sense of merging their content. We "merge" by adding all of the secrets (from the controller cli flag and policy server spec field) in the deployment configuration.

That's being said, I think asking user to merge all the auth info into a single secret  can be an unusual approach. I image that operator would like to use the commands like the following to create the secrets:

```
kubectl create secret docker-registry secret-tiger-docker \
  --docker-email=tiger@acme.example \
  --docker-username=tiger \
  --docker-password=pass1234 \
```

Therefore, they will not have a single secret for all the registries they want to authenticate into.

jvanz: > I would add the new controller flag with the imagePullSecret. In the Helm chart, use the same .Values.global.imagePullSecret as the pull secret for the controller Deployment, audit-scanner CronJob (already done) and policy-server Deployment. I think that adding a new controller flag for propagating this pull secret is fine.

This is already done in the PR.

jvanz: By the way, with the changes in this PR. If the user wants to use a single secret, they cay. Because now the controller will add the secret in the deployment configuration as well. ;)

jvanz: As we discussed in other communication channels, I've remove the merge function and keep only the secret defined in the controller cli flag in the policy server deployment configuration. The image pull secret from the policy server spec will be used only to download policies.

api/policies/v1/policyserver_types.go
```diff
@@ -67,8 +67,11 @@ type PolicyServerSpec struct {
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
-	// Name of ImagePullSecret secret in the same namespace, used for pulling
-	// policies from repositories.
+	// Name of ImagePullSecret secret in the same namespace. It is mounted
+	// inside the policy-server container so that the policy-server binary can
+	// pull OCI policy artifacts from private registries at runtime. Note: this
+	// secret is not used to pull the policy-server container image itself; use
+	// the controller's --image-pull-secrets flag for that purpose.
```
viccuad: I think this has too many implementation details as it ends in the PolicyServer CRDs end-user docs. I wouldn't mention the controller `--image-pull-secrets` flag, as users can't set that by hand. Nor I would mention that it is mounted inside the policy-server container, as we have PolicyServer CRs so users don't need to understand the implementation details.

I would mention that it is used for pulling policies from repositories (I wouldn't say private registries, they don't particularly need to be..), and have the note that is note used for container images if anything.

viccuad (changes requested):

PR #1595, opened 2026-03-19
fix(kwctl): normalize registry:// URIs to include explicit tag

### LINKED ISSUE #1547: Policy without tag is pulled successfully but then it cannot be found
When tag is missing
 - pulling step uses `latest` tag as default, policy is downloaded
 - scaffold step does not use `latest` as default and it fails

```bash
~ kwctl scaffold manifest -t ClusterAdmissionPolicy registry://ghcr.io/kubewarden/policies/pod-privileged
2026-03-04T08:33:22.275480Z  INFO kwctl: cannot find policy with uri: registry://ghcr.io/kubewarden/policies/pod-privileged, trying to pull it from remote registry
  Successfully pulled policy from registry://ghcr.io/kubewarden/policies/pod-privileged
Error: Cannot find policy with uri: registry://ghcr.io/kubewarden/policies/pod-privileged

~ kwctl --version
kwctl 1.32.0
```

It works fine when I use specific tag `registry://ghcr.io/kubewarden/policies/pod-privileged:latest`

jvanz:
## Description

When a policy URI without a tag is used (e.g. registry://ghcr.io/kubewarden/policies/pod-privileged), the pull step stores the policy under a path with ':latest' appended, but subsequent commands (scaffold, inspect, rm) looked up the URI without the tag, causing them to fail with 'Cannot find policy with uri'.

Fix by normalizing registry:// URIs in map_path_to_uri via Reference::from_str, which defaults to ':latest' when no tag is specified. This makes the lookup key consistent with the store path.

Fix #1547 


## Test

```console
kwctl scaffold manifest -t ClusterAdmissionPolicy registry://ghcr.io/kubewarden/policies/pod-privileged                                                                    
2026-03-19T19:10:29.018704Z  INFO kwctl: cannot find policy with uri: registry://ghcr.io/kubewarden/policies/pod-privileged:latest, trying to pull it from remote registry
  Successfully pulled policy from registry://ghcr.io/kubewarden/policies/pod-privileged:latest                         
kind: ClusterAdmissionPolicy                                                                                                                                                                                                                  
metadata:
  annotations:
    io.kubewarden.policy.category: PSP
    io.kubewarden.policy.severity: medium
  name: pod-privileged-policy
spec:
  module: registry://ghcr.io/kubewarden/policies/pod-privileged:latest
  settings: {}
  rules:
  - apiGroups:
    - ''
    apiVersions:
    - v1
    resources:
    - pods
    operations:
    - CREATE
    - UPDATE
  - apiGroups:
    - ''
    apiVersions:
    - v1
    resources:
    - replicationcontrollers
    operations:
    - CREATE
    - UPDATE
  - apiGroups:
    - apps
    apiVersions:
    - v1
    resources:
    - deployments
    - replicasets
    - statefulsets
    - daemonsets
    operations:
    - CREATE
    - UPDATE
  - apiGroups:
    - batch
  apiVersions:
    - v1
    resources:
    - jobs
    - cronjobs
    operations:
    - CREATE
    - UPDATE
  mutating: false
```

### DIFF
--- a/crates/kwctl/src/main.rs
+++ b/crates/kwctl/src/main.rs
@@ -22,7 +22,7 @@ use crate::{
     },
     load::load,
     save::save,
-    utils::{LookupError, find_file_matching_file},
+    utils::{LookupError, find_file_matching_file, normalize_uri},
 };
 
 mod annotate;
@@ -119,7 +119,8 @@ async fn main() -> Result<()> {
         Some("info") => info::info(),
         Some("pull") => {
             if let Some(matches) = matches.subcommand_matches("pull") {
-                let uri = matches.get_one::<String>("uri").unwrap();
+                let uri = normalize_uri(matches.get_one::<String>("uri").unwrap());
+                let uri = &uri;
                 let destination = matches
                     .get_one::<String>("output-path")
                     .map(|output| PathBuf::from_str(output).unwrap());
@@ -133,7 +134,8 @@ async fn main() -> Result<()> {
         }
         Some("verify") => {
             if let Some(matches) = matches.subcommand_matches("verify") {
-                let uri = matches.get_one::<String>("uri").unwrap();
+                let uri = normalize_uri(matches.get_one::<String>("uri").unwrap());
+                let uri = uri.as_str();
                 let sources = remote_server_options(matches)?;
                 let verification_options = build_verification_options(matches)?
                     .ok_or_else(|| anyhow!("could not retrieve sigstore options"))?;
@@ -193,8 +195,9 @@ async fn main() -> Result<()> {
         }
         Some("rm") => {
             if let Some(matches) = matches.subcommand_matches("rm") {
-                let uri_or_sha_prefix = matches.get_one::<String>("uri_or_sha_prefix").unwrap();
-                rm::rm(uri_or_sha_prefix)?;
+                let uri_or_sha_prefix =
+                    normalize_uri(matches.get_one::<String>("uri_or_sha_prefix").unwrap());
+                rm::rm(&uri_or_sha_prefix)?;
             }
             Ok(())
         }
@@ -233,7 +236,8 @@ async fn main() -> Result<()> {
         }
         Some("inspect") => {
             if let Some(matches) = matches.subcommand_matches("inspect") {
-                let uri_or_sha_prefix = matches.get_one::<String>("uri_or_sha_prefix").unwrap();
+                let uri_or_sha_prefix =
+                    normalize_uri(matches.get_one::<String>("uri_or_sha_prefix").unwrap());
                 let output = inspect::OutputType::try_from(
                     matches.get_one::<String>("output").map(|s| s.as_str()),
                 )?;
@@ -242,7 +246,7 @@ async fn main() -> Result<()> {
                     .get_one::<bool>("show-signatures")
                     .unwrap_or(&false)
                     .to_owned();
-                inspect::inspect(uri_or_sha_prefix, output, sources, no_color, no_signatures)
+                inspect::inspect(&uri_or_sha_prefix, output, sources, no_color, no_signatures)
                     .await?;
             };
             Ok(())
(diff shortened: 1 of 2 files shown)


crates/kwctl/src/utils.rs
```diff
@@ -27,6 +27,14 @@ pub(crate) enum LookupError {
 pub(crate) fn map_path_to_uri(uri_or_sha_prefix: &str) -> std::result::Result<String, LookupError> {
     let uri_has_schema = Regex::new(r"^\w+://").unwrap();
     if uri_has_schema.is_match(uri_or_sha_prefix) {
+        // Normalize registry:// URIs so that a missing tag defaults to `:latest`.
+        // This ensures the URI matches the path used by the store, which always
+        // stores policies with an explicit tag.
+        if let Some(image) = uri_or_sha_prefix.strip_prefix("registry://")
+            && let Ok(reference) = Reference::from_str(image)
```
viccuad: Ok, this was a bit unclear for me. The normalization happens on the `Reference::from_str()` call, as it does a try_from(), and does a:
```rust
if tag.is_none() && digest.is_none() {
  tag = Some(DEFAULT_TAG.into());
}
```

I've seen also that we do normalization in policy-fetcher: https://github.com/kubewarden/kubewarden-controller/blob/main/crates/policy-fetcher/src/lib.rs#L114-L117


Are we sure we always use `map_path_to_uri()` for each time that we interact with the local store?

I wonder if we can try to normalize the uri once per kwctl subcommand, when we first obtain or parse the uri or reference. We could add a `normalize_uri(uri)` function that gets called the first time or only time we pass the uri downwards in each subcommand (pull, push, pull-and-run, etc).

jvanz: The function from policy-fetcher is being called and it's properly fixing the tag. You can see that when you see the issue description issue.The pulling of the policies is running. However, the manifest command later try to find the policy. But it does not fix the tag. That's why it cannot find the policy later. 

```
kwctl scaffold manifest -t ClusterAdmissionPolicy registry://ghcr.io/kubewarden/policies/pod-privileged                                                               
2026-03-23T13:07:23.920603Z  INFO kwctl: cannot find policy with uri: registry://ghcr.io/kubewarden/policies/pod-privileged, trying to pull it from remote registry
  Successfully pulled policy from registry://ghcr.io/kubewarden/policies/pod-privileged 
Error: Cannot find policy with uri: registry://ghcr.io/kubewarden/policies/pod-privileged
```

So, yes, we can think of an normalization function to be called before any operation. Taking a look in the code, this function is the closest one that we have that is similar to what you proposed as `normalize_uri`. Because it is called several places to get the URI. Including the function `get_uri` in the same file. I'm not saying that it's called everywhere. But I can use it to fix other places to do the same thing in other places.

By the way, the `get_uri` is called by `get_wasm_path` which is called by `pull_if_needed` that later call the `fetch_policy`.

viccuad: I see, thanks. Can we ensure the normalization always happens by dealing with References in these cases, as they are normalized, just like in this proposed change? that may be a simple solution.

jvanz: I agree the we could improve the normalization. But to properly do that we should go a little bit beyond. The `References` object is used only for "registry" policy types. I think if we decide to do that we should create an enum to represent any kind of policy references (http, file and OCI). Therefore, we can pass this information everywhere and abstract the logic of normalization and so on. Which also includes a refactoring of all the API from policy-evaluator and policy-fetcher. I'm in favor of it. It looks the like good solution for the long run. But I would like to avoid doing this on this PR. Looks like we are doing two tasks: fixing the bug and refactoring the code base in a single PR. Where the goal is to fix a bug only. 

Therefore, I propose the following, I'll create an issue to fix this tech-bebt and for this PR, I'll update the kwctl code to normalize the reference string in the commands. 

How that sounds?

jvanz: I've updated the code to give an example of normalization of the string in the sub commands. Take a look. I did not review all the sub commands yet. But I would like to share an idea that can impact if we will do that or not... 

While I was doing that I though that is not required. In the end of the day, the previous code already fix the function called by the sub commands. Therefore, we could just ending by adding a second normalization check. I think we could remove the code from the sub command calling the normalization function and leave the proper fix where we replace the string by some type representing the all the possibles policies URI to a future tech-debt issue. As we are discussing in the previous comments

crates/kwctl/src/main.rs
flavio, line 1: nit: remove some repetition by doing a `use crate::utils::normalize_uri`

crates/kwctl/src/utils.rs
```diff
@@ -105,6 +119,40 @@ mod tests {
...
+        assert_eq!(
+            normalize_uri("file:///some/path/policy.wasm"),
+            "file:///some/path/policy.wasm",
+        );
+        assert_eq!(
+            normalize_uri("https://example.com/policy.wasm"),
+            "https://example.com/policy.wasm",
+        );
+    }
+
```
flavio: nit, this could be DRY-ed with rstest

PR #1631, opened 2026-04-02
ci go improvements

flavio:
This PR does some improvements to some GHA that deal with Go.

We now install the same version of go that is mentioned inside of the `go.mod` file. Moreover, we intruct govulncheck to look also at the `go.mod`.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -3,6 +3,7 @@ on:
   push:
     paths:
       - go.sum
+      - go.mod
   schedule:
     - cron: "0 0 * * *"
 
@@ -17,7 +18,7 @@ jobs:
       - name: Install Golang
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: "1.26"
+          go-version-file: "go.mod"
           check-latest: true # Always check for the latest patch release
       - name: govulncheck
         uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -20,12 +20,11 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
-      - name: Install Golang
-        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
-        with:
-          go-version-file: "go.mod"
-          check-latest: true # Always check for the latest patch release
       - name: govulncheck
         uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
         with:
+          # required to override the default value that is 'stable'
+          # in this way `go-version-file` is going to be taken into account
+          go-version-input: ""
+          go-version-file: "go.mod"
           go-package: ./...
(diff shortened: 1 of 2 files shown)


viccuad (approved):
Oh, this is nice. We should do the same for the policies repos.

I'm ok with removing `check-latest: true`, I prefer to pin the version from go.mod.

jvanz (approved):

.github/workflows/govulncheck.yml
```diff
@@ -17,7 +18,7 @@ jobs:
       - name: Install Golang
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: "1.26"
+          go-version-file: "go.mod"
```
jvanz: We do not need this step. govulncheck-action already install [go](https://github.com/golang/govulncheck-action/blob/31f7c5463448f83528bd771c2d978d940080c9fd/action.yml#L48-L65). You can define the `go.mod` file in its inputs as well

Considering the action has the `check-latest` as [`true`](https://github.com/golang/govulncheck-action/blob/b625fbe08f3bccbe446d94fbf87fcc875a4f50ee/action.yml#L8-L11) by default. I suspect that means the go lang version will be update under the hood. That's what copilot is warning us. Therefore, I would say to remove the setup-go step here and move this configuration to the `govulncheck-action` instead.

jvanz (changes requested):
I think we need to do a little change. See my comment in the `govulncheck.yml` file

.github/workflows/govulncheck.yml
flavio: I think I need to remove the `chack-latest`, then the govulncheck is going to use the version cached by this step (hence the one we have inside of `go.mod`:

> The precedence for inputs go-version-input, go-version-file, check-latest, cache, and cache-dependency-path specifying Go version and caches is inherited from [actions/setup-go](https://github.com/actions/setup-go)

From the README of the project.

What do you think?

jvanz: Ah, I just saw the `check-latest` in this `setup-go` action. But I think you cannot remove the input. Actually, we need to keep the go version configuration in sync with the govulncheck input values. Otherwise, you will install a go version here and govulncheck may update the version before run.

flavio:
@jvanz ready for review again

.github/workflows/govulncheck.yml
```diff
@@ -14,12 +15,16 @@ jobs:
...
-          check-latest: true # Always check for the latest patch release
+          fetch-depth: 0
+          persist-credentials: false
       - name: govulncheck
         uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
         with:
+          # required to override the default value that is 'stable'
+          # in this way `go-version-file` is going to be taken into account
+          go-version-input: ""
+          go-version-file: "go.mod"
```
jvanz, line 29: I would include the input to disable patch version updates

https://github.com/golang/govulncheck-action/blob/31f7c5463448f83528bd771c2d978d940080c9fd/action.yml#L8-L11

jvanz (changes requested):
Thanks for the changes! But I think we need to consider the copilot comments. Just to ensure that we are run the same go version in all CI jobs.  Maybe we can follow the suggestion from its first [comment](https://github.com/kubewarden/kubewarden-controller/pull/1631/files#r3027277889) and define the version including the patch version.

PR #1662, opened 2026-04-14
fix(ci): install kwctl version from Github Action.

jvanz:
## Description

Updates the e2e-tests.yml to install the default version from the kwctl-installer Github action. We cannot use "latest" as version now because it's necessary to verify the signature of binary using cosign and that require the actual version that trigger the release CI job.

Related to https://github.com/kubewarden/github-actions/pull/290


kravciak:
We need -rc versions in tests, I will need to find different way if we remove this.
If you want to merge please create tracking card (assign me) to switch back to `RCs` later.

viccuad (approved):
I'm Ok merging this for it to work for now, but my vote definitely goes for scheduling the work on kwctl-installer action to support `latest` (see the other PR).

PR #1682, opened 2026-04-23
feat: add hostNetwork support

### LINKED ISSUE #1597: Allow users to enable host network in Kubewarden resources
Recently we received a [request](https://github.com/kubewarden/kubewarden-controller/issues/1658) from users to allow them to run Kubewarden resources with host network enabled. After some deliberation and looking for another solutions for the user issue, we've agreed on that the only solution that we, the Kubewarden team, can do is to give users the power to enable host network. All other possible solutions for this issue is out of control of the Kubewarden project and involves changes in users infrastructure. 

As this configuration can increase the attack surface, this feature should be disable by default. And it should be well documented explaining the risks for users.

### Acceptance criteria
- Update the admission controller reconciliation loop to configure policy server deployments enabling host network when requested by user
-  Update Helm chart to allow users to enable host network when deploying admission controller
- Add tests to cover this new feature

jvanz:
## Description

This PR adds host-network support to Kubewarden. When the `--host-network` controller flag is enabled, all PolicyServer deployments are configured with `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet`. Three new optional fields are added to the PolicyServer CRD spec to allow users to customize ports on a per-PolicyServer basis: `webhookPort` changes the port the pod listens on for webhook requests, `readinessProbePort` changes the readiness probe port, and both help resolve host-port conflicts when multiple PolicyServers share a node. The `metricsPort` field customizes only the Service port (the externally visible port Prometheus scrapes). The pod-side TargetPort remains the controller-wide default because the OpenTelemetry sidecar always exports metrics on a fixed cluster-wide port. As a result, `metricsPort` does **not** resolve host-network port collisions on the node

Affinity is fully user-managed. The controller does not auto-inject pod anti-affinity rules, giving operators full control over scheduling. OpenTelemetry sidecar mode is incompatible with host-network mode: when `hostNetwork` is enabled, sidecar injection is rejected at both the helm chart level (template render fails) and the controller binary level (fail-fast error at startup), so users must use a remote collector (`telemetry.mode=custom`) instead. Both the `--webhook-server-port` flag and the `KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT` environment variable are validated at startup to be within the 1–65535 range, failing fast with clear error messages if out of bounds.

Fix https://github.com/kubewarden/kubewarden-controller/issues/1597


charts/kubewarden-controller/templates/service.yaml
```diff
@@ -10,16 +11,12 @@ metadata:
...
     port: 8080
-    targetPort: 8080
-  {{- end}}
-  - name: https
-    port: 8443
-    targetPort: https
+    targetPort: {{ .Values.ports.metrics }}
   selector:
     {{- include "kubewarden-controller.selectorLabels" . | nindent 4 }}
+{{- end}}
```
jvanz, line 19: @kubewarden/kubewarden-developers do you know we add this service with the webhook port if it should be used for metric collection?

@copilot can you investigate the git story and let us know what you found?

api/policies/v1/policyserver_webhook.go
```diff
@@ -208,3 +216,39 @@ func validateLimitsAndRequests(limits, requests corev1.ResourceList) field.Error
...
+			field.NewPath("spec").Child("metricsPort"),
+			metricsPort,
+			fmt.Sprintf("metricsPort must differ from webhookPort (%d)", webhookPort),
+		))
+	}
+	if readinessPort == metricsPort {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("metricsPort"),
+			metricsPort,
+			fmt.Sprintf("metricsPort must differ from readinessProbePort (%d)", readinessPort),
```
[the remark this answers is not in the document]
flavio: I prefer the actual code

api/policies/v1/policyserver_webhook.go
```diff
@@ -327,3 +334,39 @@ func validateSingleCapability(pattern string, path *field.Path) *field.Error {
...
+		))
+	}
+	if webhookPort == metricsPort {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("metricsPort"),
+			metricsPort,
+			fmt.Sprintf("metricsPort must differ from webhookPort (%d)", webhookPort),
+		))
+	}
+	if readinessPort == metricsPort {
```
flavio: You also have to check for readiness != webhook

api/policies/v1/policyserver_webhook_test.go
```diff
@@ -412,3 +412,104 @@ func TestPolicyServerValidateNamespacedPoliciesCapabilities(t *testing.T) {
...
+			defaultMetricsPort: constants.PolicyServerMetricsPort,
+			errContains:        "metricsPort must differ from webhookPort",
+		},
+		{
+			name:               "readinessProbePort equals metricsPort",
+			readiness:          ptr.To[int32](9000),
+			metrics:            ptr.To[int32](9000),
+			defaultMetricsPort: constants.PolicyServerMetricsPort,
+			errContains:        "metricsPort must differ from readinessProbePort",
+		},
```
flavio, line 447: test also readiness == webhook

charts/kubewarden-controller/templates/_helpers.tpl
```diff
@@ -181,3 +181,20 @@ are configured.
...
 {{- end -}}
+
+{{/*
+Compute the effective affinity for the controller deployment.
+Uses the controller-specific affinity if set, otherwise falls back to
+global.affinity for backward compatibility.
+
+NOTE: When hostNetwork is enabled, users are responsible for setting
+appropriate podAntiAffinity rules to prevent host-port conflicts between
+controller replicas on the same node.
```
flavio, line 192: I haven't finished the review yet, this information must be made explicit also inside of docs and inside of values.yml (or any other place that is more visible to end users)

charts/kubewarden-controller/values.yaml
```diff
@@ -151,6 +151,39 @@ preDeleteHook:
...
+# PolicyServer pods managed by this controller.
+# WARNING: enabling this increases the attack surface by exposing webhook
+# endpoints on the host network and giving pods visibility of all node network
+# interfaces. Use only when the Kubernetes API server cannot reach pod-network
+# webhook endpoints (e.g. clusters using a non-VPC CNI with NAT).
+# NOTE: hostNetwork mode is incompatible with telemetry sidecar mode
+# (telemetry.mode=sidecar). When hostNetwork is enabled, sidecar injection
+# is automatically disabled to avoid port conflicts. Use telemetry.mode=custom
+# with a remote collector instead.
+hostNetwork: false
```
flavio, line 177: Maybe also add another NOTE that points to the affinity stuff you wrote above. I think it could be easy for an user to miss that.

flavio (commented):
Overall LGTM, left some comments

api/policies/v1/policyserver_webhook.go
jvanz: This is done in the first `if` on this function, on line 349. ;)

api/policies/v1/policyserver_webhook_test.go
jvanz: This is done in the second test case: `webhookPort equals readinessProbePort`. Line 431. ;)

jvanz:
@flavio fyi: since your first review I've added some documentation explaining the different behavior of the `spec.metricsPort` in the PolicyServer CRDs. This also means that I've removed some check in the validation webhook for the CRD that are not necessary.

charts/kubewarden-controller/values.yaml
```diff
@@ -151,6 +151,42 @@ preDeleteHook:
...
+# appropriate podAntiAffinity rules to prevent host-port conflicts between
+# controller replicas on the same node.
+# NOTE: hostNetwork mode is incompatible with telemetry sidecar mode
+# (telemetry.mode=sidecar). This combination is rejected; sidecar injection
+# is not automatically disabled. Use telemetry.mode=custom with a remote
+# collector instead.
+hostNetwork: false
+# ports configures the network ports used by the controller pod.
+# These are especially important when hostNetwork is enabled, because each port
+# is bound on the host network interface and must not clash with other workloads.
```
flavio: That applies only to the webhook and the healthProbe, not to the metrics since its bound to the Service. However, I think it's fine to "scare" the end user

charts/kubewarden-defaults/values.yaml
```diff
@@ -196,6 +196,32 @@ policyServer:
...
+  # Port configuration for the policy server.
+  # These ports are used for the webhook, readiness probe, and metrics endpoints.
+  # Important when hostNetwork is enabled or when running multiple PolicyServers
+  # on the same node, as they must not conflict.
+  #
+  # Default values (if not specified):
+  # - webhookPort: 8443
+  # - readinessProbePort: 8081
+  # - metricsPort: controller-wide default (env var
+  #   KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT, falling back to 8080)
```
flavio: I would remove the env var comment, it's not relevant

charts/kubewarden-defaults/values.yaml
```diff
@@ -196,6 +196,32 @@ policyServer:
...
+  #
+  # webhookPort: 9443
+  # readinessProbePort: 9081
+  # metricsPort: 9080
+  # affinity configures affinity rules for the default PolicyServer.
```
flavio, line 213: nit: introduce an empty line between the `metricsPort` and this line to indicate a change of topic

cmd/controller/main.go
```diff
@@ -135,6 +148,13 @@ func main() {
 		"image-pull-secrets",
 		"",
 		"Comma-separated list of Secret names to use as imagePullSecrets on every policy-server Deployment. The secrets must exist in the deployments namespace.")
+	flag.BoolVar(&config.HostNetwork,
+		"host-network",
+		false,
+		"Enable host network mode for all PolicyServer deployments. "+
+			"WARNING: enabling this increases the attack surface by exposing webhook endpoints "+
+			"on the host network and giving pods visibility of all node network interfaces. "+
+			"Use only when the Kubernetes API server cannot reach pod-network webhook endpoints.")
```
flavio, line 157: I would also emit a log entry with this warning when host network is enabled.

flavio (commented):
thanks for the changes, it looks good

charts/kubewarden-controller/templates/NOTES.txt
```diff
@@ -11,3 +11,11 @@ If you'd like to support us, we'd love to hear from you as one of our adopters.
 Adopters can be public or private. 
 
 Learn how to add your organization as a Kubewarden adopter by checking out the ADOPTERS.md file here: https://github.com/kubewarden/community/blob/main/ADOPTERS.md
+
+{{ if .Values.hostNetwork }}
+⚠️  WARNING  ⚠️
+Host Network is enabled. Ensure you set appropriate podAntiAffinity rules to prevent host port conflicts between controller replicas on the same node.
```
viccuad: Let's wrap these lines to 80~100 chars.

charts/kubewarden-controller/templates/NOTES.txt
```diff
@@ -11,3 +11,11 @@ If you'd like to support us, we'd love to hear from you as one of our adopters.
...
+{{ if eq .Values.telemetry.mode "sidecar" }}
+Telemetry sidecar mode (telemetry.mode=sidecar) is incompatible with host network. This chart rejects that configuration and rendering/installation will fail when hostNetwork=true and telemetry.mode=sidecar. Use telemetry.mode=custom with a remote collector instead, or disable hostNetwork.
```
viccuad: This block will never print, no? We will fail and print the message in helpers.tpl instead.

Edit: it prints if one enables telemetry halfway, by not enabling `telemetry.metrics` or `telemetry.tracing`. Is that useful? Should we leave it in? The users are never going to be able to enable telemetry anyways, they will `fail`.

internal/controller/policyserver_controller_deployment.go
```diff
@@ -201,7 +202,9 @@ func (r *PolicyServerReconciler) adaptDeploymentForMetricsAndTracingConfiguratio
 	// If the otel sidecar is enabled, we need to inject the sidecar in the
 	// policy server deployment. The exporter will communicate with the sidecar
 	// using the localhost address.
-	if (r.MetricsEnabled || r.TracingEnabled) && r.OtelSidecarEnabled {
+	// NOTE: OTel sidecar injection is skipped when HostNetwork is enabled because
+	// multiple sidecars on the same node would cause port conflicts.
+	if (r.MetricsEnabled || r.TracingEnabled) && r.OtelSidecarEnabled && !r.HostNetwork {
```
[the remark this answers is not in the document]
viccuad: I agree with this. 
I would prefer to fail fast on main.go when we read flags and env vars.

I'm ok with skipping the otel sidecar here, but in that case, we should at least emit an error log (and add a comment that the branch should never be reachable).

internal/controller/policyserver_controller_deployment.go
```diff
@@ -427,6 +431,54 @@ func buildPolicyServerDeploymentSpec(
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	podSpec.HostNetwork = hostNetwork
+	podSpec.DNSPolicy = corev1.DNSClusterFirst
+	if hostNetwork {
```
viccuad, line 531: could we add `podSpec.HostNetwork = hostNetwork` inside the `if`, and not set `podSpec.DNSPolicy` outside? That leaves us a smalled podSpec when HostNetwork is disabled, which I think is better UX.

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -19,7 +20,8 @@ spec:
         {{ . | quote }}: {{ get $.Values.podAnnotations . | quote}}
         {{- end }}
         {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
-        {{- if eq .Values.telemetry.mode "sidecar" }}
+        {{- /* OTel sidecar annotation is incompatible with hostNetwork due to port conflicts */ -}}
+        {{- if and (eq .Values.telemetry.mode "sidecar") (not .Values.hostNetwork) }}
```
viccuad: I prefer to remove this change, I would either helm `fail` (we do) or later fail on the controller, but not silently continue without enabling telemetry sidecar when we are saying we are enabling it.

charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -62,8 +77,12 @@ spec:
        {{- if $imagePullSecretNames }}
         - --image-pull-secrets={{ $imagePullSecretNames }}
        {{- end }}
+       {{- if .Values.hostNetwork }}
+        - --host-network
+       {{- end }}
        {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
-       {{- if eq .Values.telemetry.mode "sidecar" }}
+       {{- /* OTel sidecar is incompatible with hostNetwork due to port conflicts */ -}}
+       {{- if and (eq .Values.telemetry.mode "sidecar") (not .Values.hostNetwork) }}
```
viccuad: I would also remove this if change, see previous comment.

charts/kubewarden-controller/templates/NOTES.txt
viccuad: The more I think about it, the more I would like this comment removed. Let's only helm `fail`.

charts/kubewarden-controller/templates/service.yaml
viccuad: Quite cool!

charts/kubewarden-controller/values.schema.json
```diff
@@ -234,6 +240,20 @@
                 }
             }
         },
+        "ports": {
+            "type": "object",
+            "properties": {
+                "healthProbe": {
+                    "type": "integer"
```
viccuad: We could check for bounds with `minimum` and `maximum` like in the other schema.

charts/kubewarden-controller/values.yaml
```diff
@@ -151,6 +151,42 @@ preDeleteHook:
...
+# WARNING: enabling this increases the attack surface by exposing webhook
+# endpoints on the host network and giving pods visibility of all node network
+# interfaces. Use only when the Kubernetes API server cannot reach pod-network
+# webhook endpoints (e.g. clusters using a non-VPC CNI with NAT).
+# WARNING: When hostNetwork is enabled, users are responsible for setting
+# appropriate podAntiAffinity rules to prevent host-port conflicts between
+# controller replicas on the same node.
+# NOTE: hostNetwork mode is incompatible with telemetry sidecar mode
+# (telemetry.mode=sidecar). This combination is rejected; sidecar injection
+# is not automatically disabled. Use telemetry.mode=custom with a remote
```
viccuad: I would simplify the note and end it at "This combination is rejected". If we say we reject the combination, why say that it isn't automatically disabled. Plus they will find out when helm errors.

internal/controller/suite_test.go
```diff
@@ -66,8 +68,25 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	var ctx context.Context
 	ctx, cancel := context.WithCancel(context.TODO())
 
+	// Copy Kubewarden CRD files to a temp directory so envtest can load
+	// them. This is necessary to ensure that the latest CRDs version are available
+	// in the tests.
+	crdSourceDir := filepath.Join("..", "..", "charts", "kubewarden-crds", "templates")
```
viccuad: This is fine, but this is yet another example of just bundling all copilot changes into 1 big commit.. I dislike PRs like this.

internal/controller/suite_test.go
```diff
@@ -122,11 +144,15 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-		ClientCAConfigMapName: clientCAConfigMapName,
-		ImagePullSecrets:      []corev1.LocalObjectReference{{Name: reconcilerImagePullSecret}},
+		Client:                  k8sManager.GetClient(),
+		Scheme:                  k8sManager.GetScheme(),
+		DeploymentsNamespace:    deploymentsNamespace,
+		ClientCAConfigMapName:   clientCAConfigMapName,
+		ImagePullSecrets:        []corev1.LocalObjectReference{{Name: reconcilerImagePullSecret}},
+		PolicyServerMetricsPort: constants.PolicyServerMetricsPort,
+		TelemetryConfiguration: TelemetryConfiguration{
+			MetricsEnabled: true,
```
viccuad, line 136: does this mean we are enabling metrics for all tests? I would prefer to only enable metrics on needed tests.

charts/kubewarden-controller/values.yaml
```diff
@@ -151,6 +151,42 @@ preDeleteHook:
...
+# collector instead.
+hostNetwork: false
+# ports configures the network ports used by the controller pod.
+# These are especially important when hostNetwork is enabled, because each port
+# is bound on the host network interface and must not clash with other workloads.
+ports:
```
viccuad, line 182: I feel we expose a bit too much to an end user that is not going to exercise the feature,which creates noise and cognitive load. Should normal users care about ports? If not, could we co-locate the ports configuration by the hostNetwork value, or have them commented out?

viccuad (commented):
All in all, looks good to me. Some minor changes and a couple of questions.

Still, for the amount of changes (affinity, hostNetwork, ports), I feel this PR has a high cognitive load. It would have benefited from more atomic commits.

jvanz:
> Still, for the amount of changes (affinity, hostNetwork, ports), I feel this PR has a high cognitive load. It would have benefited from more atomic commits.

I could not review the comments yet. I'll work on that asap. But for now, I've split the commit into 5 commits. ;)

internal/controller/suite_test.go
jvanz: This does not mean too much to the other tests. This flag controls which env vars and which ports will be added to the policy server deployments and services. This is exactly what is under [test](https://github.com/kubewarden/adm-controller/pull/1682/changes/044cace1a50ebc51086a4bc3f6dafef891cfe8b7#diff-ff4e403fd738277807098b953d4991bd2b841faabb8f7aad07df11f5e176978f). No other test update is required as you can see in the changes. 

> I would prefer to only enable metrics on needed tests.

In the current implementation, there is no way to do that without a refactoring in the tests. This flag is set in the reconciler and this is done in the test suite setup code. Therefore, for that, we would need to setup another suite for a single test. I think this is not worth in for this test case.

internal/controller/policyserver_controller_deployment.go
jvanz: I've moved the two lines (476 and 477) to the podspec above. This is required to ensure that the podspec is updated when the user enable/disable the host network feature.

charts/kubewarden-controller/values.yaml
jvanz: The ordinary users will not be affected by theses new fields. The default values user are the same of the ones user before this changes. Therefore, there is nothing that they need to do in the upgrade. This is more a cosmetic change. What I like in this approach is that the Helm chart is self-documented with no actual change. But if you really believe this is necessary, I can revisit this.

jvanz: Maybe I would suggest to flatten this fields.

viccuad:
>  I've split the commit into 5 commits. ;)

Thanks! :heart: . Sorry for the additional work there, I should have be clearer that this was general feedback but I wasn't requesting the PR to be redone.

charts/kubewarden-controller/values.yaml
flavio: I don't mind about the current structure. Maybe we can reword the comment above the whole `ports` section to:

```
Only modify this section if hostNetwork is enabled. When `hostNetwork`
is enabled, each port is bound on the host network interface and must not clash with other workloads.
```

Also, I would put an empty line between line 177 and 178

flavio (approved):

internal/controller/suite_test.go
viccuad: thanks for the explanation, resolving!

viccuad (approved):
LGTM! The Helm chart e2e tests fail, I wonder if we need to rebase on top of main.

jvanz:
> LGTM! The Helm chart e2e tests fail, I wonder if we need to rebase on top of main.

The e2e tests are failing because the helm chart from this PR requires the controller container image with the changes in this PR. And the e2e test used the latest version from the main branch. Therefore, the e2e will never be green until we merge the PR. 

Therefore, I'm merging this. ;)

PR #1683, opened 2026-04-24
fix: allow usage of Go WASI policies built by go 1.26

flavio:
These changes are required to run policies built by the official go compiler, starting from release 1.26.0

BTW: I did a sync of the default features that wasmtime is using nowadays.


flavio:
This is required to fix the policies that are not passing these tests: https://github.com/kubewarden/policies/pull/392

viccuad (approved):

PR #1705, opened 2026-04-30
refactor: Drop `config/` folder, save templates to charts, and move to `examples/`

### LINKED ISSUE #1704: investigate removal of the `config` directory. Probably the only relevant stuff is under config/samples; we could move that to `example`

viccuad:
## Description

Fix https://github.com/kubewarden/adm-controller/issues/1704
Recommend review per commit.

- Consume always CRD templates from Helm chart templates.
- Generate webhook manifest directly into
`charts/kubewarden-controller/templates/webhooks.yaml` instead of
`config/webhook/manifest.yml`, overwriting it. Given that we overwrite the Helm templated file, added a helm unittest to ensure that we never merge a broken `webhooks.yaml` manifest.
- Remove uneeded files.
- Move `config/samples/` to `examples/` and add 2 more. 

## Test

Tested locally with CI. Add a Helm unittest.

## Checklist


viccuad (commented):
this approach with the webhooks manifest is incorrect, and will mean that the generated code CI check will always fail

Makefile
```diff
@@ -159,8 +159,11 @@ manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefin
 	$(GO_BUILD_ENV) $(CONTROLLER_GEN) rbac:roleName=kubewarden-controller-manager,fileName=controller-rbac-roles.yaml crd webhook \
 			paths="./api/policies/v1" paths="./api/policies/v1alpha2" \
 			paths="./internal/controller" paths="./cmd/controller" \
-			output:crd:artifacts:config=charts/kubewarden-crds/templates \
-			output:rbac:artifacts:config=charts/kubewarden-controller/templates
+			output:crd:artifacts:config=charts/kubewarden-crds/templates/crds \
+			output:rbac:artifacts:config=charts/kubewarden-controller/templates \
+			output:webhook:artifacts:config=charts
+	echo "# to be merged manually into kubewarden-controller/templates/webhooks.yaml" | cat - charts/manifests.yaml > temp && mv temp charts/manifests.yaml
+	mv charts/manifests.yaml charts/generated-webhooks-manifests.yaml
```
jvanz, line 166: If this file must be manually merged, cannot we just not commit it in the repository? If it never used in the helm charts we can just drop a warning here say "hey! Remember to merge the `generated-webhooks-manifest.yaml` with the `webhooks.yaml` in the Helm chart"

charts/kubewarden-crds/tests/crds_test.yaml
```diff
@@ -1,7 +1,7 @@
 suite: Kubewarden CRDs
 tests:
   - it: "admissionpolicies CRD should be a CustomResourceDefinition"
-    template: policies.kubewarden.io_admissionpolicies.yaml
+    template: crds/policies.kubewarden.io_admissionpolicies.yaml
```
jvanz, line 4: Nothing against of moving the CRDs under a sub directory. We will probably will do the same in the planned single helm chart changes.  But I just want to know the reason. Is it just a matter of organization preference or is there other reason?

jvanz (changes requested):
Overall, LGTM. But I have some questions.

charts/kubewarden-crds/tests/crds_test.yaml
viccuad: The controller testsuite needs a folder that contains *only* yaml or
json files, or it errors when importing them. The previous templates/ folder contained the NOTES.txt, so
it would fail.

(it's on the commit message for future reference)

Makefile
viccuad: It's already included in history, see [here](https://github.com/kubewarden/adm-controller/pull/1705/changes#diff-9eed8e9d0218e3c2954dd740cdb147d5527635cb8188030f10cd1b541c8e913dL1). 
With the change, it moves from `config/webhook/manifests.yaml` to `charts/generated-webhooks-manifest.yml` (should show on the commits, but I think I mangled the commit and I just added the new file and then removed the old),

I'm ok adding a warning, but it will be printed every time we run `make manifests`. I thought that when committing changes, and in PRs, we will see the bump to the generated file and wonder, given that the file has that first line "to be merged manually". Would that be enough or would you like me to add an echo with a reminder for `make manifests`?

viccuad:
Rebased on top of main to consume https://github.com/kubewarden/github-actions/pull/309

Makefile
jvanz: > It's already included in history, see [here](https://github.com/kubewarden/adm-controller/pull/1705/changes#diff-9eed8e9d0218e3c2954dd740cdb147d5527635cb8188030f10cd1b541c8e913dL1).

That's the file that I was mentioning to not be committed in the history. I've open the thread here because this is the place where that file is created. ;)

> With the change, it moves from config/webhook/manifests.yaml to charts/generated-webhooks-manifest.yml (should show on the commits, but I think I mangled the commit and I just added the new file and then removed the old),
> I'm ok adding a warning, but it will be printed every time we run make manifests. I thought that when committing changes, and in PRs, we will see the bump to the generated file and wonder, given that the file has that first line "to be merged manually". Would that be enough or would you like me to add an echo with a reminder for make manifests?

I'm not big fan of adding the file that will never be used. That's the original reason of removing the `config` directory. Even if I'm not sure if the idea to detect changes in the webhooks would be efficient, I don't have a better idea besides the warning in the makefile. Thus, I'm fine try it out and iterate it afterwards if necessary.

jvanz (approved):

PR #1733, opened 2026-05-11
fix(kwctl): malformatted markdown installing text

jvanz:
## Description

Updates the installation markdown text used by kwctl in the Artifacthub file generation. The text generated a weird text formatting in the web interface with the `kwctl` used in a link.

Example of the issue in the web interface:

image


flavio:
@jvanz LGTM, once the broken tests are fixed

viccuad (approved):
Same.
Also notice, we would need to release this and consume it via github-actions, for it to take effect when annotating policies.

PR #1738, opened 2026-05-13
fix(chart): restrict restart hook to sidecar telemetry mode

jvanz:
## Description

Limit the post-install/post-upgrade restart hook to sidecar mode. This prevents a second rollout in custom telemetry mode that can cause hostPort deadlocks on single-node hostNetwork upgrades.

Found while I was testing https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/285


flavio (approved):
thanks for the fix

PR #1744, opened 2026-05-14
refactor(controller): drive policy-server metrics port via CLI flag

### LINKED ISSUE #787: Move metrics port configuration from env to cli flag

We provide a way to configure the port where the Policy Server will be exposing metrics to be collected, by using an environment variable.
However, this is the only env variable we expose, as the controller configuration is done by using cli flags.

### Solution you'd like

We need to move the metrics configuration logic to the `cmd/main.go` file by using cli flags instead of using the golang [`init()`](https://github.com/kubewarden/kubewarden-controller/blob/0915fa0fe5ec9679dce211bb40a60858c766a825/internal/admission/policy-server-service.go#L23) function, and pass it to the reconciler as a member of the reconciler struct.

This involves changing the helm chart as well: [2](https://suse.slack.com/archives/D05ARV1NS14/p1719405156029899)
https://github.com/kubewarden/helm-charts/blob/76dd5ccad7b11a4346ab8fa58371f8f259511636/charts/kubewarden-controller/templates/deployment.yaml#L53

_No response_

### Anything else?

_No response_

officialasishkumar:
## Summary

Closes #787.

The controller exposed exactly one piece of its configuration through an environment variable,
`KUBEWARDEN_POLICY_SERVER_SERVICES_METRICS_PORT`, while every other knob already lived behind a
`flag.*` CLI flag. The mismatch made the documentation harder to keep consistent and forced the
Helm chart to use a different mechanism for this single setting.

This PR moves the configuration onto the existing CLI surface as a new
`--policy-server-metrics-port` integer flag and updates everything downstream of that change.

## What changes

| File | Change |
|------|--------|
| `cmd/controller/main.go` | Adds the `--policy-server-metrics-port` flag (default `constants.PolicyServerMetricsPort`, validated against the same `[1, 65535]` range as `--webhook-server-port`). Drops the `os.Getenv` block plus the `strconv` import. Removes the now-unused `gocognit` nolint on `main()`. |
| `internal/constants/constants.go` | Drops `PolicyServerMetricsPortEnvVar`. |
| `api/policies/v1/policyserver_types.go` | The PolicyServer CRD doc comment for `metricsPort` now points users at the new flag. |
| `charts/kubewarden-crds/templates/crds/policies.kubewarden.io_policyservers.yaml` | Regenerated from the updated comment. |
| `docs/crds/CRD-docs-for-docs-repo.{adoc,md}` | Regenerated from the updated comment. |
| `charts/kubewarden-controller/templates/deployment.yaml` | Emits `--policy-server-metrics-port=...` inside the existing sidecar-mode args block (where the env var was set previously) and drops the matching `env:` entry. |
| `charts/kubewarden-controller/tests/telemetry_configuration_test.yaml` | The two assertions that pinned the env-var name are rewritten as `args:` assertions for the new flag. |
| `internal/controller/policyserver_controller.go` | Doc comment on `PolicyServerMetricsPort` now describes the CLI flag origin. |
| `internal/controller/policyserver_controller_service_test.go` | The test commentary and the two subtest names that referenced "env var" are renamed to "CLI flag". |

## Compatibility note

Operators who run the controller binary directly with the old env var set will need to switch to
`--policy-server-metrics-port=<n>`. This breaking change matches the request in the issue. The
Helm chart upgrade is transparent because the chart drives both ends.

## Test plan

- [x] `go build ./...` builds cleanly
- [x] `go vet ./...` clean
- [x] `golangci-lint run ./cmd/... ./internal/... ./api/...` reports zero issues
- [x] `make test-go` passes all 58 envtest controller specs
- [x] `make helm-unittest` passes for all three charts (62 + 25 + 5 tests). The
  `telemetry configuration` suite still validates that the sidecar-mode flag is present
  and that non-sidecar mode does not emit it.
- [x] `make generate` is idempotent (the regenerated CRD chart and CRD docs only differ in
  the comment line that switched from env-var name to CLI flag name).


charts/kubewarden-controller/templates/deployment.yaml
```diff
@@ -82,6 +82,9 @@ spec:
        {{- if or .Values.telemetry.metrics .Values.telemetry.tracing }}
        {{- if and (eq .Values.telemetry.mode "sidecar") }}
         - --enable-otel-sidecar
+        {{- if .Values.telemetry.metrics }}
```
jvanz: If you migrating to the CLI, please ensure the helm chart template follow the same "if"

cmd/controller/main.go
```diff
@@ -103,7 +102,7 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
-//nolint:funlen,gocognit // Avoid splitting the main function in multiple functions to avoid changing the retcode logic for metrics shutdown
+//nolint:funlen // Avoid splitting the main function in multiple functions to avoid changing the retcode logic for metrics shutdown
```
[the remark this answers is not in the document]
officialasishkumar: Restored the `gocognit` suppression in 0bae37d.

cmd/controller/main.go
```diff
@@ -195,31 +198,21 @@ func main() {
...
+			"invalid policy server metrics port",
+			"flag", "--policy-server-metrics-port",
+			"value", *policyServerMetricsPortFlag,
+			"min", minAllowedPort,
+			"max", maxAllowedPort,
+		)
+		retcode = 1
+		return
 	}
+	policyServerMetricsPort := int32(*policyServerMetricsPortFlag)
```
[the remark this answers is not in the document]
officialasishkumar: Added the deprecated env-var fallback with migration warnings in 0bae37d; explicit `--policy-server-metrics-port` still takes precedence.

cmd/controller/main.go
```diff
@@ -120,6 +120,10 @@ func main() {
 	flag.StringVar(&mgrOpts.MetricsAddr, "metrics-bind-address", ":8088", "The address the controller-runtime metric endpoint binds to.")
 	flag.StringVar(&mgrOpts.ProbeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.IntVar(&mgrOpts.WebhookServerPort, "webhook-server-port", 9443, "The port the webhook server listens on.")
+	policyServerMetricsPortFlag := flag.Int("policy-server-metrics-port",
+		constants.PolicyServerMetricsPort,
+		"The default port exposed by every PolicyServer metrics Service. "+
+			"Per-PolicyServer overrides (spec.metricsPort) always take priority.")
```
jvanz: Please, follow the other CLI flags parttern. Declare the var above and use it in the IntVar call.

jvanz (changes requested):

cmd/controller/main.go
officialasishkumar: Updated in f3190f5. `policyServerMetricsPort` is now declared with the other local flag variables and registered through `flag.IntVar`, matching the existing pattern.

charts/kubewarden-controller/templates/deployment.yaml
officialasishkumar: Updated in f3190f5. The template now uses the same `and .Values.telemetry.metrics (eq .Values.telemetry.mode "sidecar")` condition for `--policy-server-metrics-port`, with Helm unittest coverage for sidecar tracing-only mode.

jvanz:
Thanks for taking the time to submit this. Community contributions help, and we do read everything that comes in.

But I have to say that his will probably sit for a bit. We got hit with a wave of PRs recently, and the team is small. Right now we're focused on some roadmap work with deadlines, so external PRs keep getting bumped.

Our policy is that we review everything, but that can take longer then the expected when we're crunched like this. There's just more coming in than we can keep up with at the moment.

We'll get to yours as soon as we free up. Thanks for being patient with the wait!

PR #1760, opened 2026-05-22
chore(deps): update to latest version of sigstore-rs

flavio:
Bump the sigstore dependency to latest version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,7 +43,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
- "cipher",
+ "cipher 0.4.4",
  "cpufeatures 0.2.17",
 ]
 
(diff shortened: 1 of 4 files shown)


flavio:
It would be interesting to extend the tests to make sure always gained v3 signature support (which should be the case).

I don't have time right now to update that, but if anybody wants...

- {KUBEWARDEN_CONTROLLER_IMAGE}:v1.29.0-rc1: simple signing only
- {KUBEWARDEN_CONTROLLER_IMAGE}:v1.30.0-rc1: sigstore bundle signing only
- {KUBEWARDEN_CONTROLLER_IMAGE}:v1.31.0-rc1: sigstore bundle and simple signing

jvanz (approved):
LGTM.

> It would be interesting to extend the tests to make sure always gained v3 signature support (which should be the case).
> I don't have time right now to update that, but if anybody wants...

 I would like to add these tests. But I was trying to understand which test exactly you are looking for... are you suggesting adding a policy-fetcher test to test the verification only for some public container images? Or do you want to update the more complex sigstore test where we sign and verify a testing policy?

I believe you talking about the former option. Am I correct?

viccuad (approved):
LGTM. On first view I thought we needed to remove triangulate() calls, but I see that we use that to return the digest.

PR #1774, opened 2026-05-27
feat(kubewarden-controller): Delete all CRs in pre-delete hook

viccuad:
## Description

On 1.36, without this change:
- Deleting a PolicyServer sets its policies as scheduled.
- Recommended policies are managed via Helm and removed when `kubewarden-defaults` is uninstalled.
- User-managed policies will never be deleted, and linger after the removal of the CRDs, as they have finalizers.

With this change:
All CRs will be removed when uninstalling the `kubewarden-controller` chart. This includes charts of downstream consumers that use a single chart approach.

The used image, kuberlr/kubectl, has busybox, hence `ash` as shell. Use the shell to remove all CRs. Removing all CRs in potentially all namespaces is costly; do backgrounds jobs with `&`, wait, and only then remove the policyservers.

## Test

Tested locally by deploying charts with recommended policies, a user-managed policy and policy server, and uninstalling the charts one by one.

## Checklist

### DIFF
--- a/charts/kubewarden-controller/templates/pre-delete-hook.yaml
+++ b/charts/kubewarden-controller/templates/pre-delete-hook.yaml
@@ -36,7 +36,16 @@ spec:
       containers:
         - name: pre-delete-job
           image: '{{ template "system_default_registry" . }}{{ .Values.preDeleteJob.image.repository }}:{{ .Values.preDeleteJob.image.tag }}'
-          command: ["kubectl", "delete", "--all", "policyservers.policies.kubewarden.io"]
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete --wait --all clusteradmissionpolicies.policies.kubewarden.io &
+              kubectl delete --wait --all --all-namespaces admissionpolicies.policies.kubewarden.io &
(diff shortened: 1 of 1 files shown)


charts/kubewarden-controller/templates/pre-delete-hook.yaml
```diff
@@ -36,7 +36,16 @@ spec:
       containers:
         - name: pre-delete-job
           image: '{{ template "system_default_registry" . }}{{ .Values.preDeleteJob.image.repository }}:{{ .Values.preDeleteJob.image.tag }}'
-          command: ["kubectl", "delete", "--all", "policyservers.policies.kubewarden.io"]
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete --wait --all clusteradmissionpolicies.policies.kubewarden.io &
```
viccuad, line 43: We used to not have the `wait`s. It may be better to remove them.

jvanz (commented):
Overall, LGTM. But I think that copilot comments are good suggestions.

flavio:
closing as discussed during the daily

PR #1789, opened 2026-06-01
fix(controller): Append controller env vars last on policy-server

viccuad:
## Description

Be more defensive when setting env vars on policy-server Deployment.

User-provided vars come first, and the controller hardcoded variables are appended last.  In Kubernetes, when duplicate env var names exist in a container, the last definition wins (for most runtimes), so this ensures the controller's values always take precedence.

## Test

CI

## Checklist

### DIFF
--- a/internal/controller/policyserver_controller_deployment.go
+++ b/internal/controller/policyserver_controller_deployment.go
@@ -694,7 +694,10 @@ func getPolicyServerContainer(policyServer *policiesv1.PolicyServer) corev1.Cont
 				MountPath: policyStoreVolumePath,
 			},
 		},
-		Env: append([]corev1.EnvVar{
+		// user-provided vars from Spec.Env come first, controller-managed vars are
+		// appended last so they always take precedence (kubelet applies
+		// "last wins" semantics when duplicate names are present)
+		Env: append(policyServer.Spec.Env, []corev1.EnvVar{
 			{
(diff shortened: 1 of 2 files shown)


viccuad:
Opened as draft PR, would prefer to wait for 1.36 GA to merge this.

internal/controller/policyserver_controller_deployment.go
```diff
@@ -694,7 +694,7 @@ func getPolicyServerContainer(policyServer *policiesv1.PolicyServer) corev1.Cont
 				MountPath: policyStoreVolumePath,
 			},
 		},
-		Env: append([]corev1.EnvVar{
+		Env: append(policyServer.Spec.Env, []corev1.EnvVar{
```
jvanz, line 700: As expected, the code it's fine. I would just add a comment above to reminder the future ourselves of why we should do that. Just to avoid a regression because we forget about it. Or even better add a test to ensure we will not forget about this.

viccuad: Indeed, good point! I added both and rebased the commit.

jvanz (approved):
LGTM. Thanks for the changes! But now we need to fix the linter issue. hehe

internal/controller/policyserver_controller_deployment.go
jvanz: Thanks!

flavio (approved):

PR #1804, opened 2026-06-09
fix(kwctl): improve reporting about host capabilities

flavio:
Prior to this change, the `kwctl annotate` did a too naive comparison between what the policy metadata declared and what was actually used. The comparison didn't support `*` inside of metadata, hence a lot of wrong complains were returned to the user.

### DIFF
--- a/crates/kwctl/src/annotate.rs
+++ b/crates/kwctl/src/annotate.rs
@@ -6,7 +6,8 @@ use std::{
 
 use anyhow::{Result, anyhow};
 use policy_evaluator::{
-    ProtocolVersion, constants::*, policy_metadata::Metadata, validator::Validate,
+    ProtocolVersion, constants::*, host_capabilities::HostCapabilities, policy_metadata::Metadata,
+    validator::Validate,
 };
 use tracing::warn;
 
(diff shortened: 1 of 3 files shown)


crates/kwctl/src/annotate.rs
```diff
@@ -155,6 +219,111 @@ mod tests {
...
+        vec![],    // detected
+        vec!["*"], // declared
+        vec![],    // used_but_undeclared
+        vec!["*"], // declared_but_unused
+    )]
+    #[case::prefix_covers_single(
+        vec![("oci", "v1/verify")], // detected
+        vec!["oci/*"],              // declared
+        vec![],                     // used_but_undeclared
+        vec![],                     // declared_but_unused
```
viccuad, line 257: Shouldn't here be also `"oci/*"`, just like in line 269 or 275?

Same for line 263, shouldn't that line be also `"oci/*"`?

viccuad (approved):
LGTM, just one small question that may point a bug.

crates/kwctl/src/annotate.rs
jvanz: It looks right to me. The metadata info (`declared`) says that the policy uses everything from `oci/*` and the `detect` is in fact using a host capability from the `oci` path. Thus, the declared host capability is in fact used. 

What may be can be improved is that `oci/*` give access to [more](https://github.com/kubewarden/adm-controller/blob/0af49d28d746a8837498393b787a215516896682/crates/policy-evaluator/src/host_capabilities.rs#L29-L38) host capabilities that are declared but not used. But as far as I can see, the goal here is just compare the policy data. Not perform the actual host capability tree validation. Is that correct? If so, looks fine to me.

jvanz (approved):

crates/kwctl/src/annotate.rs
viccuad: I agree that it's using the correct thing, but I see inconsistency between  line 257 or 263 and 269 or 275.

flavio: > What may be can be improved is that oci/* give access to more host capabilities that are declared but not used. But as far as I can see, the goal here is just compare the policy data

Yes, this type of policy is giving access to more capabilities compared to the ones used. I tried to keep the code simple.

The purpose is to give some quick validation to the policy author. IMHO the priority is to find host capabilities being used but not declared (the author forgot about them) and to point out the ones that are declared but not used ("leftovers").

I also wanted to call out "laziness", like an author just giving `*`.

I didn't consider `oci/*` "laziness. If you want I can try to call out also these cases.

PR #1813, opened 2026-06-14
fix(controller): reduce API server load and policy status update churn

### LINKED ISSUE #743: Restructure controller reconciliation loops to reduce the API server load
This is a spin off of the issue #645. In that issue an user was complaining that the controller is doing too many requests to the API server. Specially regarding to the `configmap` resource. The controller is hitting the API server many times to get the `configmap` because all the reconciliation loops uses that information. However, this resource is not cached in the controller. Therefore, the first mitigation proposed in that issue was enable cache for the `configmap` resources. In the same issue another users reported that the controller is updating the policy status many times. This is also true. Because the reconciler waits for the policy server reconciliation to properly set the policy status. And every time policy server status is "not ready", it update the policy status and requeue a request to check again. Therefore, many request updating the policy status, even with the same status value, are sent. 

Both issues, the many `get` requests for the configmaps and the many policy status updates has a common root. All are related to the way the reconciliation loops wait for each other and get/updates many resources in the process. Therefore, we should review how the 3 reconciler work together and try to simplify them reducing the amount of time the policy status are updated and how many times we fetch data from the API server.

#### Acceptance criteria
- Investigate if we can simplify the reconcilers in the controller reducing the number of resources requested and policy status updates over time. 
- Simplify the controller reconciliation loops where is possible.

steverogersX:
## Description

Fix #743

Reduces the API server load and the policy status update errors described in
the issue. The reconcilers were re-writing status and re-fetching resources on
every requeue, even when nothing changed.

Note: the loop simplification is left for a follow-up PR, since it touches the
pod-readiness path and needs the e2e cluster to validate.

## Test

```shell
cd internal/controller
go test ./...
Checklist
 I have read and understood the [Kubewarden AI Policy](https://github.com/kubewarden/community/blob/main/AI_POLICY.md)

```

### DIFF
--- a/internal/controller/policy_subreconciler.go
+++ b/internal/controller/policy_subreconciler.go
@@ -24,6 +24,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


steverogersX:
hi, @jvanz 
can you review my code?

jvanz:
@steverogersX thanks for the PR! But unfortunately, I'll close it. 

Taking a look in the history of the opened issue and the PR I think this is not worth to add now. The cache is being able to reduce the load on API server quite well. Trying to guess what past myself was thinking, it looks like a over engineering. No one is complaining about this. Therefore, there is not good reason to add this code in the reconciler code. Furthermore, we are thinking of doing a [restructure](https://github.com/kubewarden/rfc/blob/main/rfc/0022-policy-lifecycle.md) of the code in the near/medium future. Which means that all the code can change again.  

I would like to ask sorry for you for closing this right away. Also, we noticed that you're trying to contribute to the project. Witch is really appreciate! I just want to ask you to ping us in the issue that you're planning to work on before starting coding. Therefore, we can review the issue again, decide if it's necessary and how to move forward. Thus, we do not waste your time. How it sounds?

Thanks and sorry again! I hope to see you around in other issues!

steverogersX:
@jvanz 
No problem! I learned a lot working through this, so it wasn't wasted time at all. Will ping the issue first before coding next time. Looking forward to contributing more to the project.

PR #1802, opened 2026-06-08
feat!: admission controller Helm chart consolidation

jvanz:
## Description

This is the first version of the consolidation of all the 3 admission controller Helm charts into one.

> [!NOTE]
> Even if the amount of the line changes are quite big, if reviewed per commit bases the process will not be that hard. Because many changes are file moves/deletion and the context of the changes help to makes sense of it. 

## Test

You can run the tests and or play with it locally. Tilt environment should work. 

> [!WARNING]
> The Helm chart e2e test CI will fail until de adapt the test to the this new Helm chart
> The zizmor also will be red until we merge this [commit](https://github.com/kubewarden/adm-controller/pull/1778/changes/538521c2833a244b7646886faea853f97a759b1b) from another [PR](https://github.com/kubewarden/adm-controller/pull/1778/commits)

PR #1816, opened 2026-06-16
Release/v1.37 migration changes

jvanz:
## Description

This changes all required to allow a migration to the single helm chart defined at #1802. This changes will allow users to do a migration with no downtime for user resources and resources installed by the `kubewarden-defaults`. 

In summary, the changes necessary to allow the migration are adding the `helm.sh/resource-policy: keep` annotations to tell Helm to not remove important resource that must be exist to keep the policies and policy server continue to work after the Helm charts uninstall. And,  the pre-delete hook must avoid removing user defined resources.

### DIFF
--- a/charts/kubewarden-crds/templates/crds/policies.kubewarden.io_clusteradmissionpolicies.yaml
+++ b/charts/kubewarden-crds/templates/crds/policies.kubewarden.io_clusteradmissionpolicies.yaml
@@ -4,6 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.18.0
+    helm.sh/resource-policy: keep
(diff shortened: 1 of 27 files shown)


charts/kubewarden-crds/templates/crds/policies.kubewarden.io_clusteradmissionpolicies.yaml
```diff
@@ -4,6 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.18.0
+    helm.sh/resource-policy: keep
```
flavio: Same here

jvanz:
Closing in favor of another solution described [here](https://github.com/kubewarden/adm-controller/pull/1802#issuecomment-4736278255)

PR #1819, opened 2026-06-18
Rename chart to adm controller

jvanz:
## Description

Rename the new single helm chart to adm-controller and update the migration script to allow the migration after the renaming. This change in the migration script is necessary to avoid issue changing immutable fields.


kravciak:
I am wondering, why do we use adm-controller instead of admission-controller?
Do we also consider short names for other products?

jvanz:
> I am wondering, why do we use adm-controller instead of admission-controller? Do we also consider short names for other products?

That's a good question. I've used `adm-controller` because it was the suggestion from the single helm chart PR. And, as people working in software industry have the tendency to use short as much as possible, I did not think too much about it.  :laughing: 

To be honest, I don't have strong opinion about this. Maybe the full name is a good idea. It's compatible with the sbomscanner chart [name](https://github.com/kubewarden/helm-charts/blob/e1afbe2cd056863b504a89d69cfa6227821315a3/charts/sbomscanner/Chart.yaml#L2)

jvanz:
> I am running from this PR (with latest images). Do we want to keep `kubewarden-` name prefix?

I do not see issues on that.

kravciak:
> > I am running from this PR (with latest images). Do we want to keep `kubewarden-` name prefix?
> 
> I do not see issues on that.

Sorry, I deleted comment as soon as I posted, it was my problem with e2e :)

kravciak:
we have `app.kubernetes.io/managed-by=kubewarden-controller`
```
k get pods -n kubewarden --show-labels | grep kubewarden-
policy-server-default-c488fd889-9n5r8   1/1     Running     0          91s   app.kubernetes.io/component=policy-server,app.kubernetes.io/instance=policy-server-default,app.kubernetes.io/managed-by=kubewarden-controller,app.kubernetes.io/part-of=kubewarden,app=kubewarden-policy-server-default,kubewarden/config-version=9195,kubewarden/policy-server=default,pod-template-hash=c488fd889
```

charts/admission-controller/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: kubewarden-controller
+name: adm-controller
 description: A Helm chart for deploying the Kubewarden stack
```
flavio, line 3:
```suggestion
description: A Helm chart for deploying the Kubewarden admission controller stack
```

flavio (commented):
Overall LGTM.

As for the naming, this is always a hard topic 😅 

I think we could name the chart `kw-adm-controller`, the current name leads to a somehow vague name:

```console
helm ls -n kubewarden
NAME            NAMESPACE       REVISION        UPDATED                                         STATUS          CHART                           APP VERSION
adm-controller  kubewarden      1               2026-06-19 09:26:16.792454777 +0200 CEST        deployed        adm-controller-6.0.0-alpha.1    v1.37.0-alpha.1
```

charts/admission-controller/templates/controller/service.yaml
jvanz, line 1: This is not a new file. It just a "rename". However, git, for some reason, could not detect the renaming. You can see the same file being deleted in this PR. :shrug:

charts/admission-controller/templates/controller/controller-serviceaccount.yaml
jvanz, line 1: This is not a new file. It just a "rename". However, git, for some reason, could not detect the renaming. You can see the same file being deleted in this PR. :shrug:

jvanz:
@kubewarden/kubewarden-developers I've renamed the helm chart to `admission-controller`. I'm investigating why the go e2e tests are not passing and testing the migration again. But you can start the review already.

kravciak:
Last issue with e2e tests is wrong versions in hauler:
```bash
✗ [basic-tests.bats] Version checks [3426]
allow-privilege-escalation-psp: [haul:v1.0.11 helm:v1.0.10]
capabilities-psp: [haul:v1.0.11 helm:v1.0.10]
host-namespaces-psp: [haul:v1.1.9 helm:v1.1.8]
hostpaths-psp: [haul:v1.1.8 helm:v1.1.7]
pod-privileged: [haul:v1.0.11 helm:v1.0.10]
user-group-psp: [haul:v1.1.5 helm:v1.1.5]
```

Otherwise https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/313 is passing locally with this PR.

flavio (approved):

PR #1820, opened 2026-06-18
feat(migration): install unified chart from merged legacy values

jvanz:
## Description

The migration script builds a values file from helm get values output of the three legacy releases (kubewarden-crds, kubewarden-controller, kubewarden-defaults) and uses that to install the unified chart instead of hardcoded --set flags.
    
nameOverride gets filled in when absent to preserve resource identity for in-place adoption. If the legacy controller release set it, the script uses that. Otherwise it defaults to kubewarden-controller to match the old resource names. User-passed --set and --values override the merged values.

This keeps whatever configuration you had on the legacy stack. Note that helm get values returns only user overrides, not chart defaults. If you relied on a default without setting it explicitly, pass it with --set or --values.

### DIFF
--- a/charts/admission-controller/README.md
+++ b/charts/admission-controller/README.md
@@ -224,13 +224,16 @@ detects the legacy release name automatically.
 `<release>-admission-controller`). The legacy resources being adopted were
 created by the old `kubewarden-controller` chart and carry
 `app.kubernetes.io/name: kubewarden-controller` in the same immutable
-selector. The script therefore installs with
-`--set nameOverride=<legacy name>` (read from the legacy release's Helm
-values, defaulting to `kubewarden-controller`) so the chart reproduces
-the old names/labels and Server-Side Apply adopts the objects in place.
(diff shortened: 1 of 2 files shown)


flavio:
> nameOverride gets filled in when absent to preserve resource identity for in-place adoption. If the legacy controller release set it, the script uses that. Otherwise it defaults to kubewarden-controller to match the old resource names. User-passed --set and --values override the merged values.

What happens in this scenario:

- User installs the 3 helm charts (1.36)
- User migrates to single helm chart using our script (1.37) <- here the name is preserved, right?
- User upgrades to new version of adm-controller (1.38).

When upgrading to 1.38, does the user have to do something special compared to an user who never used the 3 helm charts?

flavio:
> When upgrading to 1.38, does the user have to do something special compared to an user who never used the 3 helm charts?

Confirmed, the user has to keep specifying to reuse the old release name when doing upgrades. This is a shortcoming that is already documented in the instructions we provide to the user.

PR #1829, opened 2026-06-22
docs(chart): explain CRD reinstall ownership conflict

jvanz:
## Description

Kept CRDs keep their original Helm ownership metadata after an uninstall, so reinstalling under a different release name or namespace fails with an opaque invalid ownership metadata error. Document why this happens and how to recover, either by installing with --take-ownership

### DIFF
--- a/charts/admission-controller/README.md
+++ b/charts/admission-controller/README.md
@@ -288,6 +288,31 @@ CRDs are installed with the `helm.sh/resource-policy: keep` annotation:
 - `helm upgrade` updates CRDs normally
 - `helm uninstall` does not delete CRDs, which prevents cascade-deletion of all PolicyServers and policies in the cluster
 
+#### Reinstalling under a different release name or namespace
+
+Because the CRDs are kept on uninstall, they survive with the Helm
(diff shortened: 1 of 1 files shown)


charts/admission-controller/README.md
```diff
@@ -288,6 +288,31 @@ CRDs are installed with the `helm.sh/resource-policy: keep` annotation:
...
+  Error: ... invalid ownership metadata; annotation
+  meta.helm.sh/release-name must equal "<new>": current value is "<old>"
+  ```
+
+This is expected: the CRDs still belong to the previous release. To adopt
+them into the new release, install with `--take-ownership` (Helm 3.18+ or
+Helm 4), which re-stamps the ownership metadata:
+
+```sh
+helm install <release> <chart> -n <namespace> --take-ownership
```
flavio, line 313: This is already done by our migration script, isn't it? This is more for people who run into this rename scenario because of other reasons... am I right?

flavio: Clarified with @jvanz, I was right. The migration script is not affected by that.

PR #1839, opened 2026-06-25
fix(chart): update migration script to avoid name override in future upgrades

jvanz:
## Description

The migration script and the Helm chart has been updated to avoid future upgrades in the new single Helm chart the need of the users to remember to add the old name in the values. Basically, the helm chart name is removed from the selector and the migration script handle the path from the old helm chart to the new one.


README.md
```diff
@@ -110,25 +106,41 @@ kubectl apply -f clusteradmissionpolicygroups-backup.yaml
...
-can delete resources or break admission, so the script handles it all
-and verifies each step.
-
-What survives the migration:
+The Kubewarden team ships a script that runs the whole migration to the unified
+chart with no downtime. Check the prerequisites below before you run it.
+
+If you can't meet one of the prerequisites, back up the resources you have, which
+is the policy servers and policies, and reinstall them once the stack is back up.
+That path causes downtime in policy evaluation.
```
flavio:
```suggestion
If you can't meet one of the prerequisites, follow the "manual reinstallation"
instructions for your own safety.
```

README.md
```diff
@@ -243,58 +263,38 @@ Passing custom values to the unified chart:
...
+only the configuration you set yourself. Later upgrades need nothing special: a
+plain `helm upgrade admission-controller ... --reuse-values` is enough. The
+release is named `admission-controller`, and so are its controller resources.
+
+**Controller gap.** Between the legacy uninstall and the new chart becoming
+ready, no controller is running, since its plumbing is recreated rather than
+adopted. The surviving policy-server pods still serve the existing webhook
+configurations, so admission for active policies keeps working. Any policy CR you
+create during this window waits until the new controller starts before it becomes
+a webhook.
```
flavio: The user won't be able to define and modify policy CRs since they have to be validated by the webhook that runs on the controller. They will get a rejection from the API server

README.md
```diff
@@ -243,58 +263,38 @@ Passing custom values to the unified chart:
...
+**Legacy defaults not carried over.** The script builds the merged values with
+`helm get values`, which returns only the overrides you supplied, not `--all`.
+Anything that was just a legacy chart _default_ and you never set yourself does
+not come across. Pass those with `--set` or `--values` if you relied on them. One
+to watch: the unified chart defaults to `recommendedPolicies.enabled: false` and
+`recommendedPolicies.defaultPolicyMode: "monitor"`, and the script no longer
+forces `enabled=true` / `protect`. If the old defaults chart turned recommended
+policies on for you and you never changed that, add
+`--set recommendedPolicies.enabled=true` and
+`--set recommendedPolicies.defaultPolicyMode=protect`.
```
flavio: I got confused by this section. I doubled checked, we kept the same default values for `recommendedPolicies.enabled` and `recommendedPolicies.defaultPolicyMode`.

Do you mean that, if a user did some customization to the `kubewarden-defaults` chart at installation time, these new values are **not** propagated by the migration tool?

flavio (commented):
This README is becoming too big. I think we should do something like...

Inside of the "Migration from three-chart setup" section:
- Brief explanation, starting from 1.37.0 there's only one helm chart that manages the whole KW Admission Controller stack instead of 3
- Migration instructions are available at <link to docs>. Like we do inside of the "Installation" section
- Move all the contents you wrote into a new dedicated page of the docs

The migration doc should be referenced also inside of:

- README of the single helm chart. Are you migrating from an older version of KW Adm controller that uses 3 charts? check this out
- READMEs of the old charts: deprecation notice, link to the migration docs

README.md
```diff
@@ -48,22 +48,19 @@ more details.
...
-without admission downtime. Or uninstall the old Helm charts and reinstall the
-new single Helm chart acknowledging that this procedure will cause down time in the
-policy evaluations.
+`kubewarden-controller`, `kubewarden-defaults`), you have two ways to move to the
+single chart. The migration script
+(`kubewarden-unified-adm-controller-chart-migration.sh`) does it without
+admission downtime. Otherwise you can uninstall the old charts and reinstall the
+new one, which stops policy evaluation while you do it.
 
 ### Manual reinstallation
```
flavio: I **think** quite some user put the customization the make into their own `values.yaml` file instead of specifying them via a multitude of `--set` flags.

We've kept the same names into the new `values.yaml`, can a user concatenate all the different `values.yml` files into a single one and use that at installation time with the new single helm chart?

README.md
```diff
@@ -243,58 +263,38 @@ Passing custom values to the unified chart:
...
-`--values` to preserve your configuration.
+**Plain release, no overrides.** The migrated release stores no
+`nameOverride` or `fullnameOverride`. The plumbing is recreated under the
+chart's own names instead of being adopted in place, so there is no immutable
+selector to preserve. The script builds the values file from `helm get values`
+on the three legacy releases (`kubewarden-crds`, `kubewarden-controller`,
+`kubewarden-defaults`) and strips any name or fullname overrides, so it carries
+only the configuration you set yourself. Later upgrades need nothing special: a
+plain `helm upgrade admission-controller ... --reuse-values` is enough. The
+release is named `admission-controller`, and so are its controller resources.
```
flavio: I remember we were discussing the fact that, if someone used a custom release name, then we would have to use some extra flags at migration time and the user would have to deal with future upgrades in a special way (by continuously passing this name).

Am I mixing things up?

jvanz: This PR is trying to fix this issue for future upgrades. The PR change the helm chart to not use the name in the selectors to avoid this issue in the future. And, manage the migration to the new controller deployment using the new set of selectors.

README.md
jvanz: Yes. This is the price to be pay to allow allow future upgrades without the need to remember to set the nameOverride forever.

README.md
jvanz: This documentation paragraph is wrong. 

The documentation is saying that if the user rely on the old helm chart default values, and the new default values from the new helm chart is different, the final values used to install the unified helm chart will be different. This is caused because the migration script only get the values customized by the user in the installation of the old helm charts. However, the default values are the same in all the old and new helm charts. Therefore, the default values in the unified chart are the same.

charts/admission-controller/tests/controller_selector_test.yaml
```diff
@@ -0,0 +1,16 @@
...
+      name: my-release
+    asserts:
+      - equal:
+          path: spec.selector.matchLabels["app.kubernetes.io/component"]
+          value: controller
+      - equal:
+          path: spec.selector.matchLabels["app.kubernetes.io/instance"]
+          value: my-release
+      - notExists:
+          path: spec.selector.matchLabels["app.kubernetes.io/name"]
```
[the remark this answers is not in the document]
flavio: I think this is reasonable too

README.md
flavio: I see, nice idea. I looked at the selector and I think it should be fine

README.md
flavio: Yes, I'm fine with that. I think the paragraph should be expanded to clarify that it won't be possible to create or edit policies during this time. The current text might lead someone to think they can define new policies, but these will be enforced only later once the controller is back; which is not the case, they will get a API server error.

README.md
flavio: Given we haven't changed the default values, I would remove this text to avoid confusion.

charts/admission-controller/kubewarden-unified-adm-controller-chart-migration.sh
```diff
@@ -507,7 +517,7 @@ EOF
 phase_uninstall_legacy() {
   step "Phase 4: Uninstall legacy releases (--no-hooks)"
 
-  confirm "About to uninstall the three legacy Helm releases with --no-hooks. Resources will be preserved by keep annotations."
+  confirm "About to uninstall the three legacy Helm releases with --no-hooks. Data/identity resources (CRDs, cert Secrets, RBAC, PolicyServers, policies) are preserved by keep annotations; the controller's stateless plumbing is removed and recreated by the unified install. This causes a brief controller-only restart — policy enforcement on workloads is unaffected (served by the running policy-server pods)."
```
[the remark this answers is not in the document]
flavio: I would take the grammar fixes and make it a bit more DRY. Like skip the part about the stateless plumbing. Focus on what is kept and warn about the short blackout of the controller (and its implications)

flavio (changes requested):

README.md
jvanz: Yes, they can. This is mentioned in the manual installation procedure:

```
The new single Helm chart also merges the three values files into one, and your
old values still apply. To get the same stack back, combine the values from the
old three charts into one file and pass it to the install:


helm install kubewarden kubewarden/admission-controller -n kubewarden --values all-values.yaml

```

charts/admission-controller/kubewarden-unified-adm-controller-chart-migration.sh
```diff
@@ -507,7 +517,7 @@ EOF
 phase_uninstall_legacy() {
   step "Phase 4: Uninstall legacy releases (--no-hooks)"
 
-  confirm "About to uninstall the three legacy Helm releases with --no-hooks. Resources will be preserved by keep annotations."
+  confirm "About to uninstall the three legacy Helm releases with --no-hooks. CRDs, CA cert Secret, RBAC, PolicyServers, and policies are kept. The controller goes down for a short time. Existing policy enforcement keeps running, but attempts to create new policies or PolicyServers will fail until the unified chart is installed."
```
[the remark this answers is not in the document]
jvanz: @flavio, this actually true. The policy/policy server creation will not return an error from API server because the uninstall of the old helm chart will remove the [webhooks](https://github.com/kubewarden/helm-charts/blob/d8a12e9b96e992e3154c93642c050c257b31be73/charts/kubewarden-controller/templates/webhooks.yaml#L89-L327). Therefore, the API server will not try to send a request to the kubewarden admission controller.

charts/admission-controller/templates/controller/deployment.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   replicas: {{ .Values.replicas }}
   selector:
     matchLabels:
-      {{- include "admission-controller.selectorLabels" . | nindent 6 }}
+      {{- include "admission-controller.controllerSelectorLabels" . | nindent 6 }}
   template:
```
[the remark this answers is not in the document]
jvanz: This change is actually to fix the upgrade path from the legacy helm charts. This should not change for future upgrades of the single helm chart.

flavio (approved):

viccuad (approved):
LGTM too.

PR #1840, opened 2026-06-25
chore(deps): update wasmtime to 46.0

jvanz:
## Description

Update wasmtime and related crates to version 46.0.1 to stay current with the latest features and bug fixes. The wasmtime-provider crate was also updated to 2.21.0 for compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ clap-documentation = "0.1.0"
 futures = "0.3"
 hex = "0.4"
 itertools = "0.14.0"
-k8s-openapi = { version = "0.27.0", default-features = false, features = [
+k8s-openapi = { version = "0.28.0", default-features = false, features = [
   "v1_33",
 ] }
 kubewarden-policy-sdk = "0.17.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -32,9 +32,9 @@ tokio = { version = "^1", features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["fmt"] }
 url = { version = "2.5", features = ["serde"] }
-wasi-common = "45.0"
+wasi-common = "46.0.1"
```
flavio: I would prefer to stick to `46.0` to be more flexible

Cargo.toml
```diff
@@ -32,9 +32,9 @@ tokio = { version = "^1", features = ["full"] }
...
 # We need to disable the default features here to be able to provide a specific
 # set of features for kwctl on macOS x86_64
 # Workaround for https://github.com/bytecodealliance/wasmtime/issues/12217
-wasmtime      = { version = "45.0", default-features = false }
-wasmtime-wasi = "45.0"
+wasmtime      = { version = "46.0.1", default-features = false }
```
flavio: I would prefer to stick to `46.0` to be more flexible

flavio (changes requested):

PR #1858, opened 2026-07-02
feat(migration script): Perform helm upgrade after install

### LINKED ISSUE #1857: Migration script: After migration, old resources still have resource-policy keep
Performing the migration, with the following values:

The migration was succesful, but all the old resources that got the annotation  `helm.sh/resource-policy: keep` still have them.
This seems to happen because on `helm install --take-ownership`, helm still doesn't own the resources, so it cannot remove the annotations that it doesn't have.

#### Possible solution

Performing a second normal `helm upgrade unified-chart --reuse-values`, after the one with `helm install --take-ownership`  will remove the leftover keep annotations for those resources that are owned by Helm.
It will not remove the annotations from resources not managed by helm anymore of course (recommended policies and default policyserver), but as listed in https://github.com/kubewarden/adm-controller/issues/1853, they do not take effect anyways.

viccuad:
## Description

Fix https://github.com/kubewarden/adm-controller/issues/1857

One of the possible approaches for that issue.

This ensures that `helm.sh/resource-policy: keep` annotations are removed for those resources owned by Helm. This doesn't include the default PolicyServer and the recommended policies, as those aren't owned by Helm (and the annotations don't take effect).

## Test

## Checklist

### DIFF
--- a/charts/admission-controller/kubewarden-unified-adm-controller-chart-migration.sh
+++ b/charts/admission-controller/kubewarden-unified-adm-controller-chart-migration.sh
@@ -43,7 +43,11 @@
 #      are used) preserves the user's prior configuration. The migrated release is vanilla
 #      — it stores no overrides, so future upgrades need no special values.
 #
-#   5. Post-migration verification
+#   5. Upgrade unified chart
+#      Now that Helm owns the release, doing a simple upgrade --keep-values
+#      removes the leftover injected `helm.sh/resource-policy: keep` annotations.
+#
+#   6. Post-migration verification
 #      Confirms CRDs are owned by the new release, the controller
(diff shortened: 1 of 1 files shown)


viccuad:
I'm totally OK not merging this PR: the annotations that this PR removes by doing the additional `helm upgrade` will go away anyways when someone upgrades the charts in the future, and it adds noise to the whole migration.

jvanz (approved):
LGTM. I agree that it's nice to have as clean as possible cluster after the migration. Thus, I'll merge this even if we have all this discussion about change the script overall. I think it does not hurt and if the noise it's too much, we can easily remove the function that do the additional upgrade. ;)

PR #1859, opened 2026-07-02
fix(audit-scanner): delete stale reports by write-set, not label selector

### LINKED ISSUE #1838: Audit scanner job produces different results each run
Each audit-scanner job run produces different number of reports. I randomly get:
 - 1 reps, 3 creps
 - 2 reps, 5 creps
 - 2 reps, 3creps
 - ...

To reproduce:
```bash
# I watch generated reports
kubectl get reps,creps -A

# install and run audit scanner test - test should fail around line 123
make clean install audit-scanner.bats KEEP=1

# I run this command repeatedly to produce audit reports
jobname=auditjob
NAMESPACE=kubewarden
kubectl create job --from=cronjob/audit-scanner "$jobname" --namespace "$NAMESPACE" | grep "$jobname created"; \
kubectl wait --timeout=3m --for=condition="Complete" job "$jobname" --namespace "$NAMESPACE"; \
kubectl delete job "$jobname" --namespace "$NAMESPACE"
```

flavio:
The GC step deleted reports with a DeleteAllOf call using a "run-uid != current" label selector. That selector is evaluated against the API server's watch cache, which can lag behind patches committed in the same scan run. As a result, reports that were just updated with the current run UID could still appear stale to the delete, causing them to be wrongly removed.

This commit changes how the deletion of the stale reports happens.

Each auditResource/auditClusterResource goroutine records the report name in a concurrency-safe set after a successful CreateOrPatch. DeleteOldReports/DeleteOldClusterReports then list managed reports and delete only those whose name is absent from that set, using per-item Delete with UID+ResourceVersion preconditions for optimistic concurrency.

I've tested the fix by running the broken e2e test locally, everything seems to be fine now.

Fixes: https://github.com/kubewarden/adm-controller/issues/1838

### DIFF
--- a/charts/admission-controller/templates/controller/audit-scanner-role.yaml
+++ b/charts/admission-controller/templates/controller/audit-scanner-role.yaml
@@ -33,6 +33,7 @@ rules:
     - clusterpolicyreports
   verbs:
     - create
+    - delete
     - deletecollection
     - get
     - list
@@ -46,6 +47,7 @@ rules:
     - clusterreports
   verbs:
     - create
+    - delete
     - deletecollection
     - get
     - list
--- a/internal/audit-scanner/report/legacy_migration.go
+++ b/internal/audit-scanner/report/legacy_migration.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"log/slog"
 
-	"github.com/google/uuid"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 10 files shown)


jvanz (approved):
LGTM. I'm still wondering with there is a way to reduce the amount of API server requests. But I cannot think in nothing now. I'll try tomorrow again. But for now I think we are good and we can iterate over it later if necessary.

flavio:
I've sent some additional commit to this PR, they are mostly optimizations and one bug fix:

- (perf) b5be0fdc5468cc8b4a09e4e9212bc11ddc747600: when listing the reports to delete, fetch only metadata instead of the full object. This reduces the load on the API server and the resource usage on the client
- (fix) f8b6331850804160be912ae27e0568437e4a7586: it turns out the migration code we had in place had a bug. The old namespaced reports were deleted with a single deleteCollection request, providing an `""` namespace. The k8s API doesn't offer an API that allows the deletion of a namespaced-type across all the namespaces, you always have to provide the name of the namespace. I tested that on a real cluster and it's true. The fakeclient tests were passing because the implementation of fake client diverges from the real one
- (perf) 17653aaead18c4219e563f67c3b18a32d436fcce: small one, when all the reports inside of a namespace have to be deleted (because for example there are no policies that apply to the resources inside of that namespace) -> issue a deleteCollection request instead of `N` delete resource requests

flavio:
BTW: the e2e-test is failing because it's not using the audit-scanner from this PR, it uses the one with the `latest` tag. Everything is going to be green once we merge this PR

jvanz (changes requested):
I think we need to make two changes in this PR:

1. Use the resource version in the delete operation: this ensures that the API server will not fetch the wrong version of the resource when performing the delete.
2. I think the scanner can delete all the reports at once converting the set into [selectors](https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#SelectorFromSet) and passing it to the DeleteAllOf call: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#DeleteAllOfOption

I did not test the delete options yet. But it seems doable. On top of my head, I think if this is doable, we could remove the audit scanner UID and keep working only with labels and resource version.

viccuad:
LGTM.

What concerns me is the possible increase in memory usage. Maybe it is worth to paginate `deleteReportsNotInSet()`. But I don't see easy gains for memory anywhere else, and the rest seems fine.

>     1. Use the resource version in the delete operation: this ensures that the API server will not fetch the wrong version of the resource when performing the delete.

Does it matter? It is getting deleted anyways because no policy is touching the report. 


>     2. I think the scanner can delete all the reports at once converting the set into [selectors](https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#SelectorFromSet) and passing it to the DeleteAllOf call: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#DeleteAllOfOption

I cannot think of a labelSelector or fieldSelector that can pick reports not in the `kept` set. I can only think of a long label `notin` with a very long list of report names (uids) but that's not practical.

jvanz (approved):
I still think we should validate the resourceVersion idea. But this solve our issues for new and we have other fishes to fry. So, I would say that LGTM. We can go back to this later.

viccuad (approved):

PR #1860, opened 2026-07-02
fix: recommended policies should always have the "backgroundAudit" set to true

### LINKED ISSUE #1851: Recommended policies default to BACKGROUNDAUDIT=false
Recommended policies are skipped by audit scan. Default value changed to `false` in 1.37-alpha.

```bash
# Admission Controller 1.36
k get cap,ap -A
NAME                                                                      POLICY SERVER   MUTATING   BACKGROUNDAUDIT   MODE      OBSERVED MODE   STATUS    AGE
clusteradmissionpolicy.policies.kubewarden.io/do-not-run-as-root          default         true       true              monitor   monitor         pending   10s
clusteradmissionpolicy.policies.kubewarden.io/do-not-share-host-paths     default         false      true              monitor   monitor         pending   10s
clusteradmissionpolicy.policies.kubewarden.io/drop-capabilities           default         true       true              monitor   monitor         pending   10s
clusteradmissionpolicy.policies.kubewarden.io/no-host-namespace-sharing   default         false      true              monitor   monitor         pending   10s
clusteradmissionpolicy.policies.kubewarden.io/no-privilege-escalation     default         true       true              monitor   monitor         pending   10s
clusteradmissionpolicy.policies.kubewarden.io/no-privileged-pod           default         false      true              monitor   monitor         pending   10s

# Admission Controller 1.37.0-alpha
k get cap,ap -A
NAME                                                                      POLICY SERVER   MUTATING   BACKGROUNDAUDIT   MODE      OBSERVED MODE   STATUS    AGE
clusteradmissionpolicy.policies.kubewarden.io/do-not-run-as-root          default         true       false             monitor   monitor         pending   4s
clusteradmissionpolicy.policies.kubewarden.io/do-not-share-host-paths     default         false      false             monitor   monitor         pending   4s
clusteradmissionpolicy.policies.kubewarden.io/drop-capabilities           default         true       false             monitor   monitor         pending   4s
clusteradmissionpolicy.policies.kubewarden.io/no-host-namespace-sharing   default         false      false             monitor   monitor         pending   4s
clusteradmissionpolicy.policies.kubewarden.io/no-privilege-escalation     default         true       false             monitor   monitor         pending   4s
clusteradmissionpolicy.policies.kubewarden.io/no-privileged-pod           default         false      false             monitor   monitor         pending   4s
```

jvanz:
## Description

The recommended policies are rendered from the defaults ConfigMap and applied by the controller. When the template left `backgroundAudit` unset, the value reaching the API server was false instead of the CRD default of true, so the audit scanner skipped those policies. This was a regression from 1.36, where the policies were audited by default.

This PR fixes that in two ways: first, we add the missing flag in the Helm chart. This ensures that the field will always be present and set to `true`. The second fix is in the controller where it now uses server-side apply to update the resources. This way, we ensure that default values defined in the CRDs will be applied when the field is not defined. This will work not only for the `backgroundAudit` flag but for any field that we can add in the future with no need to change the code. 

Fix #1851

### DIFF
--- a/internal/controller/defaults_applier.go
+++ b/internal/controller/defaults_applier.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/go-logr/logr"
@@ -11,14 +10,13 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/sets"
+	sigsyaml "sigs.k8s.io/yaml"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
-	policiesv1 "github.com/kubewarden/adm-controller/api/policies/v1"
 	"github.com/kubewarden/adm-controller/internal/constants"
 )
 
@@ -35,7 +33,6 @@ type DefaultsApplierReconciler struct {
 	Log                  logr.Logger
 	DeploymentsNamespace string
 	ConfigMapName        string
-	decoder              runtime.Decoder
 }
 
 // Reconcile watches the defaults ConfigMap and applies the resources it contains.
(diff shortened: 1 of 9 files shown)


flavio (approved):

internal/controller/defaults_applier.go
```diff
@@ -113,96 +103,53 @@ func (r *DefaultsApplierReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-	_, err := controllerutil.CreateOrPatch(ctx, r.Client, desired, func() error {
-		copySpec(desiredCopy, desired)
-		applyManagedMetadata(desiredCopy, desired)
+	// Stamp the ownership label used by cleanupStale to identify managed resources.
+	labels := desired.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[constants.DefaultsManagedByLabelKey] = constants.DefaultsManagedByLabelValue
```
viccuad, line 122: This label was always added, and previously it was obtained from the resource definitions on the Helm ConfigMap.
I'm ok adding it here, but then, I would remove it from the Helm ConfigMap.

jvanz: Actually, the label was always added even in the old code. See the function `applyManagedMetadata` , line 197 in the removed code in this PR..

viccuad: You are right! Resolving.

I confused
  `constants.DefaultsManagedByLabelKey` (`kubewarden.io/managed-by"`)
with
  `constants.ManagedByKey` (`"app.kubernetes.io/managed-by"`)

viccuad (approved):
LGTM!

internal/controller/defaults_applier.go
```diff
@@ -113,96 +103,53 @@ func (r *DefaultsApplierReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		*policiesv1.AdmissionPolicy,
-		*policiesv1.ClusterAdmissionPolicyGroup,
-		*policiesv1.AdmissionPolicyGroup:
+// isSupportedGVK returns true if the GroupVersionKind is a known Kubewarden resource type.
+func isSupportedGVK(gvk schema.GroupVersionKind) bool {
+	if gvk.Group != constants.KubewardenPoliciesGroup {
+		return false
+	}
+	switch gvk.Kind {
+	case "PolicyServer",
```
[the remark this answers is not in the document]
jvanz: Yeah, that's makes sense. But this is not a issue that we will should face in the near future. So, as we have bigger fish to fry I'll let this for the future as a enhancement.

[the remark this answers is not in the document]
jvanz: @copilot can you create an enhancement issue in this repository for this?

PR #1863, opened 2026-07-03
fix(audit scanner): Disable store side effect

flavio:
> WARNING: built on top of https://github.com/kubewarden/adm-controller/pull/1859

GC ran unconditionally regardless of --disable-store, so a "read-only" scan still deleted every managed (Cluster)Report and legacy wgpolicy PolicyReport in scope.

Now, when --disable-store is used, no changes are actually done to etcd contents.


viccuad (commented):
LGTM, pending the parent PR.

PR #1875, opened 2026-07-09
fix(chart): Delete all CRs on helm uninstall

### LINKED ISSUE #1849: Helm chart unsintall should keep only custom resource defintions and their CRDs
Currently, when the users uninstall the Kubewarden admission controller Helm chart, all the policy servers deployments for their custom policies are kept running. This means that after the uninstall, Kubewarden admissions controller is gone. But the policies and policy server are still running and evaluating resources. No reconciliation will be done.

This is caused because in recent [change](https://github.com/kubewarden/adm-controller/pull/1724) we change on how the controller handle the policies when we delete policy servers. In the past, when a policy server was deleted, all the policies attached to it were removed as well. This is no long true. Furthermore, now the [pre-delete hook](https://github.com/kubewarden/adm-controller/blob/a517dfa033ab4accc51a6d3bb8d69da2a964f3d4/charts/admission-controller/templates/controller/pre-delete-hook.yaml#L35) delete only the default resources created on the Helm chart installation. The legacy Helm chart also remove all the [policy servers](https://github.com/kubewarden/helm-charts/blob/a9cd2fb5725e7bf4507b5aa0f7b046c6793cf76d/charts/kubewarden-controller/templates/pre-delete-hook.yaml#L39)

The helm chart should be updated to ensure that when the Helm chart is uninstalled, **only** the policy server and policies kinds definition are kept. All the infrastructure and deployments to make the policies to run should be deleted. Therefore, when the adm-controller is reinstalled, the controller reconcile the policies and policy server left from the previous uninstall. 

## Acceptance criteria
- Update the Helm chart to ensure that after the uninstall **only** the policy server and policies kinds and their CRDs are keep installed in the cluster. 
- Ensure that the re-installation after the uninstall continue to work
- Add tests to cover this scenario

viccuad:
## Description

Fix https://github.com/kubewarden/adm-controller/issues/1849 (solution C)

Expand the existing helm pre-delete-hook to delete all CRs across all
namespaces.

This matches the behavior of 1.36.

## Test

chart e2e tests.

## Checklist

### DIFF
--- a/internal/controller/policyserver_controller_configmap.go
+++ b/internal/controller/policyserver_controller_configmap.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -165,9 +166,15 @@ func (r *PolicyServerReconciler) updateConfigMapData(cfg *corev1.ConfigMap, poli
 	}
 
 	cfg.Data = data
-	cfg.ObjectMeta.Labels = map[string]string{
-		constants.PolicyServerLabelKey: policyServer.ObjectMeta.Name,
+	labels := map[string]string{
+		// PolicyServerLabelKey is used to map ConfigMap events to policy reconcile
+		// requests
+		constants.PolicyServerLabelKey: policyServer.Name,
 	}
+	// Use the same labels as other resources backing the PolicyServer, to track
+	// for deletion later on
+	maps.Copy(labels, policyServer.CommonLabels())
+	cfg.Labels = labels
 	if err = controllerutil.SetOwnerReference(policyServer, cfg, r.Client.Scheme()); err != nil {
 		return errors.Join(errors.New("failed to set policy server configmap owner reference"), err)
 	}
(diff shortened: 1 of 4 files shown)


internal/controller/policyserver_controller_configmap.go
```diff
@@ -165,9 +166,15 @@ func (r *PolicyServerReconciler) updateConfigMapData(cfg *corev1.ConfigMap, poli
...
-		constants.PolicyServerLabelKey: policyServer.ObjectMeta.Name,
+	labels := map[string]string{
+		// PolicyServerLabelKey is used to map ConfigMap events to policy reconcile
+		// requests
+		constants.PolicyServerLabelKey: policyServer.Name,
 	}
+	// Use the same labels as other resources backing the PolicyServer, to track
+	// for deletion later on
+	maps.Copy(labels, policyServer.CommonLabels())
+	cfg.Labels = labels
```
jvanz, line 177: I'm little confused. Why do we need this label if you do not used in the delete hook in this same PR? Furthermore, the configmap has the owner ref set to the policy server (see the next line). Thus, the label is not required. Right?

jvanz (changes requested):

internal/controller/policyserver_controller_configmap.go
viccuad: correct, this PR was built on top of the other branch with the ConfigMap labels, as I saw that as needed regardless.

My fault, let me put this PR as draft until I have both solutions opened as PRs.

flavio:
Closing, we're going to do that by extending our controller binary to have a `--cleanup` flag.

PR #1879, opened 2026-07-10
feat!(controller,helm): Implement `controller --cleanup` and consume it in `pre-delete-hook`

### LINKED ISSUE #1849: Helm chart unsintall should keep only custom resource defintions and their CRDs

viccuad:
## Description

Fix https://github.com/kubewarden/adm-controller/issues/1849 (via solution B)

Breaking change new in v1.37: Kubewarden CRs will be left around after `helm uninstall`. CRDs were already left around prior to this PR.

- Add a new `controller --cleanup --deployments-namespace=<release ns> --controller-deployment-name=<controller deployment name>`
  Implement the cleanup logic run by the Helm chart `pre-delete-hook` when the Kubewarden stack is uninstalled.

  The uninstall leaves behind ONLY the Kubewarden CRDs and the user-defined PolicyServer/policy CRs, without their Kubewarden finalizers, so that:
  - nothing keeps evaluating admission requests without a controller reconciling it,
  - a future re-installation of the chart can adopt and reconcile the kept custom resources again,
  - the kept custom resources (and their CRDs) can be deleted with plain kubectl commands, without a running controller honoring finalizers.

- Fix corner case with new behavior: 
  Historically, we had a MutatingWebhookConfiguration that used a Defaulter to add a Finalizer to the PolicyServers. This is not enough if users exercise this PR feature, or they are migrating, as  PolicyServers aren't CREATEd/UPDATEd by the controller but exist prior to the installation of the controller (so the MutatingWebhook doesn't trigger). Therefore, set their Finalizer in the PolicyServer Reconciler.
  - The PolicyServer Defaulter and MutatingWebhookConfiguration was only used for adding that Finalizer. Drop it entirely now.

## Test

Added unit tests and helm tests.
Tried some Go e2e tests, but discarded them as they were heavy:
- They take 9-11min, which runs over the default timeout of 10mins
- I needed to increase inotify max_user_instances and max_user_watches, to run the e2e tests + 1 unrelated cluster in my machine

## Checklist

### DIFF
--- a/charts/admission-controller/README.md
+++ b/charts/admission-controller/README.md
@@ -367,18 +367,19 @@ helm install <release> <chart> -n <namespace> --take-ownership
 helm uninstall kubewarden-controller -n kubewarden
 ```
 
-This removes:
-
-- The controller Deployment
-- Managed defaults (resources labeled `kubewarden.io/managed-by=kubewarden-controller-defaults`)
-- ConfigMaps, Secrets, Services
+This removes the controller Deployment, default PolicyServer, recommended
+policies, and everything that makes the policies run, including the regular
+chart resources (RBAC, Secrets, Services, ConfigMaps, and so on).
 
 It does not remove:
 
-- CRDs (kept by `helm.sh/resource-policy: keep`)
-- User-managed PolicyServers and policies
+- CRDs (kept by `helm.sh/resource-policy: keep`).
+- User-managed PolicyServer and policy custom resources.
 
-To remove CRDs after uninstall:
+The user-managed custom resources are kept on purpose. Re-installing the chart
+deploys the controller, which reconciles them, and activates them back.
+
+To remove all user-managed custom resources and the CRDs do:
 
 ```sh
 kubectl delete crd policyservers.policies.kubewarden.io
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -48,6 +48,7 @@ import (
 
 	policiesv1 "github.com/kubewarden/adm-controller/api/policies/v1"
 	"github.com/kubewarden/adm-controller/api/policies/v1alpha2"
+	"github.com/kubewarden/adm-controller/internal/cleanup"
 	"github.com/kubewarden/adm-controller/internal/constants"
 	"github.com/kubewarden/adm-controller/internal/controller"
 	"github.com/kubewarden/adm-controller/internal/featuregates"
@@ -117,6 +118,8 @@ func main() {
 	var openTelemetryClientCertificateSecret string
 	var openTelemetryCertificateSecret string
 	var imagePullSecretsFlag string
+	var cleanupMode bool
+	var controllerDeploymentName string
 
 	flag.StringVar(&mgrOpts.MetricsAddr, "metrics-bind-address", ":8088", "The address the controller-runtime metric endpoint binds to.")
 	flag.StringVar(&mgrOpts.ProbeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
@@ -160,6 +163,18 @@ func main() {
 			"WARNING: enabling this increases the attack surface by exposing webhook endpoints "+
 			"on the host network and giving pods visibility of all node network interfaces. "+
 			"Use only when the Kubernetes API server cannot reach pod-network webhook endpoints.")
+	flag.BoolVar(&cleanupMode,
+		"cleanup",
+		false,
+		"Run in cleanup mode instead of starting the controller manager. "+
+			"Used by the Helm chart pre-delete hook: deletes the controller Deployment, "+
+			"the Kubewarden webhook configurations, the chart-managed default resources and "+
+			"the resources backing the policy servers, and strips the Kubewarden finalizers "+
+			"from the kept Kubewarden custom resources.")
+	flag.StringVar(&controllerDeploymentName,
+		"controller-deployment-name",
+		"",
+		"The name of the controller Deployment to delete in cleanup mode.")
 
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
@@ -167,6 +182,14 @@ func main() {
 	mgrOpts.EnableMutualTLS = config.ClientCAConfigMapName != ""
 	config.ImagePullSecrets = parseImagePullSecrets(imagePullSecretsFlag)
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	if cleanupMode {
+		if err := runCleanup(mgrOpts.DeploymentsNamespace, controllerDeploymentName); err != nil {
+			setupLog.Error(err, "cleanup failed")
+			retcode = 1
+		}
+		return
+	}
 
 	// Validate --webhook-server-port range.
 	if int64(mgrOpts.WebhookServerPort) < minAllowedPort || int64(mgrOpts.WebhookServerPort) > maxAllowedPort {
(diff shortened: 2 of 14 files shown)


internal/cleanup/cleanup.go
```diff
@@ -0,0 +1,309 @@
...
+		&appsv1.DeploymentList{},
+		&corev1.ServiceList{},
+		&corev1.SecretList{},
+		&corev1.ConfigMapList{},
+		&policyv1.PodDisruptionBudgetList{},
+	} {
+		if err := deleteEachListItem(ctx, c, list, inNamespace, partOfKubewarden); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
viccuad: This is valid. Fixed in https://github.com/kubewarden/adm-controller/pull/1879/commits/9a20606264ba277fe9cb6f80b00172cf2d7e66e2, making sure that `controller --cleanup` only deletes the Deployment, webhooks, and PolicyServers stack, but leaves ca, certs. This ensures a call of `controller --cleanup` doesn't destroy the stack, and one must perform a full helm uninstall.

cmd/controller/main.go
```diff
@@ -168,6 +183,14 @@ func main() {
 	config.ImagePullSecrets = parseImagePullSecrets(imagePullSecretsFlag)
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	if cleanupMode {
+		if err := runCleanup(mgrOpts.DeploymentsNamespace, controllerDeploymentName); err != nil {
+			setupLog.Error(err, "cleanup failed")
+			retcode = 1
+		}
+		return
+	}
```
jvanz, line 192: I do not consider this an actual blocker. But my preference would be having a new command file just for the cleanup. But as this also required more building time work. I'm fine to keep it in the controller code.

charts/admission-controller/README.md
```diff
@@ -367,18 +367,19 @@ helm install <release> <chart> -n <namespace> --take-ownership
...
-- CRDs (kept by `helm.sh/resource-policy: keep`)
-- User-managed PolicyServers and policies
+- CRDs (kept by `helm.sh/resource-policy: keep`).
+- User-managed PolicyServer and policy custom resources.
+
+The user-managed custom resources are kept on purpose. Re-installing the chart
+deploys the controller, which reconciles them, and activates them back.
 
-To remove CRDs after uninstall:
+To remove all user-managed custom resources and the CRDs do:
```
jvanz, line 382: This reminder me that during the issue investigation you list several documentation pages about uninstall procedures. I think we should also update the [uninstall section](https://docs.kubewarden.io/admission-controller/1.37/en/quick-start.html#_uninstall) from the quickstart for the kubewarden admission controller docs website with this information as well.

internal/cleanup/cleanup.go
```diff
@@ -0,0 +1,321 @@
...
+//nolint:gochecknoglobals // overridden by the tests
+var (
+	// controllerPodsGoneTimeout is the maximum time to wait for the
+	// controller pods to be fully terminated after their Deployment is
+	// deleted.
+	controllerPodsGoneTimeout = 3 * time.Minute
+	// controllerPodsGonePollInterval is the poll interval used while waiting
+	// for the controller pods to be gone.
+	controllerPodsGonePollInterval = 1 * time.Second
+)
```
jvanz, line 49: For most of the cases, I think this would be fine. But I'm wondering what could happen with some user make the timeout for the helm command shorter. Therefore, there is the possibility to the helm command to exit in error before the hook finishes. Should not be safer to allow configure this values?

internal/cleanup/cleanup.go
```diff
@@ -0,0 +1,321 @@
...
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return fmt.Errorf("failed to get the object: %w", err)
+		}
+
+		changed := controllerutil.RemoveFinalizer(obj, constants.KubewardenFinalizer)
+		if controllerutil.RemoveFinalizer(obj, constants.KubewardenFinalizerPre114) {
+			changed = true
+		}
```
jvanz, line 210:
```suggestion
		changed := controllerutil.RemoveFinalizer(obj, constants.KubewardenFinalizer) || controllerutil.RemoveFinalizer(obj, constants.KubewardenFinalizerPre114) 
```

jvanz (commented):
LGTM. I'm not dropping an approval here just because I would like to discuss a little bit the timeout configuration first. Even if that question, I think the code looks good enough to be merged already.

jvanz:
> Added unit tests and helm tests. Tried some Go e2e tests, but discarded them as they were heavy:
> 
>     * They take 9-11min, which runs over the default timeout of 10mins
> 
>     * I needed to increase inotify max_user_instances and max_user_watches, to run the e2e tests + 1 unrelated cluster in my machine

I think we must have some e2e test somewhere. If not in the go e2e test, we should add it in the kubewarden-end-to-end-tests repository.

cmd/controller/main.go
viccuad: I wanted it to be part of the same binary to reuse the controller image. We can do a second binary, but that means adapting the CIs, airgap story, etc.

internal/cleanup/cleanup.go
viccuad: It's possible, but the hook is configured to retrigger on error, so it will try again and again. Each time, more things are deleted, so the run time is shorter, so at the end it should succeed.

internal/cleanup/cleanup.go
viccuad: I think we need to ensure that both Finalizers are gone if they were present (or the resource would not be garbage collected), it's not enough to just have an OR.

jvanz (approved):

flavio (approved):
LGTM, we can do the e2e tests as part of another PR

charts/admission-controller/README.md
viccuad: Opened https://github.com/rancher/kubewarden/issues/68, working on it.

viccuad:
Opened an [issue for the docs](https://github.com/rancher/kubewarden/issues/68) and another one for the [e2e tests](https://github.com/rancher/kubewarden/issues/69).

PR #1952, opened 2026-08-21
refactor(policy-evaluator): Move from cached to moka, free mem automatically

### LINKED ISSUE #1950: policy-server memory grows unbounded with context-aware policies due to unbounded cached TtlCache

- [x] I have searched the existing issues

### Current Behavior

`policy-server`'s memory (RSS) grows steadily on every admission evaluation and never comes back down. In a cluster running context-aware policies, it climbed to ~22 GB overnight, eventually causing OOM.

The growth comes from two in-memory caches used by the context-aware Kubernetes host calls, in `crates/policy-evaluator/src/callback_handler/kubernetes.rs`:
- `get_resource_cached` (`#[cached(ttl = 5, …)]`) : keyed by resource name and namespace, value is the full object.
- `can_i_cached` (`#[cached(ttl = 5, …)]`): keyed by the whole `SubjectAccessReview`.

With the cached crate (2.0.2), `ttl` without `max_size` selects an unbounded `TtlCache`: an expired entry is only removed if that same key is looked up again, a key seen once and never again stays in the map forever, and there is no size limit. Since context-aware policies look up a constant stream of distinct resources / subjects (pods with random suffixes, etc.), the cache grows without bound. `ttl = 5` only makes a re-queried value refresh, it does not bound memory.

### Expected Behavior

policy-server memory should stay bounded during long-running operation regardless of how many distinct resources context-aware policies inspect. Caches should have an upper size limit and evict old entries.

### Steps To Reproduce

1. Install Kubewarden on a cluster
2. Enable heap profiling on the policy server: set `KUBEWARDEN_ENABLE_PPROF=true` via `PolicyServer.spec.env` (exposes `/debug/pprof/heap`).
3. Deploy a context-aware policy that calls `get_resource`/`can_i` with keys thatvary per request , e.g. the published `ghcr.io/kubewarden/tests/context-aware-test-policy`, which calls `can_i` with a user derived from the incoming object.
4. Send a stream of admission requests with distinct values (e.g. Deployments each with a unique `serviceAccountName`), using server side dry-run so nothing is persisted.
5. Watch `kubectl top pod` RSS climb and `GET /debug/pprof/heap` grow, monotonically, with no recovery.

### Environment

```markdown
- OS: Linux
- Architecture: amd64
- Kubernetes: kind (kindest/node v1.36)
- Kubewarden policy-server: v1.36.0 (policy-evaluator v0.32.0, cached crate 2.0.2)
```

### Anything else?

The same unbounded `TtlCache` pattern (ttl without max_size) is used by 10 `#[cached]` functions:
- `callback_handler/kubernetes.rs`: `get_resource_cached`, `can_i_cached` , request-varying keys, large values.
- `callback_handler/oci.rs`: 3 caches keyed by image ref.
- `callback_handler/sigstore_verification.rs`: 5 caches, keyed by verification params.

As a fix (which I tried and works), I'd suggest adding `max_size = N` to each `#[cached]` block. In cached 2.0 `max_size` + `ttl` selects `LruTtlCache` (LRU + TTL), which limits the entry count and evicts LRU entries, bounding memory while keeping the freshness window.

I attached a zip with the whole setup to reproduce. Also a txt in case you don't trust the zip.
[policy-server-memleak-evidence.zip](https://github.com/user-attachments/files/31173211/policy-server-memleak-evidence.zip)
[heap-evidence.txt](https://github.com/user-attachments/files/31173243/heap-evidence.txt)

viccuad:
## Description

Fix #1950

The cached crate does not free memory when the TTL passes. It removes an expired entry only when the same key is queried again. Context-aware policies query many distinct keys, thus the caches grow without limit.

Replace the `#[cached]` macros with moka caches. moka removes expired entries during routine cache operations. As a result, the memory usage follows the current request rate, and old entries do not accumulate.
The caches have no entry-count bound. A bound is not necessary, because the TTL windows are short (5 s and 60 s) and moka removes the expired entries without a new query for the same key.

The behavior does not change:
- Only successful results enter the cache.
- Concurrent misses for the same key make one backend call.
- The was_cached flag comes from Entry::is_fresh().
- The TTL windows and the key expressions stay the same.

## Test

No new tests are needed for the wrappers.
Added new test to see if the cache static approach is being exercised for `can_i_cached()`.

Tested locally.
Testing performance changes (see issue comments), hence why draft request. 

### Tradeoff

The cached macro held one global lock per cache, and a miss held this lock for the full backend call. moka uses a lock-free hash table, thus hits do not wait, and misses for different keys run in parallel. As a cost, moka does its housekeeping on the caller threads in small batches, and each entry uses more metadata bytes than a plain hash map entry.
These costs are small compared to the network round-trips of the functions we cache.

### Potential improvement

We could set entry count bound + weighted counts on moka (as we could on cached). This would make hot entries survive more. But there's no 1-size-fits-all counts per function. The best default I can think of is to put as highest default counters as possible, to ensure the cache gets exercised in big production cluster. 

## Checklist

### DIFF
--- a/crates/policy-evaluator/src/callback_handler/cache_return.rs
+++ b/crates/policy-evaluator/src/callback_handler/cache_return.rs
@@ -12,10 +12,18 @@ pub(crate) struct Return<T> {
 
 impl<T> Return<T> {
     /// Wraps a value that did not come from the cache.
-    pub fn new(value: T) -> Self {
+    pub fn not_cached(value: T) -> Self {
         Return {
             value,
             was_cached: false,
+        }
+    }
+
+    /// Wraps a value that came from the cache.
+    pub fn cached(value: T) -> Self {
+        Return {
+            value,
+            was_cached: true,
         }
     }
 }
@@ -40,9 +48,10 @@ where
         .await
         .map_err(|e| anyhow!("{e:#}"))?;
 
-    Ok(Return {
-        was_cached: !entry.is_fresh(),
-        value: entry.into_value(),
+    Ok(if entry.is_fresh() {
+        Return::not_cached(entry.into_value())
+    } else {
+        Return::cached(entry.into_value())
     })
 }
 
--- a/crates/policy-evaluator/src/callback_handler/cache_return.rs
+++ b/crates/policy-evaluator/src/callback_handler/cache_return.rs
@@ -1,0 +1,31 @@
+/// Wraps a callback handler function result. The `was_cached` flag shows if the value came from the
+/// cache.
+#[derive(Debug, Clone)]
+pub(crate) struct Return<T> {
+    pub value: T,
+    pub was_cached: bool,
+}
+
+impl<T> Return<T> {
+    /// Wraps a value that did not come from the cache.
+    pub fn new(value: T) -> Self {
+        Return {
+            value,
+            was_cached: false,
+        }
+    }
+}
+
+impl<T> std::ops::Deref for Return<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        &self.value
+    }
+}
+
+impl<T> std::ops::DerefMut for Return<T> {
+    fn deref_mut(&mut self) -> &mut T {
+        &mut self.value
+    }
+}
--- a/crates/policy-evaluator/src/callback_handler/cache_return.rs
+++ b/crates/policy-evaluator/src/callback_handler/cache_return.rs
@@ -1,3 +1,7 @@
+use std::{future::Future, hash::Hash};
+
+use anyhow::{Result, anyhow};
+
 /// Wraps a callback handler function result. The `was_cached` flag shows if the value came from the
 /// cache.
 #[derive(Debug, Clone)]
@@ -14,6 +18,32 @@ impl<T> Return<T> {
             was_cached: false,
         }
     }
+}
+
+/// Looks up `key` inside of `cache`. On a miss, `init` is awaited to compute the value, which is
+/// then stored inside of the cache. Only successful results enter the cache.
+///
+/// This centralizes the moka `entry().or_try_insert_with()` dance used by all the caches owned by
+/// the callback handler's clients (Kubernetes, OCI registry, Sigstore verification).
+pub(crate) async fn try_cached<K, V>(
+    cache: &moka::future::Cache<K, V>,
+    key: K,
+    init: impl Future<Output = Result<V>>,
+) -> Result<Return<V>>
+where
+    K: Hash + Eq + Send + Sync + Clone + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    let entry = cache
+        .entry(key)
+        .or_try_insert_with(init)
+        .await
+        .map_err(|e| anyhow!("{e:#}"))?;
+
+    Ok(Return {
+        was_cached: !entry.is_fresh(),
+        value: entry.into_value(),
+    })
 }
 
 impl<T> std::ops::Deref for Return<T> {
--- a/crates/policy-evaluator/src/callback_handler/crypto.rs
+++ b/crates/policy-evaluator/src/callback_handler/crypto.rs
@@ -85,54 +85,38 @@ fn verify_cert_chain(
     );
 
     match verification_result {
-        Ok(_) => Ok(Return {
-            value: CertificateVerificationResponse {
-                trusted: true,
-                reason: "".to_string(),
-            },
-            was_cached: false,
-        }),
-        Err(Error::InvalidSignatureForPublicKey) => Ok(Return {
-            value: CertificateVerificationResponse {
+        Ok(_) => Ok(Return::not_cached(CertificateVerificationResponse {
+            trusted: true,
+            reason: "".to_string(),
+        })),
+        Err(Error::InvalidSignatureForPublicKey) => {
+            Ok(Return::not_cached(CertificateVerificationResponse {
                 trusted: false,
                 reason: CERTIFICATE_NOT_TRUSTED_BY_CHAIN.to_string(),
-            },
-            was_cached: false,
-        }),
+            }))
+        }
         Err(Error::CertExpired {
             time: _,
             not_after: _,
-        }) => Ok(Return {
-            value: CertificateVerificationResponse {
-                trusted: false,
-                reason: CERTIFICATE_USED_AFTER_EXPIRATION.to_string(),
-            },
-            was_cached: false,
-        }),
+        }) => Ok(Return::not_cached(CertificateVerificationResponse {
+            trusted: false,
+            reason: CERTIFICATE_USED_AFTER_EXPIRATION.to_string(),
+        })),
         Err(Error::CertNotValidYet {
             time: _,
             not_before: _,
-        }) => Ok(Return {
-            value: CertificateVerificationResponse {
-                trusted: false,
-                reason: CERTIFICATE_USED_BEFORE_VALIDITY.to_string(),
-            },
-            was_cached: false,
-        }),
-        Err(Error::UnknownIssuer) => Ok(Return {
-            value: CertificateVerificationResponse {
-                trusted: false,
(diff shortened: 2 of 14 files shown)


crates/policy-evaluator/src/callback_handler/kubernetes.rs
```diff
@@ -92,34 +95,54 @@ pub(crate) async fn get_resource(
...
+    let key = format!("get_resource_cached({api_version},{kind}),{name},{namespace:?}");
+    let entry = GET_RESOURCE_CACHE
+        .entry(key)
+        .or_try_insert_with(async {
+            get_resource(client, api_version, kind, name, namespace)
+                .await
+                .map(|response| response.value)
+        })
+        .await
+        .map_err(|e| anyhow!("{e:#}"))?;
```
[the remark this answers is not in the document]
viccuad: Using `?` doesn't compile, as we have `Arc<anyhow::Error>`, and `Arc` doesn't implement `Error` nor `anyhow::Error`. I could add a helper type with all the implementations, but maybe it's not worth it.

crates/policy-evaluator/src/callback_handler/oci.rs
```diff
@@ -75,69 +79,96 @@ impl Client {
...
+        .or_try_insert_with(async {
+            oci_client
+                .digest(img)
+                .await
+                .map(|digest| ManifestDigestResponse { digest })
+        })
         .await
-        .map(|digest| ManifestDigestResponse { digest })
-        .map(cached::Return::new)
+        .map_err(|e| anyhow!("{e:#}"))?;
```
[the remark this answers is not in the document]
viccuad: same as https://github.com/kubewarden/adm-controller/pull/1952#discussion_r3852867249.

crates/policy-evaluator/src/callback_handler/sigstore_verification.rs
```diff
@@ -258,109 +259,126 @@ impl Client {
...
-    client
-        .verify_public_key(image, pub_keys, annotations)
+) -> Result<Return<VerificationResponse>> {
+    let key = format!("{image}{pub_keys:?}{annotations:?}");
+    let entry = PUB_KEY_VERIFICATION_CACHE
+        .entry(key)
+        .or_try_insert_with(client.verify_public_key(image, pub_keys, annotations))
         .await
-        .map(cached::Return::new)
+        .map_err(|e| anyhow!("{e:#}"))?;
```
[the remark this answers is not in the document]
viccuad: same as https://github.com/kubewarden/adm-controller/pull/1952#discussion_r3852867249.

crates/policy-evaluator/src/callback_handler/cache_return.rs
```diff
@@ -0,0 +1,31 @@
...
+/// cache.
+#[derive(Debug, Clone)]
+pub(crate) struct Return<T> {
+    pub value: T,
+    pub was_cached: bool,
+}
+
+impl<T> Return<T> {
+    /// Wraps a value that did not come from the cache.
+    pub fn new(value: T) -> Self {
```
flavio: nit: I personally find a bit counterintuitive that this is creating something with `was_cached` set to `false`. I would rename it to `not_cached(value: T)` and eventually have also `cached(value: T)`.

crates/policy-evaluator/src/callback_handler/crypto.rs
```diff
@@ -1,4 +1,6 @@
 use anyhow::{Result, anyhow};
+
+use crate::callback_handler::cache_return::Return;
```
flavio, line 3: nit:

```suggestion
use crate::callback_handler::cache_return::Return;

```

crates/policy-evaluator/src/callback_handler/kubernetes.rs
```diff
@@ -1,11 +1,14 @@
...
+use std::time::Duration;
 
 mod client;
 pub(crate) mod field_mask;
 mod reflector;
 
 use anyhow::{Result, anyhow};
-use cached::macros::cached;
+
+use crate::callback_handler::cache_return::Return;
```
flavio, line 9: nit, crate imports should be the last ones

crates/policy-evaluator/src/callback_handler/kubernetes.rs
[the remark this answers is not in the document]
flavio: I think this is not worth the effort. Leave things as they are and resolve this conversation.

crates/policy-evaluator/src/callback_handler/kubernetes.rs
```diff
@@ -162,41 +185,112 @@ pub(crate) async fn has_list_resources_all_result_changed_since_instant(
...
-        .can_i(request)
-        .await
-        .map(|value| cached::Return {
-            was_cached: false,
-            value,
-        })
+    client.unwrap().can_i(request).await.map(|value| Return {
+        was_cached: false,
+        value,
+    })
```
flavio: nit: here I would refactor the code to use the `Return::not_cached` helper we defined earlier

crates/policy-evaluator/src/callback_handler/cache_return.rs
```diff
@@ -0,0 +1,31 @@
+/// Wraps a callback handler function result. The `was_cached` flag shows if the value came from the
+/// cache.
+#[derive(Debug, Clone)]
+pub(crate) struct Return<T> {
+    pub value: T,
+    pub was_cached: bool,
```
jvanz: Where is `was_cached` used?

jvanz, line 10: I wonder if we cannot remove this `was_cached` field. As far as I can see, we do not use it for anything unless debug message: https://github.com/search?q=org%3Akubewarden+was_cached&type=code.

jvanz (changes requested):

flavio (commented):
I've left some notes, then I noticed the usage of the static cache objects and investigated whether we actually needed them.

I ended up pushing d770c5dcc810cf3a599870debee3df0346c594aa which:
- moves the caches into the clients
- introduce the `try_cached` helper to remove some duplicated code
- remove the mutability requirement around some of the clients. It turned out (thanks to claude!) that we don't need that, since the attributes being accessed are behind mutexes/rwlocks. This polishes the code a bit and makes the rust types more "close to reality"

There are some small things that need to be addressed.

Great finding from community, and nice job!

I would also do a follow up to stop using `cached` inside of sigstore-rs. I don't mean we have to remove the dependency from sigstore-rs (maybe), but on the super short term (this PR, another one?) we could just avoid using the cached version of the sigstore-rs methods we consume inside of policy-fetcher.

BTW, this behavior of the `cached` crate was really surprising!

crates/policy-evaluator/src/callback_handler/cache_return.rs
viccuad: It's only used for that indeed, but I think it should stay. Debugging caching issues is difficult, and it's helpful to see if one is hitting the cache.

flavio: I would keep it

crates/policy-evaluator/src/callback_handler/kubernetes.rs
viccuad: Done in https://github.com/kubewarden/adm-controller/pull/1952/commits/effa2974f147ca708f361e8d00bbe86ad6918dd8.

crates/policy-evaluator/src/callback_handler/cache_return.rs
viccuad: Done in https://github.com/kubewarden/adm-controller/pull/1952/commits/effa2974f147ca708f361e8d00bbe86ad6918dd8.

flavio (approved):
I've reviewed the new commits you added. Everything looks good. I think this is good to be merged

jvanz (approved):

PR #1979, opened 2026-08-31
ci: Configure release-drafter to use PR tiles instead of PR labels

### LINKED ISSUE #1936: Move release-drafter from labels
Since 7.7.0, release-drafter can use the PR title to directly create the GH release draft with the changelog. See:
https://github.com/release-drafter/release-drafter/pull/1556/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

Until now, we:
1. Use release-drafter/autolabeler run in a workflow, and label PRs from the conventional commits style in PR titles. This needs permissions to modify PRs, and is disabled for PRs from forks (almost all).
1. If point 1) didn't kick in, we would manually label PRs even after merging them, but before a release.
1. release-drafter listens to changes in the PR labels and updates the GH release draft.
1. Profit.

Now we can:
1. Use only release-drafter (github/workflows/release-drafter.yml), keep .github/workflows/autolabeler.yml for labels.
1. Configure release-drafter to use conventional commit style from PR titles.
1. release-drafter will take every PR title into account to create GH release drafts, including forks.
1. Profit.

#### Acceptance criteria

Update all repos to release-drafter 7.7.0, keep `.github/workflows/autolabelel.yml`, configure `.github/release-drafter.yml` to use `categories.when.conventional` instead of `categories.when.labels`.

Repos:
- adm-controller
- sbomscanner
- sdks
- policies: Update policies repo automation that creates release-drafter.yaml for each policy

viccuad:
## Description

Fix https://github.com/kubewarden/adm-controller/issues/1936

We used `autolabeler` to, from the PR title, label the PR, and then
`release-drafter` checked the PR labels to draft the GH release.
We don't allow `autolabeler` permissions to add labels for PRs coming
for forks, for security reasons. This meant we needed to label these
PRs manually.

Now, we configure `release-drafter` to read the PR title (new in
release-drafter 7.7.0) to draft the GH release.
We also add an "Other changes" section as a catch-all for all PRs.

This means we correctly add all PRs to our GH release, no matter
if they come from forks or not.

We still run `autolabeler` to label PRs automatically, same as before.

Add workflow triggers to also run release-drafter on PR edits.

## Test

Untested.

## Checklist

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,19 +1,26 @@
 categories:
   - title: "⚠️  Breaking changes"
-    labels:
-      - "kind/major"
-      - "kind/breaking-change"
+    semver-increment: "major"
+    when:
+      conventional:
+        breaking: true
   - title: "🚀 Features"
-    labels:
-      - "kind/enhancement"
-      - "kind/feature"
+    semver-increment: "minor"
+    when:
+      conventional:
+        type: "feat"
   - title: "🐛 Bug Fixes"
-    labels:
-      - "kind/bug"
+    semver-increment: "patch"
+    when:
+      conventional:
+        type: "fix"
   - title: "🧰 Maintenance"
-    labels:
-      - "kind/chore"
-      - "area/dependencies"
+    semver-increment: "patch"
+    when:
+      conventional:
+        types: ["chore", "build", "ci", "docs", "refactor", "test"]
+  - title: "📦 Other Changes"
+    semver-increment: "patch"
 
 exclude-labels:
   - duplicate
@@ -56,22 +63,3 @@ autolabeler:
   - label: "kind/chore"
     title:
       - "chore"
-
-version-resolver:
-  major:
-    labels:
-      - "kind/major"
-      - "kind/breaking-change"
-  minor:
-    labels:
-      - "kind/minor"
-      - "kind/feature"
-      - "kind/enhancement"
-  patch:
-    labels:
-      - "kind/patch"
-      - "kind/fix"
-      - "kind/bug"
-      - "kind/chore"
-      - "area/dependencies"
-  default: patch


.github/release-drafter.yml
```diff
@@ -1,19 +1,26 @@
 categories:
   - title: "⚠️  Breaking changes"
-    labels:
-      - "kind/major"
-      - "kind/breaking-change"
+    semver-increment: "major"
+    when:
+      conventional:
+        breaking: true
```
jvanz, line 6: Is this means that when we add a breaking change the version will do a major bump? If so, I think we should do a minor bump. We do major bump in very rare cases. In the past we did minor even for breaking changes.

jvanz (changes requested):

.github/release-drafter.yml
viccuad: It's the same version behavior we had before with  `version-resolver.major.labels["kind/breaking-change","kind/major"]`. This bump is for the GH release draft, and matches the usual Chart.yaml `version`. At the end it doesn't matter much, we override the version when we tag a specific tag, and we take the last GH release draft.

.github/workflows/release-drafter.yml
```diff
@@ -6,12 +6,18 @@ on:
     # branches to consider in the event; optional, defaults to all
     branches:
       - main
+  pull_request:
+    # Refresh the release draft when a maintainer edits the title of a
+    # merged PR
+    types: [edited]
```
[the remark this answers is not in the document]
viccuad: Argh, true. Dropped the second commit that changed the `types: [edited]`.

flavio (approved):

jvanz (approved):

PR #1981, opened 2026-08-31
chore(crds): update CRDs after deps updates

jvanz:
## Description

After some deps updates the some fields from the policy server CRD has the documentation updated.


viccuad (approved):
Thanks!

PR #1985, opened 2026-09-03
docs: add AGENTS.md guidance for AI coding agents

jvanz:
## Description

Add a root AGENTS.md and three nested files for the Rust workspace, the Helm chart and the controller package. They record the build, test and lint commands, the generated artifacts that must be refreshed with make generate, and the commit rules of the project.

The files also carry the Kubewarden AI policy. Contributors stay accountable for the changes, and every AI-assisted commit needs an Assisted-by trailer.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,9 +10,9 @@ one has precedence for the files in its subtree:
 
 ## Project overview
 
-Kubewarden is a CNCF dynamic admission controller for Kubernetes. It evaluates
-policies that are compiled to WebAssembly. This repository is a monorepo. It
-contains these components:
+- Kubewarden is a CNCF dynamic admission controller for Kubernetes.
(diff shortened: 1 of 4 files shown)


viccuad (approved):
I think this is a good start as any, but I would be more interested not on entry points (make targets) but more on architectural decisions that the agents must not trump over.

I believe the agents are knowledgeable enough to know nowadays where the CRD changes must go, or read the make target description. All this information about make targets and folders needs to be kept in sync too for little gain.

Points like the generated code are good, as those are more difficult to infer.

In general. I would also like to point to https://github.com/kubewarden/adm-controller/blob/main/CONTRIBUTING.md or similar, to not duplicate between that and the AGENTS.md.

viccuad (approved):

